PR #17, opened 2015-03-07
Add listed flag to registration blobs

RanjiniM:
1. Listed flag
2. Fix the bug of going from small->large registration and having 2
edits for the same version


xavierdecoster:
Is this (stale) PR still relevant? /cc @joelverhagen @johnataylor @joyhui

PR #18, opened 2015-04-23
Added namespace validation against AAD

maartenba:
Added namespace validation against AAD. Check the StorageRegistrationOwnership class' HasNamespace method. There is a functional question in there as to what should happen with MSA accounts.


maartenba:
@xavierdecoster could you do a quick review as well?

PR #21, opened 2015-05-18
Retrieve and store license agreement

maartenba:


src/Catalog/Ownership/StorageRegistration.cs
```diff
@@ -5,6 +5,7 @@
 using System;
```
johnataylor: I think we should just drop the whole StorageRegistration class at this point. We have the interface and the table implementation; its tedious keeping this up to date.

tests/PowerShellPublishTests/MockIRegistrationOwnership.cs
```diff
@@ -5,6 +5,7 @@
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
+using NuGet.Services.Metadata.Catalog.Ownership;
 using NuGet.Services.Publish;
 
 namespace PowerShellPublishTests
```
johnataylor: We can also drop the PowerShell specific code too. The PowerShell packages are just regular NuGet packages with a different type to filter on in search.

johnataylor:
Looks great. Couple of comments on dropping a couple of files (and we could do that as a separate check in if you like.)

maartenba:
Let's do those separately. Thanks @johnataylor

PR #22, opened 2015-06-02
Updated V2 searchermanager to use the new download look up

bhuvak:


bhuvak:
Only lowercase is added to the dict : https://github.com/NuGet/NuGet.Services.Metadata/blob/f94ff8f8134948f7796568988853ebbb284ed743/src/NuGet.Indexing/DownloadLookup.cs#L37

maartenba:
Looks good. Maybe consider using `new Dictionary<string, int>(StringComparer.InvariantCultureIgnoreCase);` to make the dictionary ensure lowercase.

PR #23, opened 2015-06-04
V3 search and autocomplete throw NullReferenceException on index swap

maartenba:
This PR fixes https://github.com/NuGet/NuGetGallery/issues/2520:
* Adds support for SearcherManager-specific IndexSearcher implementation
* Updates NuGetSearcherManager to use a NuGetIndexSearcher
* Adds all lookups and auxiliary data that is specific to the loaded index on NuGetIndexSearcher instead of NuGetSearcherManager
* Removes a lock while loading auxiliary data (can now be safely done as it is part of the NuGetIndexSearcher)
* Reduces memory usage for auxiliary data (e.g. versions by document lookup) as it no longer stores all data for bot http and https scheme

@bhuvak @johnataylor @xavierdecoster Can you review code changes?


johnataylor:
:shipit:

PR #28, opened 2015-10-09
Integrate package delete in NuGet gallery

maartenba:
See https://github.com/NuGet/NuGetGallery/issues/2714


src/Canton/CantonLib/jobs/CatalogPageCommitJob.cs
```diff
@@ -329,7 +329,7 @@ public CatalogPageCommitJob(Config config, Storage storage)
                 Stopwatch timer = new Stopwatch();
                 timer.Start();
 
-                IGraph commitData = PackageCatalog.CreateCommitMetadata(writer.RootUri, latestPublished, latestPublished);
+                IGraph commitData = PackageCatalog.CreateCommitMetadata(writer.RootUri, latestPublished, latestPublished, latestPublished);
```
yishaigalatzer: break the long lines

src/Catalog/PackageCatalog.cs
```diff
@@ -14,7 +14,7 @@ namespace NuGet.Services.Metadata.Catalog
 {
     public static class PackageCatalog
     {
-        public static IGraph CreateCommitMetadata(Uri indexUri, DateTime? lastCreated, DateTime? lastEdited)
+        public static IGraph CreateCommitMetadata(Uri indexUri, DateTime? lastCreated, DateTime? lastEdited, DateTime? lastDeleted)
```
yishaigalatzer: break long lines

src/Catalog/CloudBlobStorageExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.WindowsAzure.Storage.Blob;
+
+namespace NuGet.Services.Metadata.Catalog
+{
+    internal static class CloudBlobStorageExtensions
+    {
+        public static async Task<IEnumerable<IListBlobItem>> ListBlobsAsync(this CloudBlobContainer container, CancellationToken cancellationToken, string prefix = null, bool useFlatBlobListing = false, BlobListingDetails blobListingDetails = BlobListingDetails.None, long? maxResults = null)
```
yishaigalatzer: Don't use optional parameters if you don't need them (or another way, keep the signature to the minimum required by the code). Hard to tell in this review if all of them are need or not.

Also note that completely forbid optional parameters on any public API, because of backcompat issues

src/Catalog/CloudBlobStorageExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            List<IListBlobItem> items = new List<IListBlobItem>();
```
yishaigalatzer: use var

src/Catalog/CloudBlobStorageExtensions.cs
yishaigalatzer: since this method can return very large result sets and it runs over network, it seems much more appropriate to return IEnumerable and yield return instead of putting everything in a list.

src/Catalog/CloudBlobStorageExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+            BlobContinuationToken token = null;
+            do
+            {
+                var seg = await container.ListBlobsSegmentedAsync(prefix, useFlatBlobListing, blobListingDetails, null, token, null, null);
+                token = seg.ContinuationToken;
+                items.AddRange(seg.Results);
+
+                if (maxResults.HasValue && items.Count > maxResults.Value)
```
yishaigalatzer: just make maxResults be long (rather than long?) and make the default long.MaxValue

src/Catalog/CloudBlobStorageExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+                {
+                    break;
+                }
+            }
+            while (token != null && !cancellationToken.IsCancellationRequested);
+            return items;
```
yishaigalatzer: extra newline between while and return

src/Catalog/CloudBlobStorageExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
```
yishaigalatzer: one newline between methods

src/Catalog/CloudBlobStorageExtensions.cs
yishaigalatzer: Reading some more below. You never want to have two methods with optional parameters and the same name, once you add a third all kinds of breaks start happening

src/Catalog/DeleteCatalogItem.cs
```diff
@@ -0,0 +1,121 @@
...
+            _id = id;
+            _version = ParseVersion(version);
+            _published = published;
+
+            _catalogItemId = Guid.NewGuid();
+        }
+
+        private string ParseVersion(string version)
+        {
+            SemanticVersion semVer;
```
yishaigalatzer: Can we avoid nuget.core dependencies on all new code?

src/Catalog/DeleteCatalogItem.cs
```diff
@@ -0,0 +1,121 @@
...
+using Newtonsoft.Json.Linq;
+using NuGet.Services.Metadata.Catalog.Persistence;
+using NuGet.Versioning;
+using VDS.RDF;
+
+namespace NuGet.Services.Metadata.Catalog
+{
+    public class DeleteCatalogItem : AppendOnlyCatalogItem
+    {
+        private string _id;
```
yishaigalatzer: readonly to all members that can be readonly

src/Catalog/DeleteCatalogItem.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        public override IGraph CreatePageContent(CatalogContext context)
+        {
+            Uri resourceUri = new Uri(GetBaseAddress() + GetRelativeAddress());
+
+            Graph graph = new Graph();
+
+            INode subject = graph.CreateUriNode(resourceUri);
+
+            INode idPredicate = graph.CreateUriNode(Schema.Predicates.Id);
```
yishaigalatzer: var every day

src/Catalog/DeleteCatalogItem.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            if (_version != null)
+            {
+                graph.Assert(subject, versionPredicate, graph.CreateLiteralNode(_version));
+            }
+
+            return graph;
+        }
+
+        void SetIdVersionFromGraph(IGraph graph)
```
yishaigalatzer: private void

src/Catalog/DeleteCatalogItem.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            INode idPredicate = graph.CreateUriNode(Schema.Predicates.Id);
+            INode versionPredicate = graph.CreateUriNode(Schema.Predicates.Version);
+
+            INode rdfTypePredicate = graph.CreateUriNode(Schema.Predicates.Type);
+            Triple resource = graph.GetTriplesWithPredicateObject(rdfTypePredicate, graph.CreateUriNode(GetItemType())).First();
```
yishaigalatzer: are we guaranteed to get a value, why first vs. single? Can you add some comments.

src/Catalog/DeleteCatalogItem.cs
```diff
@@ -0,0 +1,121 @@
...
+            Triple id = graph.GetTriplesWithSubjectPredicate(resource.Subject, idPredicate).FirstOrDefault();
+            if (id != null)
+            {
+                _id = ((ILiteralNode)id.Object).Value;
+            }
+
+            Triple version = graph.GetTriplesWithSubjectPredicate(resource.Subject, versionPredicate).FirstOrDefault();
+            if (version != null)
```
yishaigalatzer: nit: `_version = version ?? ((ILiteralNode)version.Object).value;`

src/Catalog/PackageCatalog.cs
```diff
@@ -26,14 +26,19 @@ public static IGraph CreateCommitMetadata(Uri indexUri, DateTime? lastCreated, D
...
+            if (lastDeleted != null)
+            {
+                graph.Assert(graph.CreateUriNode(indexUri), graph.CreateUriNode(Schema.Predicates.LastDeleted), graph.CreateLiteralNode(lastDeleted.Value.ToString("O"), Schema.DataTypes.DateTime));
+            }
 
             return graph;
         }
 
-        public static async Task<Tuple<DateTime?, DateTime?>> ReadCommitMetadata(CatalogWriterBase writer, CancellationToken cancellationToken)
+        public static async Task<Tuple<DateTime?, DateTime?, DateTime?>> ReadCommitMetadata(CatalogWriterBase writer, CancellationToken cancellationToken)
```
yishaigalatzer: Unrelated, but please create a data structure instead of a tuple.

src/Catalog/CloudBlobStorageExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        public static async Task<IEnumerable<IListBlobItem>> ListBlobsAsync(this CloudBlobDirectory directory, CancellationToken cancellationToken, bool useFlatBlobListing = false, BlobListingDetails blobListingDetails = BlobListingDetails.None, long? maxResults = null)
+        {
+            List<IListBlobItem> items = new List<IListBlobItem>();
+            BlobContinuationToken token = null;
+            do
+            {
+                var seg = await directory.ListBlobsSegmentedAsync(useFlatBlobListing, blobListingDetails, null, token, null, null);
```
yishaigalatzer: use named variables for the nulls

src/Catalog/CloudBlobStorageExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+                token = seg.ContinuationToken;
+                items.AddRange(seg.Results);
+
+                if (maxResults.HasValue && items.Count > maxResults.Value)
+                {
+                    break;
+                }
+            }
+            while (token != null && !cancellationToken.IsCancellationRequested);
```
yishaigalatzer: the pattern is to throw a TaskCancelledException when cancellation is required

src/Catalog/DeleteCatalogItem.cs
```diff
@@ -0,0 +1,121 @@
...
+            _catalogItemId = Guid.NewGuid();
+        }
+
+        private string ParseVersion(string version)
+        {
+            SemanticVersion semVer;
+            if (SemanticVersion.TryParse(version, out semVer))
+            {
+                return semVer.ToNormalizedString();
+            }
```
yishaigalatzer: is the point of this to normalize the version? If so this should be called `TryNormalize` or `Normalize`

src/Catalog/DeleteCatalogItem.cs
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        public override Uri GetItemType()
+        {
+            return Schema.DataTypes.PackageDelete;
+        }
+
+        protected override string GetItemIdentity()
+        {
+            return _catalogItemId.ToString().ToLowerInvariant();
```
yishaigalatzer: why not make it a string to beginwith?

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -96,7 +105,7 @@ public override bool Exists(string fileName)
 
         protected override async Task<StorageContent> OnLoad(Uri resourceUri, CancellationToken cancellationToken)
         {
-            string name = GetName(resourceUri);
+            string name = GetName(resourceUri).TrimStart('/');
```
yishaigalatzer: add a comment to why this cleanup is necessary, and why it can't be done by GetName proper

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -76,6 +78,13 @@ public override bool Exists(string fileName)
             return false;
         }
 
+        public override async Task<IEnumerable<Uri>> List(bool recursive, CancellationToken cancellationToken)
```
yishaigalatzer: I don't see you ever using the recursive flat, so can we get rid of it?

src/Ng/DnxCatalogCollector.cs
```diff
@@ -35,35 +34,64 @@ public DnxCatalogCollector(Uri index, StorageFactory storageFactory, Func<HttpMe
...
+                    {
+                        Trace.TraceWarning("no nuspec available for {0}/{1} skipping", id, version);
+                    }
                 }
-                else
+                else if (type == Schema.DataTypes.PackageDelete.ToString())
                 {
-                    Trace.TraceWarning("no nuspec available for {0}/{1} skipping", id, version);
+                    // Delete package
+                    await UpdateMetadata(storage, versions => versions.Remove(NuGetVersion.Parse(version)), cancellationToken);
```
yishaigalatzer: what happens when one of these calls fails? Where is the retry queue for these?

src/Ng/Feed2Catalog.cs
```diff
@@ -240,6 +354,49 @@ private static DateTime ForceUTC(DateTime date)
             return lastDate;
         }
 
+        static async Task<DateTime> Deletes2Catalog(SortedList<DateTime, IList<Tuple<string, string>>> packages, Storage storage, DateTime lastCreated, DateTime lastEdited, DateTime lastDeleted, bool? createdPackages, CancellationToken cancellationToken)
```
yishaigalatzer: 1. private
2. When your list of parameters becomes this long, its a good sign you want an object model passed in

src/Ng/Feed2Catalog.cs
```diff
@@ -240,6 +354,49 @@ private static DateTime ForceUTC(DateTime date)
...
+        {
+            AppendOnlyCatalogWriter writer = new AppendOnlyCatalogWriter(storage, 550);
+
+            DateTime lastDate = createdPackages.HasValue ? (createdPackages.Value ? lastCreated : lastEdited) : DateTime.MinValue;
```
yishaigalatzer: break it down to if/else statements. nested ternaries are just impossible to read and maintain

src/Ng/Feed2Catalog.cs
```diff
@@ -240,6 +354,49 @@ private static DateTime ForceUTC(DateTime date)
             return lastDate;
         }
 
+        static async Task<DateTime> Deletes2Catalog(SortedList<DateTime, IList<Tuple<string, string>>> packages, Storage storage, DateTime lastCreated, DateTime lastEdited, DateTime lastDeleted, bool? createdPackages, CancellationToken cancellationToken)
+        {
+            AppendOnlyCatalogWriter writer = new AppendOnlyCatalogWriter(storage, 550);
```
yishaigalatzer: name parameter, what does 550 mean?

src/Ng/Feed2Catalog.cs
yishaigalatzer: switch SortedList<DateTime, IList<Tuple<string, string>>> to an object model, it is unreadable and unmaintanable

src/Ng/Feed2Catalog.cs
```diff
@@ -258,12 +415,14 @@ private static DateTime ForceUTC(DateTime date)
...
-            Storage storage = storageFactory.Create();
+            Storage catalogStorage = catalogStorageFactory.Create();
+            Storage auditingStorage = auditingStorageFactory.Create();
 
             const string LastCreated = "nuget:lastCreated";
             const string LastEdited = "nuget:lastEdited";
+            const string LastDeleted = "nuget:lastDeleted";
 
             int top = 20;
             int timeout = 300;
```
yishaigalatzer: nit: This should just be a timespan to beginwith

src/NuGet.Services.Publish/NuGet.Services.Publish.csproj
```diff
@@ -131,10 +132,10 @@
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\..\packages\Microsoft.WindowsAzure.ConfigurationManager.3.1.0\lib\net40\Microsoft.WindowsAzure.Configuration.dll</HintPath>
     </Reference>
-    <Reference Include="Microsoft.WindowsAzure.Diagnostics, Version=2.5.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35">
+    <Reference Include="Microsoft.WindowsAzure.Diagnostics, Version=2.7.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35">
```
yishaigalatzer: why is the upgrade merged into this change? Shouldn't we do it separately (separate deployment and verification)

maartenba: Moved the upgrade to master so it's not part of this change.

src/Catalog/DeleteCatalogItem.cs
maartenba: Making use of NuGet.Versioning here.

src/Catalog/DeleteCatalogItem.cs
maartenba: Different types (version is Triple, _version is string)

src/Ng/DnxCatalogCollector.cs
maartenba: Job fails and monitoring kicks in. @bhuvak may have more details

johnataylor:
:shipit:

PR #30, opened 2015-11-02
add Db2Luene as standalone command (since there are conflicts when ad…

joyhui:
Add this as a standalone command since we need it very often.


maartenba:
Mind if I add a few extra parameters to it?

joyhui:
Pls go ahead.

PR #31, opened 2015-11-18
Add tests for CamelCaseFilter and PackageAnalyzer

joelverhagen:
1. The `Owners` field in `PackageAnalyzer` is not referenced by the `NuGetQuery` code and it should be renamed to `Owner`.
2. Add unit tests for `CamelCaseFilter`.
2. Add unit tests for `PackageAnalyzer`.

### DIFF
--- a/tests/NuGet.IndexingTests/CamelCaseFilterTests.cs
+++ b/tests/NuGet.IndexingTests/CamelCaseFilterTests.cs
@@ -1,162 +1,138 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using Lucene.Net.Analysis.Standard;
-using Lucene.Net.Analysis.Tokenattributes;
-using NuGet.Indexing;
-using Xunit;
 using Lucene.Net.Util;
+using NuGet.Indexing;
 using NuGet.IndexingTests.TestSupport;
+using Xunit;
 
 namespace NuGet.IndexingTests
 {
     public class CamelCaseFilterTests
     {
-        [Theory, MemberData("TheoryData")]
-        public void Theory(string text, TokenAttributes[] expected)
+        [Theory]
+        [MemberData(nameof(TokenizingReturnsExpectedTermAndOffsetsData))]
+        public void TokenizingReturnsExpectedTermAndOffsets(string text, TokenAttributes[] expected)
         {
-            // ARRANGE, ACT
-            var actual = this.GetTokenAttributes(text).ToArray();
-
-            // ASSERT
-            Assert.Equal(expected.Length, actual.Length);
-            for (int i = 0; i < expected.Length; i++)
-            {
-                Assert.Equal(expected[i].Term, actual[i].Term);
-                Assert.Equal(expected[i].StartOffset, actual[i].StartOffset);
-                Assert.Equal(expected[i].EndOffset, actual[i].EndOffset);
-                Assert.Equal(expected[i].PositionIncrement, actual[i].PositionIncrement);
-            }
-        }
-
-        private IEnumerable<TokenAttributes> GetTokenAttributes(string text)
-        {
-            var textReader = new StringReader(text);
-            var tokenStream = new StandardTokenizer(Version.LUCENE_30, textReader);
+            // arrange
+            var tokenStream = new StandardTokenizer(Version.LUCENE_30, new StringReader(text));
             var filter = new CamelCaseFilter(tokenStream);
+            
+            // act
+            var actual = filter.Tokenize().ToArray();
 
-            var termAttribute = filter.GetAttribute<ITermAttribute>();
-            var offsetAttribute = filter.GetAttribute<IOffsetAttribute>();
-            var positionIncrementAttribute = filter.GetAttribute<IPositionIncrementAttribute>();
-
-            while (filter.IncrementToken())
-            {
-                yield return new TokenAttributes
-                {
-                    Term = termAttribute.Term,
-                    StartOffset = offsetAttribute.StartOffset,
-                    EndOffset = offsetAttribute.EndOffset,
-                    PositionIncrement = positionIncrementAttribute.PositionIncrement
-                };
-            }
+            // assert
+            Assert.Equal(expected, actual);
         }
 
-        public static IEnumerable<object[]> TheoryData
+        public static IEnumerable<object[]> TokenizingReturnsExpectedTermAndOffsetsData
         {
             get
             {
+                // shingle depth two with four terms
                 yield return new object[]
                 {
-                    "AaBbCcDd",
+                    "DotNetZipFoo",
                     new[]
                     {
-                        new TokenAttributes("AaBbCcDd", 0, 8, 1),
-                        new TokenAttributes("Aa", 0, 2, 0),
-                        new TokenAttributes("AaBb", 0, 4, 0),
-                        new TokenAttributes("Bb", 2, 4, 1),
-                        new TokenAttributes("BbCc", 2, 6, 0),
-                        new TokenAttributes("Cc", 4, 6, 1),
-                        new TokenAttributes("CcDd", 4, 8, 0),
-                        new TokenAttributes("Dd", 6, 8, 1)
+                        new TokenAttributes("DotNetZipFoo", 0, 12, 1),
+                        new TokenAttributes("Dot", 0, 3, 0),
+                        new TokenAttributes("DotNet", 0, 6, 0),
+                        new TokenAttributes("Net", 3, 6, 1),
+                        new TokenAttributes("NetZip", 3, 9, 0),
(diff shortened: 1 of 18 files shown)


tests/NuGet.IndexingTests/CamelCaseFilterTests.cs
```diff
@@ -0,0 +1,165 @@
...
+using Xunit;
+using Lucene.Net.Util;
+using NuGet.IndexingTests.TestSupport;
+
+namespace NuGet.IndexingTests
+{
+    public class CamelCaseFilterTests
+    {
+        [Theory, MemberData("TheoryData")]
+        public void Theory(string text, TokenAttributes[] expected)
```
yishaigalatzer: Name a test so it is meaningful.

e.g. TokenizingReturnsTheExpectedTermAndOffsets

tests/NuGet.IndexingTests/CamelCaseFilterTests.cs
```diff
@@ -0,0 +1,165 @@
...
+using NuGet.Indexing;
+using Xunit;
+using Lucene.Net.Util;
+using NuGet.IndexingTests.TestSupport;
+
+namespace NuGet.IndexingTests
+{
+    public class CamelCaseFilterTests
+    {
+        [Theory, MemberData("TheoryData")]
```
yishaigalatzer: two attributes

yishaigalatzer: `nameof(TheoryData)`

tests/NuGet.IndexingTests/PackageAnalyzerTests.cs
```diff
@@ -0,0 +1,125 @@
...
+                yield return GetDescriptionTestCase("Summary");
+                yield return GetDescriptionTestCase("Authors");
+                
+                yield return new object[]
+                {
+                    "Owner",
+                    "AAA Bbb",
+                    new[]
+                    {
+                        new TokenAttributes("aaa bbb", 0, 7)
```
johnataylor: isn't this a bug?

tests/NuGet.IndexingTests/CamelCaseFilterTests.cs
```diff
@@ -0,0 +1,165 @@
...
+        public void Theory(string text, TokenAttributes[] expected)
+        {
+            // ARRANGE, ACT
+            var actual = this.GetTokenAttributes(text).ToArray();
```
yishaigalatzer: no `this.`

tests/NuGet.IndexingTests/CamelCaseFilterTests.cs
```diff
@@ -0,0 +1,165 @@
...
+            for (int i = 0; i < expected.Length; i++)
+            {
+                Assert.Equal(expected[i].Term, actual[i].Term);
+                Assert.Equal(expected[i].StartOffset, actual[i].StartOffset);
+                Assert.Equal(expected[i].EndOffset, actual[i].EndOffset);
+                Assert.Equal(expected[i].PositionIncrement, actual[i].PositionIncrement);
+            }
+        }
+
+        private IEnumerable<TokenAttributes> GetTokenAttributes(string text)
```
yishaigalatzer: `Tokenize`?

tests/NuGet.IndexingTests/PackageAnalyzerTests.cs
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        public static IEnumerable<object[]> TheoryData
+        {
+            get
+            {
+                yield return new object[]
+                {
+                    "Id",
+                    "AaBbCc",
```
johnataylor: wouldn't it also be nice to have a 4 deep?
and this test is a little more than just a CamelCase test, having decided to go there, why haven't you also tested some permutations that includes dot characters for example, something like: YouNeverTurnedAround.ToSee.TheJugglers.AndThe.Clowns

tests/NuGet.IndexingTests/CamelCaseFilterTests.cs
```diff
@@ -0,0 +1,165 @@
...
+                    "Abcd",
+                    new[]
+                    {
+                        new TokenAttributes("Abcd", 0, 4, 1)
+                    }
+                };
+
+                yield return new object[]
+                {
+                    "A1a2B3b",
```
yishaigalatzer: Add comments to describe that this is the current behavior, but something that we might want to tweak in the future. So this doesn't get interpreted as spec

tests/NuGet.IndexingTests/PackageAnalyzerTests.cs
```diff
@@ -0,0 +1,125 @@
...
+using NuGet.Indexing;
+using NuGet.IndexingTests.TestSupport;
+using Xunit;
+
+namespace NuGet.IndexingTests
+{
+    public class PackageAnalyzerTests
+    {
+        [Theory, MemberData("TheoryData")]
+        public void AddsCorrectFieldAnalyzers(string field, string text, TokenAttributes[] expected)
```
johnataylor: just a question really, but structuring the tests like this as an enumeration rather than separate tests - isn't that harder to isolate results - the whole thing is reported as one shot rather than one part failing - isn't that harder to reason over?

tests/NuGet.IndexingTests/TestSupport/TokenAttributes.cs
```diff
@@ -0,0 +1,59 @@
...
+
+namespace NuGet.IndexingTests.TestSupport
+{
+    public class TokenAttributes
+    {
+        public TokenAttributes()
+        {
+        }
+
+        public TokenAttributes(string term, int startOffset, int endOffset) : this(term, startOffset, endOffset, null)
```
yishaigalatzer: this constructor is unnecesssary, for a test project feel free to use default parameters

tests/NuGet.IndexingTests/TestSupport/TokenAttributes.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        public override bool Equals(object obj)
+        {
+            var other = obj as TokenAttributes;
+            if (other == null)
+            {
+                return false;
+            }
+
```
yishaigalatzer: one line spacing

yishaigalatzer:
Looks good.

Use meaningful identifiers in the test and add comments to what cases you are covering (so it is easy to understand the next time someone updates the test)

Using meaningful identifiers makes the scenarios tested make more sense.

:shipit: when @johnataylor is happy

tests/NuGet.IndexingTests/PackageAnalyzerTests.cs
joelverhagen: I don't think so. All four of these fields have similar types of content and should be tokenized similarly (DotTokenizer, case insensitive, remove stop words, split on camel case). However, you have a lot more context here so please correct me.

tests/NuGet.IndexingTests/PackageAnalyzerTests.cs
joelverhagen: At least in VS, the failures and the successes are reported separately (including the assertion failure, the test output, etc).

tests/NuGet.IndexingTests/CamelCaseFilterTests.cs
joelverhagen: Fixed.

tests/NuGet.IndexingTests/CamelCaseFilterTests.cs
joelverhagen: Fixed.

tests/NuGet.IndexingTests/CamelCaseFilterTests.cs
joelverhagen: Fixed.

tests/NuGet.IndexingTests/CamelCaseFilterTests.cs
joelverhagen: Fixed.

tests/NuGet.IndexingTests/CamelCaseFilterTests.cs
joelverhagen: Done.

tests/NuGet.IndexingTests/TestSupport/TokenAttributes.cs
joelverhagen: Fixed.

tests/NuGet.IndexingTests/TestSupport/TokenAttributes.cs
joelverhagen: Fixed.

tests/NuGet.IndexingTests/PackageAnalyzerTests.cs
joelverhagen: I tried to limit the test cases in PackageAnalyzer to only verify that the correct analyzer per field is selected. I added new test classes for each analyzer to cover these additional, deeper cases.

johnataylor:
:shipit:

PR #36, opened 2015-11-25
Add first functional test for the search service

joelverhagen:
1. Add stitch together code that downloads packages, builds a Lucene index, and starts the server.
2. Add simple functional test for NuGet/Home#833.

Please give me any feedback you can on this framework. I am still pretty ignorant about this stuff. Areas in particular that need review:

- Should we download packages from nuget.org? I am `DataServicePackageRepository` to do this. Alternatives are reading .nupkgs from a share or checking them in maybe?
- Should each test start and stop the service on its own? Right now I use a class fixture to do all of the setup once per *test run*. That is, all tests share the same service, Lucene index, and data directory.
- Are the directories for the test data okay? I am using `Directory.GetCurrentDirectory()` as the base. I think TeamCity cleans this up automatically.
- Should I explicitly clean up the test data? Always? On successful tests (this one may be tricky)?

@johnataylor @joyhui @maartenba @xavierdecoster @yishaigalatzer

### DIFF
--- a/src/NuGet.Indexing/DocumentCreator.cs
+++ b/src/NuGet.Indexing/DocumentCreator.cs
@@ -42,7 +42,7 @@ namespace NuGet.Indexing
 
             AddId(document, package, errors);
             AddVersion(document, package, errors);
-            AddField(document, "Title", package, "title", Field.Index.ANALYZED, 2.0f);
+            AddFieldWithDefault(document, "Title", package, "title", package["id"], Field.Index.ANALYZED, 2.0f);
             AddField(document, "Description", package, "description", Field.Index.ANALYZED);
             AddField(document, "Summary", package, "summary", Field.Index.ANALYZED);
             AddField(document, "Tags", package, "tags", Field.Index.ANALYZED, 2.0f);
@@ -65,7 +65,7 @@ namespace NuGet.Indexing
             AddDates(document, package, errors);
             AddSupportedFrameworks(document, package);
 
-            AddField(document, "Listed", package, "listed", Field.Index.NOT_ANALYZED);
+            AddRequiredField(document, "Listed", package, errors, "listed", Field.Index.NOT_ANALYZED);
 
             DetermineLanguageBoost(document, package);
 
@@ -74,7 +74,7 @@ namespace NuGet.Indexing
             return document;
         }
 
-        static void CheckErrors(List<string> errors)
+        private static void CheckErrors(List<string> errors)
         {
             if (errors.Count > 0)
             {
@@ -87,7 +87,7 @@ namespace NuGet.Indexing
             }
         }
 
-        static void AddId(Document document, IDictionary<string, string> package, List<string> errors)
+        private static void AddId(Document document, IDictionary<string, string> package, List<string> errors)
         {
             string value;
             if (package.TryGetValue("id", out value))
@@ -108,7 +108,7 @@ namespace NuGet.Indexing
             }
         }
 
-        static void AddVersion(Document document, IDictionary<string, string> package, List<string> errors)
+        private static void AddVersion(Document document, IDictionary<string, string> package, List<string> errors)
         {
             string originalVersion = null;
             if (package.TryGetValue("originalVersion", out originalVersion))
@@ -143,7 +143,7 @@ namespace NuGet.Indexing
             }
         }
 
-        static void AddDependencies(Document document, IDictionary<string, string> package)
+        private static void AddDependencies(Document document, IDictionary<string, string> package)
         {
             string value;
             if (package.TryGetValue("flattenedDependencies", out value))
@@ -191,7 +191,7 @@ namespace NuGet.Indexing
             }
         }
 
-        static void AddDates(Document document, IDictionary<string, string> package, List<string> errors)
+        private static void AddDates(Document document, IDictionary<string, string> package, List<string> errors)
         {
             string created;
             if (package.TryGetValue("created", out created))
@@ -240,7 +240,7 @@ namespace NuGet.Indexing
             }
         }
 
-        static void AddSupportedFrameworks(Document document, IDictionary<string, string> package)
+        private static void AddSupportedFrameworks(Document document, IDictionary<string, string> package)
         {
             string value;
             if (package.TryGetValue("supportedFrameworks", out value))
@@ -265,16 +265,55 @@ namespace NuGet.Indexing
             }
         }
 
-        static void AddField(Document document, string destName, IDictionary<string, string> package, string sourceName, Field.Index fieldIndex, float boost = 1.0f)
+        private static void AddRequiredField(
+            Document document,
+            string destName,
+            IDictionary<string, string> package,
+            List<string> errors,
+            string sourceName,
+            Field.Index fieldIndex,
+            float boost = 1.0f)
+        {
+            if (!AddField(document, destName, package, sourceName, fieldIndex, boost))
+            {
+                errors.Add($"Required property '{sourceName}' not found.");
+            }
+        }
+
+        private static void AddFieldWithDefault(
+            Document document,
+            string destName,
+            IDictionary<string, string> package,
+            string sourceName,
+            string defaultValue,
+            Field.Index fieldIndex,
+            float boost = 1.0f)
+        {
+            if (!AddField(document, destName, package, sourceName, fieldIndex, boost))
+            {
+                document.Add(new Field(destName, defaultValue, Field.Store.YES, fieldIndex) { Boost = boost });
+            }
+        }
+
+        private static bool AddField(
+            Document document,
+            string destName,
+            IDictionary<string, string> package,
+            string sourceName,
+            Field.Index fieldIndex,
+            float boost = 1.0f)
         {
             string value;
             if (package.TryGetValue(sourceName, out value))
             {
                 document.Add(new Field(destName, value, Field.Store.YES, fieldIndex) { Boost = boost });
+                return true;
             }
+
+            return false;
         }
 
-        static void DetermineLanguageBoost(Document document, IDictionary<string, string> package)
+        private static void DetermineLanguageBoost(Document document, IDictionary<string, string> package)
         {
             string id;
             string language;
--- a/src/NuGet.Indexing/PackageMetadataExtraction.cs
+++ b/src/NuGet.Indexing/PackageMetadataExtraction.cs
@@ -78,11 +78,12 @@ namespace NuGet.Indexing
                 return;
             }
 
-            ExtractId(package, document, errors);
-            ExtractVersion(package, document, errors);
+            ExtractRequiredProperty(package, document, errors, "id");
+            ExtractRequiredProperty(package, document, errors, "version");
+            ExtractProperty(package, document, "title");
             ExtractProperty(package, document, "summary");
             ExtractProperty(package, document, "tags");
-            ExtractProperty(package, document, "authors");
+            ExtractProperty(package, document, "authors", "flattenedAuthors");
             ExtractProperty(package, document, "description");
             ExtractProperty(package, document, "iconUrl");
             ExtractProperty(package, document, "projectUrl");
@@ -93,9 +94,12 @@ namespace NuGet.Indexing
             ExtractProperty(package, document, "licenseUrl");
             ExtractProperty(package, document, "requiresLicenseAcceptance");
             ExtractDependencies(package, document);
+
+            package["published"] = DateTimeOffset.UtcNow.ToString("O");
+            package["listed"] = "true";
         }
 
-        static bool TryLoad(Stream stream, out XDocument document, List<string> errors)
+        private static bool TryLoad(Stream stream, out XDocument document, List<string> errors)
         {
             try
             {
@@ -110,42 +114,45 @@ namespace NuGet.Indexing
             }
         }
 
-        static void ExtractId(IDictionary<string, string> package, XDocument document, List<string> errors)
+        /// <summary>
+        /// Same as <see cref="ExtractProperty"/> except an error is added if the property is not found in the .nuspec.
+        /// </summary>
+        /// <param name="package">The package data.</param>
+        /// <param name="document">The parsed .nuspec.</param>
+        /// <param name="errors">The list of errors encountered while parsing the .nuspec.</param>
+        /// <param name="localName">The local name of the XML element to read the text from.</param>
+        /// <param name="key">Optionally, the key to set in the <see cref="package"/> dictionary.</param>
+        private static void ExtractRequiredProperty(IDictionary<string, string> package, XDocument document, List<string> errors, string localName, string key = null)
         {
-            XElement idElement = document.Root.DescendantsAndSelf().Elements().Where(d => d.Name.LocalName == "id").FirstOrDefault();
-            if (idElement != null)
+            if (!ExtractProperty(package, document, localName, key))
             {
-                package["id"] = (idElement.Value);
-            }
-            else
-            {
-                errors.Add("unable to find the id element in the nuspec");
+                errors.Add($"unable to find the '{localName}' element in the nuspec");
             }
         }
 
-        static void ExtractVersion(IDictionary<string, string> package, XDocument document, List<string> errors)
+        /// <summary>
+        /// Extract the property with local name <see cref="localName"/> from the .nuspec XML and added it to the
+        /// <see cref="package"/> dictionary. If <see cref="key"/> is provided, this value will be used as the key in
+        /// <see cref="package"/>. Otherwise, <see cref="localName"/> is used as the key.
+        /// </summary>
+        /// <param name="package">The package data.</param>
+        /// <param name="document">The parsed .nuspec.</param>
+        /// <param name="localName">The local name of the XML element to read the text from.</param>
+        /// <param name="key">Optionally, the key to set in the <see cref="package"/> dictionary.</param>
+        /// <returns>True if the element with the provided <see cref="localName"/> is found. False, otherwise.</returns>
+        private static bool ExtractProperty(IDictionary<string, string> package, XDocument document, string localName, string key = null)
         {
-            XElement idElement = document.Root.DescendantsAndSelf().Elements().Where(d => d.Name.LocalName == "version").FirstOrDefault();
-            if (idElement != null)
-            {
-                package["version"] = (idElement.Value);
-            }
-            else
-            {
-                errors.Add("unable to find the version element in the nuspec");
-            }
-        }
-
-        static void ExtractProperty(IDictionary<string, string> package, XDocument document, string name)
-        {
-            XElement element = document.Root.DescendantsAndSelf().Elements().Where(d => d.Name.LocalName == name).FirstOrDefault();
+            XElement element = document.Root.DescendantsAndSelf().Elements().FirstOrDefault(d => d.Name.LocalName == localName);
             if (element != null)
             {
-                package[name] = element.Value;
+                package[key ?? localName] = element.Value;
+                return true;
             }
+
+            return false;
         }
 
-        static void ExtractDependencies(IDictionary<string, string> package, XDocument document)
+        private static void ExtractDependencies(IDictionary<string, string> package, XDocument document)
         {
             //TODO: extract from the XML - refer to the XSLT for an accurate definition of the generic structure
         }
--- a/src/NuGet.Services.BasicSearch/Startup.cs
+++ b/src/NuGet.Services.BasicSearch/Startup.cs
@@ -171,16 +171,16 @@ namespace NuGet.Services.BasicSearch
             {
                 if (_searcherManager == null)
                 {
-                    await context.Response.WriteAsync("UNINITIALIZED");
                     context.Response.StatusCode = (int)HttpStatusCode.OK;
+                    await context.Response.WriteAsync("UNINITIALIZED");
                 }
                 else
                 {
                     switch (context.Request.Path.Value)
                     {
                         case "/":
-                            await context.Response.WriteAsync("READY");
                             context.Response.StatusCode = (int)HttpStatusCode.OK;
+                            await context.Response.WriteAsync("READY");
                             break;
                         case "/find":
                             await ServiceHelpers.WriteResponse(context, HttpStatusCode.OK, ServiceImpl.Find(context, _searcherManager));
@@ -204,8 +204,8 @@ namespace NuGet.Services.BasicSearch
                             break;
 
                         default:
-                            await context.Response.WriteAsync("unrecognized");
                             context.Response.StatusCode = (int)HttpStatusCode.NotFound;
+                            await context.Response.WriteAsync("UNRECOGNIZED");
                             break;
                     }
                 }
--- a/tests/NuGet.Services.BasicSearchTests/StartupTests.cs
+++ b/tests/NuGet.Services.BasicSearchTests/StartupTests.cs
@@ -1,0 +1,50 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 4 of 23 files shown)


src/NuGet.Services.BasicSearch/Startup.cs
```diff
@@ -171,16 +171,16 @@ static Uri MakeRegistrationBaseAddress(string scheme, string registrationBaseAdd
             {
                 if (_searcherManager == null)
                 {
-                    await context.Response.WriteAsync("UNINITIALIZED");
                     context.Response.StatusCode = (int)HttpStatusCode.OK;
+                    await context.Response.WriteAsync("UNINITIALIZED");
```
joelverhagen: Response needs to be written after status code is sent.

tests/NuGet.Services.BasicSearchTests/StartupTests.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        private readonly StartupTestsFixture _fixture;
+
+        public StartupTests(StartupTestsFixture fixture)
+        {
+            _fixture = fixture;
+        }
+
+        [Fact]
+        public async Task AutocompleteParameterIsObserved()
```
joelverhagen: This is the one test.

tests/NuGet.Services.BasicSearchTests/TestSupport/DataInitializer.cs
```diff
@@ -0,0 +1,39 @@
...
+            _settings = settings;
+        }
+
+        public async Task InitializeAsync()
+        {
+            Directory.CreateDirectory(_settings.DataDirectory);
+            await WriteFileAsync("downloads.v1.json", "[]");
+            await WriteFileAsync("curatedfeeds.json", "[]");
+            await WriteFileAsync("owners.json", "[]");
+            await WriteFileAsync("rankings.v1.json", "{\"Rank\": []}");
```
joelverhagen: This essentially makes search ranking by download count ineffective. Future tests may want to test this.

tests/NuGet.Services.BasicSearchTests/Fixtures/StartupTestsFixture.cs
```diff
@@ -0,0 +1,84 @@
...
+                DataDirectory = GetDirectoryPath("Data"),
+                RegistrationBaseAddress = "http://api.nuget.org/v3/registration0/"
+            };
+
+            // initialize the Lucene and Data directories
+            await new LuceneInitializer(settings).InitializeAsync();
+            await new DataInitializer(settings).InitializeAsync();
+
+            // establish configuration for the web service
+            ConfigurationManager.AppSettings.Set("Local.Lucene.Directory", settings.LuceneDirectory);
```
joelverhagen: This is intrinsically not thread safe. Therefore, if we ever want different tests having different settings, we will need to refactor the service.

tests/NuGet.Services.BasicSearchTests/TestSupport/PackageVersionIdentifier.cs
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace NuGet.Services.BasicSearchTests.TestSupport
+{
+    public class PackageVersionIdentifier
```
joyhui: It is typical example for a "struct".

tests/NuGet.Services.BasicSearchTests/TestSupport/LuceneInitializer.cs
```diff
@@ -0,0 +1,103 @@
...
+        private async Task DownloadPackagesAsync()
+        {
+            System.IO.Directory.CreateDirectory(_settings.PackageDirectory);
+            await Task.WhenAll(_settings.PackageVersions.Select(DownloadPackageAsync));
+        }
+
+        private void CreateLuceneIndex()
+        {
+            var directoryInfo = new DirectoryInfo(_settings.LuceneDirectory);
+            directoryInfo.Create();
```
joyhui: If the lucene index already created, do we clear old one or reuse old one?
People may run function test on his box more than once.

joelverhagen: Yes, the `true` parameter to `Document.CreateIndexWriter` does this.

tests/NuGet.Services.BasicSearchTests/TestSupport/DataInitializer.cs
joyhui: We may have to expose interface to change/initiate those file too. 
One idea is to add different section to the BasicSearchTestSettings itself, or let individual test to set content by themselves using interface.

joelverhagen: Yup, that's one option. I was also thinking that it could be set as part of the list of packages to download and include in the Lucene index (that is, the `PackageVersionIdentifier` type). If it's okay with you, I'll wait until we need it to build it.

tests/NuGet.Services.BasicSearchTests/TestSupport/PackageVersionIdentifier.cs
joelverhagen: Good point. Fixed.

tests/NuGet.Services.BasicSearchTests/Fixtures/StartupTestsFixture.cs
```diff
@@ -0,0 +1,84 @@
...
+                        new PackageVersionIdentifier("Antlr", "3.1.3.42154"),
+                        new PackageVersionIdentifier("Antlr", "3.4.1.9004-pre"),
+                        new PackageVersionIdentifier("angularjs", "1.2.0-RC1"),
+                        new PackageVersionIdentifier("WebGrease", "1.6.0")
+                    }
+                }
+            };
+
+            // construct the test settings
+            var settings = new BasicSearchTestSettings
```
joyhui: Please use a xml file or json file for this.

tests/NuGet.Services.BasicSearchTests/StartupTests.cs
```diff
@@ -0,0 +1,61 @@
...
+            Assert.Equal("1.2.0-RC1", withPrerelease.GetPackageVersion("angularjs"));
+            Assert.Equal("3.4.1.9004-pre", withPrerelease.GetPackageVersion("Antlr"));
+            Assert.Equal("1.6.0", withPrerelease.GetPackageVersion("WebGrease"));
+
+            Assert.False(withoutPrerelease.ContainsPackage("angularjs"));
+            Assert.Equal("3.1.3.42154", withoutPrerelease.GetPackageVersion("Antlr"));
+            Assert.Equal("1.6.0", withoutPrerelease.GetPackageVersion("WebGrease"));
+        }
+
+        private static string Query(object obj)
```
joyhui: Since the format of query for search is fixed, can we add a helper class for it?

class SearchQueryBuilder
{
    public static string base; --- static or retrieve from somewhere else 
    public int skip {get;set}
    public bool prerease {get;set}
    public override ToString() ---//  base?query?.....
}

tests/NuGet.Services.BasicSearchTests/StartupTests.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            var withoutPrereleaseResponse = await app.Client.GetAsync("/query?" + Query(new { q = query, prerelease = false }));
+            var withoutPrerelease = await withoutPrereleaseResponse.Content.ReadAsAsync<QueryResult>();
+
+            // Assert
+            Assert.Equal(HttpStatusCode.OK, withPrereleaseResponse.StatusCode);
+            Assert.Equal(HttpStatusCode.OK, withoutPrereleaseResponse.StatusCode);
+
+            Assert.Equal("1.2.0-RC1", withPrerelease.GetPackageVersion("angularjs"));
+            Assert.Equal("3.4.1.9004-pre", withPrerelease.GetPackageVersion("Antlr"));
```
joyhui: It is good always put an failed msg at the end:
Assert.Equal("3.4.1.9004-pre", withPrerelease.GetPackageVersion("Antlr"), "Antlr package should be returned");

tests/NuGet.Services.BasicSearchTests/StartupTests.cs
```diff
@@ -0,0 +1,61 @@
...
+namespace NuGet.Services.BasicSearchTests
+{
+    public class StartupTests : IClassFixture<StartupTestsFixture>
+    {
+        private readonly StartupTestsFixture _fixture;
+
+        public StartupTests(StartupTestsFixture fixture)
+        {
+            _fixture = fixture;
+        }
```
joyhui: We may want to set the packageVersion at startup, build lucene, and let service refresh. In this way we could make sure lucene is clean and won't affect other tests.

tests/NuGet.Services.BasicSearchTests/StartupTests.cs
joelverhagen: I was wanting to do this too, but Xunit does not support this (except on `Assert.Fail`).

tests/NuGet.Services.BasicSearchTests/StartupTests.cs
joelverhagen: I think you're right. I'll do this.

tests/NuGet.Services.BasicSearchTests/Fixtures/StartupTestsFixture.cs
maartenba: My vote is on using the approach we have in the gallery functional tests: use a `ConfigurationProvider` that does a fallback:

* Environment variable with that name set? Use environment variable.
* No environment variable but a setting exists? Use setting.
* None set? `null`

This makes it easier to change the test setting during a run. Optionally a test setup/teardown can set/unset a variable as well.

tests/NuGet.Services.BasicSearchTests/TestSupport/LuceneInitializer.cs
maartenba: Why not a simple `RAMDirectory` here? Wondering if each test(class) should just create a fresh index for it to run. Makes the tests very isolated and less prone to environmental issues (like a `write.lock` that remains and other things that can go wrong with the file system)

tests/NuGet.Services.BasicSearchTests/Fixtures/StartupTestsFixture.cs
maartenba: If it's only four of them I prefer code. Makes me grasp the test when reading it without having to context switch to yet another file.

tests/NuGet.Services.BasicSearchTests/StartupTests.cs
maartenba: `Assert` can be extended but I don't think we should write our own xUnit. Perhaps add a line comment describing the assertion so that when you navigate to the failed assertion some extra info is available to the dev?

maartenba:
In reply to your questions:

> Should we download packages from nuget.org? I am DataServicePackageRepository to do this. Alternatives are reading .nupkgs from a share or checking them in maybe?

No strong opinion. Since NuGet.org has the premise of "no deletes", it should be safe to depend on it to fetch packages.

> Should each test start and stop the service on its own? Right now I use a class fixture to do all of the setup once per test run. That is, all tests share the same service, Lucene index, and data directory.

My vote is on service per test. Makes them all isolated and very much able to run in parallel. If we expect a delete in some test, for example, a shared service means the test can not run in parallel if another test depends on the deleted package being there.

> Are the directories for the test data okay? I am using Directory.GetCurrentDirectory() as the base. I think TeamCity cleans this up automatically.
Should I explicitly clean up the test data? Always? On successful tests (this one may be tricky)?

How about a `RAMDirectory`?

src/NuGet.Services.BasicSearch/Startup.cs
```diff
@@ -204,8 +204,8 @@ static Uri MakeRegistrationBaseAddress(string scheme, string registrationBaseAdd
                             break;
 
                         default:
-                            await context.Response.WriteAsync("unrecognized");
                             context.Response.StatusCode = (int)HttpStatusCode.NotFound;
+                            await context.Response.WriteAsync("unrecognized");
```
maartenba: Other messages are in all caps.

tests/NuGet.Services.BasicSearchTests/TestSupport/DataInitializer.cs
maartenba: Just like the `RAMDirectory` comment, I always feel wary when tests depend on the file system. If we introduce a test that does require this data to be populated, we can no longer run them in parallel (unless we use a different directory per test). Imo the code should take in the dictionary/enumerable, not the file.

tests/NuGet.Services.BasicSearchTests/TestSupport/LuceneInitializer.cs
```diff
@@ -0,0 +1,103 @@
...
+            var path = GetPackagePath(version);
+            using (var fileStream = new FileStream(path, FileMode.Open, FileAccess.Read))
+            {
+                var errors = new List<string>();
+                var metadata = PackageMetadataExtraction.MakePackageMetadata(fileStream, errors);
+
+                if (errors.Any())
+                {
+                    throw new InvalidOperationException(
+                        $"NuGet package for '{version.Id}' (version '{version.Version}') could be read for metadata. " +
```
maartenba: could *not* be read

tests/NuGet.Services.BasicSearchTests/TestSupport/LuceneInitializer.cs
```diff
@@ -0,0 +1,103 @@
...
+                return metadata;
+            }
+        }
+
+        private string GetPackagePath(PackageVersionIdentifier version)
+        {
+            return Path.Combine(_settings.PackageDirectory, $"{version.Id}.{version.Version}.nupkg".ToLower());
+        }
+        
+        private async Task DownloadPackageAsync(PackageVersionIdentifier version)
```
maartenba: Wondering if a `NuGet.Core` dependency is needed. We could just do a download from the DNX flat container?

tests/NuGet.Services.BasicSearchTests/Fixtures/StartupTestsFixture.cs
joelverhagen: I think I will address this in a different PR. I wanted to minimize changes to the product code to get this functional test in place.

tests/NuGet.Services.BasicSearchTests/StartupTests.cs
joelverhagen: Good compromise. I'll do that.

tests/NuGet.Services.BasicSearchTests/TestSupport/LuceneInitializer.cs
joelverhagen: This will required a lot of changes to the service (since it only uses a file system directory right now). I would like to do this as a separate PR, if that's okay.

maartenba: Sounds good!

tests/NuGet.Services.BasicSearchTests/TestSupport/LuceneInitializer.cs
joelverhagen: I am not sure what you mean about this. Could you elaborate or point me to some code that does this?

tests/NuGet.Services.BasicSearchTests/TestSupport/LuceneInitializer.cs
joelverhagen: Fixed.

src/NuGet.Services.BasicSearch/Startup.cs
joelverhagen: Fixed.

tests/NuGet.Services.BasicSearchTests/StartupTests.cs
joelverhagen: Fixed.

tests/NuGet.Services.BasicSearchTests/Fixtures/StartupTestsFixture.cs
joelverhagen: I've made the package list inline with the code and values the same for all tests in XML.

tests/NuGet.Services.BasicSearchTests/TestSupport/LuceneInitializer.cs
maartenba: If this is the only place `NuGet.Core` is used, you could just HTTP GET the package instead using the following URL format:
`https://api.nuget.org/v3-flatcontainer/{packageid-lowercase}/{packageversion-lowercase}/{packageid-lowercase}.{packageversion-lowercase}.nupkg`

(example `https://api.nuget.org/v3-flatcontainer/newtonsoft.json/4.0.1/newtonsoft.json.4.0.1.nupkg`)

joelverhagen: Okay, this is a lot simpler. I'll do this.

src/NuGet.Indexing/DocumentCreator.cs
```diff
@@ -264,14 +264,32 @@ static void AddSupportedFrameworks(Document document, IDictionary<string, string
                 }
             }
         }
+        static void AddRequiredField(Document document, string destName, IDictionary<string, string> package, List<string> errors, string sourceName, Field.Index fieldIndex, float boost = 1.0f)
```
yishaigalatzer: private static void (and newline above this method), also try to break the long lines, so the whole line is readable when reviewing the PR

src/NuGet.Indexing/DocumentCreator.cs
```diff
@@ -264,14 +264,32 @@ static void AddSupportedFrameworks(Document document, IDictionary<string, string
...
+            {
+                document.Add(new Field(destName, defaultValue, Field.Store.YES, fieldIndex) { Boost = boost });
+            }
+        }
 
-        static void AddField(Document document, string destName, IDictionary<string, string> package, string sourceName, Field.Index fieldIndex, float boost = 1.0f)
+        static bool AddField(Document document, string destName, IDictionary<string, string> package, string sourceName, Field.Index fieldIndex, float boost = 1.0f)
         {
             string value;
             if (package.TryGetValue(sourceName, out value))
```
yishaigalatzer: can the value be accidentially be null or empty?

src/NuGet.Indexing/PackageMetadataExtraction.cs
```diff
@@ -93,6 +94,9 @@ public static void MakePackage(Stream stream, IDictionary<string, string> packag
             ExtractProperty(package, document, "licenseUrl");
             ExtractProperty(package, document, "requiresLicenseAcceptance");
             ExtractDependencies(package, document);
+
+            package["published"] = DateTime.UtcNow.ToString("O");
```
yishaigalatzer: can we switch to using DateTimeOffset instead of DateTime. Generally DateTimeOffset is recommended to be used whenever you can instead of DateTime.

src/NuGet.Indexing/PackageMetadataExtraction.cs
```diff
@@ -110,39 +114,24 @@ static bool TryLoad(Stream stream, out XDocument document, List<string> errors)
...
-            else
+            if (!ExtractProperty(package, document, localName, key))
             {
-                errors.Add("unable to find the version element in the nuspec");
+                errors.Add($"unable to find the '{localName}' element in the nuspec");
             }
         }
 
-        static void ExtractProperty(IDictionary<string, string> package, XDocument document, string name)
+        static bool ExtractProperty(IDictionary<string, string> package, XDocument document, string localName, string key = null)
```
yishaigalatzer: this is kind of a weird API, if you really want to keep it, please add an doc explaining the usage.

Also in general you want to avoid methods with default parameters values

src/NuGet.Services.BasicSearch/Startup.cs
yishaigalatzer: :+1: which is why in webapi a response is a single object that travels up the pipe. Such that if any error happens along the way of producing the result, you have a chance to replace the result with a 500 or another proper error that end user can see

tests/NuGet.Services.BasicSearchTests/TestSupport/StartedWebApp.cs
```diff
@@ -0,0 +1,76 @@
...
+            ConfigurationManager.AppSettings.Set("Local.Lucene.Directory", luceneDirectory);
+            ConfigurationManager.AppSettings.Set("Local.Data.Directory", dataDirectory);
+            ConfigurationManager.AppSettings.Set("Search.RegistrationBaseAddress", _settings.RegistrationBaseAddress);
+
+            // start the app
+            _webApp = WebApp.Start<Startup>(_portReserver.BaseUri);
+            Client = new System.Net.Http.HttpClient { BaseAddress = new Uri(_portReserver.BaseUri) };
+        }
+
+        public System.Net.Http.HttpClient Client { get; private set; }
```
yishaigalatzer: put namespaces in the using statements on top

tests/NuGet.Services.BasicSearchTests/TestSupport/StartedWebApp.cs
```diff
@@ -0,0 +1,76 @@
...
+        private TestSettings _settings;
+        private INupkgDownloader _nupkgDownloader;
+        private LuceneDirectoryInitializer _luceneDirectoryInitializer;
+        private DataDirectoryInitializer _dataDirectoryInitializer;
+        private PortReserver _portReserver;
+        private IDisposable _webApp;
+
+        public static async Task<StartedWebApp> StartAsync(IEnumerable<PackageVersion> packages = null)
+        {
+            var startedWebApp = new StartedWebApp();
```
yishaigalatzer: if you don't want it to be constructed elsewhere, make the constructor private, and the same for initialize async

src/NuGet.Indexing/DocumentCreator.cs
joelverhagen: `null` will be rejected by rejected by Lucene so adding additional validation is not necessary. `String.Empty` is an acceptable input.

tests/NuGet.Services.BasicSearchTests/TestSupport/StartedWebApp.cs
joelverhagen: This was necessary when referencing NuGet.Core because it has an HttpClient in the NuGet.* parent namespace. Per @maartenba's comment above, this full qualification is no longer necessary as this dependency has been removed.

tests/NuGet.Services.BasicSearchTests/TestSupport/StartedWebApp.cs
joelverhagen: Fixed.

src/NuGet.Indexing/PackageMetadataExtraction.cs
joelverhagen: Fixed.

src/NuGet.Indexing/PackageMetadataExtraction.cs
joelverhagen: Fixed.

src/NuGet.Indexing/DocumentCreator.cs
joelverhagen: Fixed.

maartenba:
:shipit:

joelverhagen:
This makes some progress on NuGet/NuGetGallery#2798.

PR #37, opened 2015-12-01
Extract NuGetSearchManager dependencies to improve testability

joelverhagen:
This change to the search service allows functional tests to inject in-memory implementations of dependencies that were previously static (`ConfigurationManager`) or coupled to the file system (JSON data files, Lucene index).

This also changes the functional tests to, by default, use only in-memory dependencies (with the exception of downloaded .nupkg files).

Changes verified by running the search service locally and seeing a successful build ([3.0.989](https://nugetbuild.cloudapp.net/viewLog.html?buildId=9718&tab=buildResultsDiv&buildTypeId=Gallery_MetadataService_ContinuousIntegrati)).

@johnataylor @maartenba @xavierdecoster


src/NuGet.Indexing/NuGetSearcherManager.cs
```diff
@@ -26,38 +27,65 @@ public NuGetSearcherManager(string indexName, Lucene.Net.Store.Directory directo
...
+                }
+
+                if (string.IsNullOrEmpty(dataContainer))
+                {
+                    dataContainer = "ng-search-data";
+                }
+
+                searcherManager = new NuGetSearcherManager(
+                    indexContainer,
+                    directory ?? new AzureDirectory(storageAccount, indexContainer, new RAMDirectory()),
```
maartenba: Not scoped to this commit probably, but can we try the `SimpleFSDirectory` here as well instead of `RAMDirectory`? I think our app will be as fast as it is today but consume much less memory.

maartenba:
:shipit:

src/NuGet.Indexing/NuGetSearcherManager.cs
joelverhagen: Tracked by NuGet/NuGetGallery#2796.

xavierdecoster:
:shipit:

joelverhagen:
This makes some progress on NuGet/NuGetGallery#2798.

PR #39, opened 2015-12-08
Switch to file system cache for AzureDirectory

joelverhagen:
This is the change analogous to that of [Maarten's fix to V2](https://github.com/NuGet/NuGet.Services.Metadata/commit/e3431e8d91fd248e9fca956f5774a9dad63ebaa2), but for the V3 search service.

I've investigated file system limits for the TEMP folder on Azure Websites and found no hard limits aside that the one specified for the entire machine tier. In the case of of INT, Standard tier has 50 GB. In INT, the entire Lucene database is 1.1 GB.

@maartenba, @johnataylor, @xavierdecoster


maartenba:
:shipit:

xavierdecoster:
:shipit:

PR #41, opened 2015-12-14
Searchcloud logs

xavierdecoster:
Please review so we can merge logging into consolidation branch

cc @joyhui @joelverhagen @johnataylor @maartenba 

Questions? Fire away! :)


src/NuGet.Indexing/IndexingUtils.cs
```diff
@@ -28,7 +29,7 @@ public static class IndexingUtils
                 {
                     throw;
                 }
-                Trace.TraceInformation("Unable to load {0}. Exception Message : {1}", name, e.Message);
+                logger.LogInformation("Unable to load {0}. Exception Message : {1}", name, e.Message);
```
maartenba: Stack trace could be useful to have here as well.

src/NuGet.Indexing/NuGetSearcherManager.cs
```diff
@@ -183,7 +212,7 @@ protected override NuGetIndexSearcher CreateSearcher(IndexReader reader)
                 // The point of having a specific subclass of the IndexSearcher is that we want to associate a bunch of auxilliary data along
                 // with that specific instance of the reader. The lifetimes are assocaited, hense the inheritance relationship.
 
-                Trace.TraceInformation("about to create a new NuGetIndexSearcher");
+                _logger.LogInformation("about to create a new NuGetIndexSearcher");
```
maartenba: Other logs start with capital. Perhaps for consistency make this one (and a few others) cap as well?

src/NuGet.Indexing/ServiceHelpers.cs
```diff
@@ -51,29 +52,28 @@ public static void WriteResponse(IOwinContext context, ClientException e)
         }
 
 
-        public static void WriteResponse(IOwinContext context, Exception e)
+        public static void WriteResponse(IOwinContext context, Exception e, FrameworkLogger logger)
```
joelverhagen: In general it seems this file seems tied to running in a web service, maybe it should be moved to the Search project?  The one exception that I can see is the TraceException method which seems like a) a good candidate for the Indexing project and b) useful for addressing @maartenba's comment about stack trace.

src/NuGet.Indexing/IndexingUtils.cs
joelverhagen: Maybe could use `ServiceHelpers.TraceException`?

xavierdecoster: I did not want to change this to become `LogError` (which would automatically include stack trace etc.), because the current `Trace` statement is calling `Trace*Information*`. No idea why that's the case though, happy to change it to become `logger.LogError("Unable to load {IndexName}.", name, e)`.

src/NuGet.Indexing/ServiceHelpers.cs
xavierdecoster: `ServiceHelpers` should be removed entirely from the NuGet.Indexing project IMO :)

That's expanding on the intent of this PR though. Happy to make it happen, but would prefer small, incremental and focused changes. Intent of this PR was to focus on the proposed logging *design*, and when merged into `searchcloud`, we can optimize different aspects of the code? Extracting ServiceHelpers can be done outside of the scope of this PR.

joelverhagen: Sounds good!

joelverhagen:
Really looking forward to having solid tracing!

:shipit:

src/NuGet.Indexing/IndexingUtils.cs
xavierdecoster: No one gave me a reason why this was logged as information, so I updated to log it as an error (which auto-includes stacktrace) :) 
https://github.com/NuGet/NuGet.Services.Metadata/commit/da171b4afbf2a4bbbdc79447ff9a5f6a70e85ba8

src/NuGet.Indexing/NuGetSearcherManager.cs
xavierdecoster: Fixed :)
https://github.com/NuGet/NuGet.Services.Metadata/commit/0f233cb6c677ef096f1124f7f3e0bd81f9b34d7e

maartenba: Thanks, my OCD level is back to normal.

xavierdecoster:
Addressed a few minor comments (thanks for the input :)), will squash-n-rebase when I get a second squirrel.

xavierdecoster:
cc @maartenba @joyhui @johnataylor

maartenba:
(ship-it tree)

xavierdecoster:
Merged

PR #42, opened 2015-12-15
Improve V2 compatibility

joelverhagen:
1. Detect supported frameworks from package entries as well as framework assembly groups.
2. Support the `ignoreFilter` parameter on V2 search which allows the search hijacker in the gallery to find unlisted packages.

@maartenba @joyhui @xavierdecoster @johnataylor

### DIFF
--- a/src/Ng/packages.config
+++ b/src/Ng/packages.config
@@ -12,7 +12,7 @@
   <package id="Microsoft.WindowsAzure.ConfigurationManager" version="3.1.0" targetFramework="net45" />
   <package id="Newtonsoft.Json" version="6.0.8" targetFramework="net45" />
   <package id="NuGet.Core" version="2.8.5" targetFramework="net45" />
-  <package id="nuget.versioning" version="3.2.0" targetFramework="net45" />
+  <package id="NuGet.Versioning" version="3.3.0" targetFramework="net45" />
   <package id="SharpZipLib" version="0.86.0" targetFramework="net45" />
   <package id="System.Spatial" version="5.6.4" targetFramework="net45" />
   <package id="VDS.Common" version="1.5.0" targetFramework="net45" />
--- a/src/NuGet.Indexing/Extraction/CatalogPackageMetadataExtraction.cs
+++ b/src/NuGet.Indexing/Extraction/CatalogPackageMetadataExtraction.cs
@@ -4,8 +4,11 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.Versioning;
 using System.Text;
 using Newtonsoft.Json.Linq;
+using NuGet.Frameworks;
+using NuGet.Versioning;
 
 namespace NuGet.Indexing
 {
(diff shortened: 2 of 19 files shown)


src/Ng/packages.config
```diff
@@ -11,7 +11,7 @@
   <package id="Microsoft.Web.Xdt" version="2.1.1" targetFramework="net45" />
   <package id="Microsoft.WindowsAzure.ConfigurationManager" version="3.1.0" targetFramework="net45" />
   <package id="Newtonsoft.Json" version="6.0.8" targetFramework="net45" />
-  <package id="NuGet.Core" version="2.8.5" targetFramework="net45" />
+  <package id="NuGet.Core" version="2.10.1" targetFramework="net45" />
```
joelverhagen: It seems like there is a general dislike of NuGet.Core -- is there a newer alternative I should be using for parsing target frameworks? Perhaps I can have a seperate PR as switching away from NuGet.Core like [Maarten's PR in the gallery](https://github.com/NuGet/NuGetGallery/pull/2783)?

NgTests/app.config
```diff
@@ -22,6 +22,10 @@
         <assemblyIdentity name="Microsoft.WindowsAzure.Storage" publicKeyToken="31bf3856ad364e35" culture="neutral" />
         <bindingRedirect oldVersion="0.0.0.0-4.3.0.0" newVersion="4.3.0.0" />
       </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="NuGet.Core" publicKeyToken="31bf3856ad364e35" culture="neutral" />
```
yishaigalatzer: I suggest you don't take new dependencies on nuget.core

joelverhagen: Sure, I'll look into switching this over.

src/Ng/packages.config
maartenba: Could be good, I believe the plan is to deprecate NuGet.Core somewhere in the future so being on the new packages certainly won't hurt.

maartenba:
:shipit:

src/Ng/packages.config
```diff
@@ -11,7 +11,7 @@
...
   <package id="nuget.versioning" version="3.2.0" targetFramework="net45" />
   <package id="SharpZipLib" version="0.86.0" targetFramework="net45" />
```
yishaigalatzer: Consider moving the project off packages.config to project.json (still with csproj) that should work if the project doesn't use any content

src/NuGet.Indexing/Extraction/CatalogPackageMetadataExtraction.cs
```diff
@@ -189,29 +195,44 @@ private void AddSupportedFrameworks()
...
-                                builder.Append("|");
-                            }
-
-                            builder.Append(shortFrameworkName);
-                        }
+                        yield return VersionUtility.ParseFrameworkName(frameworkName);
                     }
                 }
+            }
+            private IEnumerable<FrameworkName> GetFrameworksFromPaths()
```
yishaigalatzer: nit: new line between the two methods

NgTests/app.config
joelverhagen: Fixed.

src/NuGet.Indexing/Extraction/CatalogPackageMetadataExtraction.cs
joelverhagen: Fixed.

src/Ng/packages.config
joelverhagen: Out of scope for this PR. Tracked by NuGet/NuGetGallery#2813.

PR #44, opened 2015-12-18
Fix Web.Release.config

joelverhagen:
This was causing an error on app startup. The `<monitoring>` tag should be under `<system.webServer>`, not `<system.web>`.


xavierdecoster:
:shipit:

maartenba:
Shipit

PR #45, opened 2015-12-29
Increase Lucene index refresh frequency.

joelverhagen:
It was 10 minutes. Now it's 30 seconds. Also added better logging around index reloading.

@maartenba @joyhui @johnataylor

### DIFF
--- a/src/NuGet.Services.BasicSearch/Startup.cs
+++ b/src/NuGet.Services.BasicSearch/Startup.cs
@@ -8,6 +8,7 @@ using Microsoft.Owin.StaticFiles.Infrastructure;
 using NuGet.Indexing;
 using Owin;
 using System;
+using System.Diagnostics;
 using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


maartenba:
:shipit: - note we want to monitor this closely after launch

yishaigalatzer:
@joelverhagen do you have any data indicating that there is no possible stress from this change?

joelverhagen:
The testing I did on DEV focused on improving how quickly new packages show up in the search service. One indicative metric I have is the average time to complete request to the search service, against DEV and using an exact ID query (1 or zero search results). My test polled the search service on three threads, sleeping 250ms between each attempt.

Before the change, the average time to complete the request was 103 ms with packages showing up in searches after 7 minutes, 51 seconds. After the change, the average request time was 115 ms and packages showed up in 1 minute, 53 seconds.

Is this the kind of data you are looking for? In my mind, the minor uptick in request time is worth the 6 minute increase in search availability.

yishaigalatzer:
That's actually a pretty large number, particularly because you don't run in production.

The way I would make this change is

1. Deploy the logging into production
2. Get data on CPU, memory and GC during the update.
3. Get data on how long the update takes.

Based on the data, decide how to make changes in production, and when we do, make small incremental changes and monitor (so to go down from 10 minutes to 30 sec might take a few days), but will reduce the risk of it blowing out the CPU or memory and causing a downtime

joelverhagen:
Sounds good. I'll revert the web.config change. Thanks!

PR #46, opened 2015-12-31
Add Application Initialization for search web site

maartenba:
This one preloads the `/search/diag` endpoint, which in turn reads the entire index.

Do we need other URL's hit during warmup? I figure loading the full index is probably the only thing we want to do here, which should work with this change.

### DIFF
--- a/src/NuGet.Services.BasicSearch/Web.config
+++ b/src/NuGet.Services.BasicSearch/Web.config
@@ -32,6 +32,9 @@
     <!-- END Serilog logging configuration -->
   </appSettings>
   <system.webServer>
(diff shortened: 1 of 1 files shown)


yishaigalatzer:
How much this delay the startup? The question is about balancing cost of service startup (which is interesting when we swap recover or scale) vs perhaps slower calls for some packages

joyhui:
I assume the load balance won't put this instance into loop until the call to this URL finished. 
Please correct me if I am wrong.

maartenba:
We'll have to try once Antares supports it.

xavierdecoster:
:shipit:

PR #47, opened 2016-01-08
In progress - Improve catalog to registration performance

joelverhagen:
1. Don't read the leaf nodes of the registration, only reading the index.json (and pages on large registration) is necessary.
2. Only write new registration leaf blobs. With this change, the new (that is added or updated) leaf nodes are updates along with the page and index.json blobs.

### DIFF
--- a/NgTests/RegistrationCollectorTests.cs
+++ b/NgTests/RegistrationCollectorTests.cs
@@ -35,8 +35,7 @@ namespace NgTests
             // Setup collector
             var target = new RegistrationCollector(new Uri("http://tempuri.org/index.json"), catalogToRegistrationStorageFactory, () => mockServer)
             {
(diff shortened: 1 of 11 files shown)


maartenba:
@joelverhagen What's the status on this one?

joelverhagen:
I verified the fix manually, but the `searchcloud` branch has moved forward significantly since then so the code should be verified again. Essentially, I ensured that the registration blobs generated from the catalog for a big package (e.g. `Catel.MVVM`) are the equivalent before and after my change.

@johnataylor, you looked at this code as well, right?

maartenba:
Going to close this one in favor of https://github.com/NuGet/NuGet.Services.Metadata/pull/59

joelverhagen:
Thanks for taking care of this.

PR #48, opened 2016-01-14
remove obsolete scripts

xavierdecoster:
Think we can safely remove these...

@maartenba @johnataylor @joelverhagen @joyhui


maartenba:
Yup

xavierdecoster:
Fixed with https://github.com/NuGet/NuGet.Services.Metadata/commit/64b1eedec45e1113e28f3fda9bc44b7b54c3962e

PR #49, opened 2016-01-15
Add "lightning" tool to ng.exe

maartenba:
New "lightning" tool for fast registration creation.


src/Ng/Lightning.cs
```diff
@@ -0,0 +1,397 @@
...
+            {
+                Directory.CreateDirectory(outputFolder);
+            }
+
+            // Create reindex file
+            log.WriteLine("Start preparing lightning reindex file...");
+
+            var latestCommit = DateTime.MinValue;
+            int numberOfEntries = 0;
+            string indexFile = Path.Combine(outputFolder, "index.txt");
```
johnataylor: You are essentially introducing an alternative "index" structure to the catalog's existing index structure. We could consolidate on a single structure this should [theoretically] reduce the overall complexity because we'd just have one model for iteration, one model for cursors etc.

However, I don't think there is an urgent need for this consolidation. So not really a code-review comment but something we should discuss and experiment with perhaps. In the process of doing that perhaps we could also look at a couple of other things in the catalog - for example the read batch size could be independent of the write batch commit.

src/Ng/Lightning.cs
```diff
@@ -0,0 +1,397 @@
...
+                if (batchEnd >= numberOfEntries)
+                {
+                    batchEnd = numberOfEntries - 1;
+                }
+
+                var cursorCommandFileName = "cursor" + batchNumber + ".cmd";
+                var cursorTextFileName = "cursor" + batchNumber + ".txt";
+
+                using (var cursorCommandStreamWriter = new StreamWriter(Path.Combine(outputFolder, cursorCommandFileName)))
+                using (var cursorTextStreamWriter = new StreamWriter(Path.Combine(outputFolder, cursorTextFileName)))
```
johnataylor: how about we always include { } even after "using" I realize this moves the code to the right but still clearer in my mind.

src/Ng/Lightning.cs
```diff
@@ -0,0 +1,397 @@
...
+                streamWriter.Write(",");
+                streamWriter.Write(endIndex);
+            }
+
+            return Task.FromResult(true);
+        }
+
+        private static Task ProcessGraphsAsync(string packageId, IDictionary<string, IGraph> sortedGraphs, StorageFactory storageFactory, string contentBaseAddress)
+        {
+            return RegistrationCatalogCreator.ProcessGraphs(packageId, sortedGraphs, storageFactory, new Uri(contentBaseAddress), 64, 128, CancellationToken.None);
```
johnataylor: I think this is a different implementation of Registration creation than is happening in the incremental job. The incremental job uses RegistrationCollector that delegates to RegistrationMaker. I'd rewritten as part of the apiapps push because I got the feedback that this code was hard to understand. However, the code in RegistrationMaker has some inefficiencies - Joel fixed the worst of these before the break.

src/Ng/Lightning.cs
maartenba: This one is just internal for the tool, basically doing the same as our first/previous tool did. Not sure we need to standardize this?

src/Ng/Lightning.cs
maartenba: Updated.

src/Ng/Lightning.cs
maartenba: Updated to use `RegistrationMaker` instead.

maartenba:
Feel free to let me know when to : shipit:  :-)

maartenba:
Ran it for the V2V3 containers, results look very good. This PR is ready for prime time :)
@johnataylor @RanjiniM @joyhui @joelverhagen can I get some :shipit: ?

maartenba:
Since the V2V3 containers have been built succesfully with this tool, merging it.

PR #50, opened 2016-01-15
Hook up Serilog logging into ng.exe jobs

xavierdecoster:
The scope of this PR is to wire up the new logging infrastructure with the ng.exe based jobs.
Improving actual logging coverage (e.g. in `Catalog2Registration`, `Catalog2Dnx` and `Catalog2Lucene`) will be done in a separate PR depending on this one.

@joyhui @johnataylor @joelverhagen @maartenba

### DIFF
--- a/src/Ng/Scripts/Feed2CatalogV3-Secondary.cmd
+++ b/src/Ng/Scripts/Feed2CatalogV3-Secondary.cmd
@@ -10,7 +10,7 @@ cd Ng
 
 	title #{Jobs.feed2catalogv3secondary.Title}
 
-    start /w .\Ng.exe feed2catalog -gallery #{Jobs.common.v3.f2c.Gallery} -storageBaseAddress #{Jobs.feed2catalogv3secondary.StorageBaseAddress} -storageType azure -storageAccountName #{Jobs.common.v3.Storage.Secondary.Name} -storageKeyValue #{Jobs.common.v3.Storage.Secondary.Key} -storageContainer #{Jobs.feed2catalogv3secondary.StorageContainer} -storageTypeAuditing azure -storageAccountNameAuditing #{Jobs.common.v3.c2r.StorageAccountName} -storageKeyValueAuditing #{Jobs.common.v3.c2r.StorageAccountKey} -storageContainerAuditing auditing -storagePathAuditing package -verbose true -interval #{Jobs.common.v3.Interval}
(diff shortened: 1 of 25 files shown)


src/Ng/Scripts/Feed2CatalogV3-Secondary.cmd
```diff
@@ -10,7 +10,7 @@ cd Ng
...
+    start /w .\Ng.exe feed2catalog -gallery #{Jobs.common.v3.f2c.Gallery} -storageBaseAddress #{Jobs.feed2catalogv3secondary.StorageBaseAddress} -storageType azure -storageAccountName #{Jobs.common.v3.Storage.Secondary.Name} -storageKeyValue #{Jobs.common.v3.Storage.Secondary.Key} -storageContainer #{Jobs.feed2catalogv3secondary.StorageContainer} -storageTypeAuditing azure -storageAccountNameAuditing #{Jobs.common.v3.c2r.StorageAccountName} -storageKeyValueAuditing #{Jobs.common.v3.c2r.StorageAccountKey} -storageContainerAuditing auditing -storagePathAuditing package -elasticsearchendpoint #{Jobs.common.v3.Logging.ElasticsearchEndpoint} -verbose true -interval #{Jobs.common.v3.Interval}
```
maartenba: Are the new config values also added to Octopus Deploy?

maartenba:
Looks good, put it on the boat.
:shipit:

src/Ng/Scripts/Feed2CatalogV3-Secondary.cmd
xavierdecoster: Yes :)

xavierdecoster:
Keep those squirrels coming! :)

src/Ng/Logging.cs
```diff
@@ -0,0 +1,244 @@
...
+using System.Runtime.Remoting.Messaging;
+using Microsoft.Extensions.Logging;
+using Serilog;
+using Serilog.Core;
+using Serilog.Events;
+using ILogger = Microsoft.Extensions.Logging.ILogger;
+
+namespace Ng
+{
+    internal static class Logging
```
joelverhagen: How similar is this to the logging infrastructure in the search service? Could the two share code somehow?

joelverhagen:
Looks good. I especially like the critical logging in the case of unhandled exceptions. @maartenba, would this have covered the case where the catalog2lucene job was crashing due to the C5 package's invalid lib folder name?
:shipit:

maartenba:
Looks like it would have, although the job would still have crashed

src/Ng/Logging.cs
xavierdecoster: Both svc and jobs now share exact same logging code, fixed in https://github.com/NuGet/NuGet.Services.Metadata/commit/e5479a79105e4e73016730dcefddf035071a4e31 which also adds basic auth support against the ES cluster.

PR #51, opened 2016-01-20
support basic auth in ES logging sink

xavierdecoster:
Tweaked ES logging configuration to support multiple node uris and basic auth credentials

@maartenba @joyhui


maartenba:
:shipit:

joyhui:
:shipit: 

One questions, how can we know the logging is working?  We may want to introduce a ping logic in our code, which write a special verbal event every 1 min (for example), and we could monitor it to make sure logging in all services/roles/web/jobs are working.

PR #53, opened 2016-02-05
Fix - Potentially multiple indices are read at the same time

maartenba:
/cc @joyhui @johnataylor @skofman1 @xavierdecoster


xavierdecoster:
:shipit:

johnataylor:
But the first thing the timer callback does is increment the _gate in the process of checking it (it reads and increments it as an atomic operation.) Having incremented it, it must decrement it even if it doesn’t allow the action to be taken, and of course if the action is taken the decrement should happen.

However, if MaybeReopen does throw this trailing decrement will be skipped and so the work will never be attempted again – by any thread for the lifetime of this process – so the behavior would be the opposite of the suggested bug. So looks to like we do need to add the finally but I don’t think you can just swap the increment for a read.

maartenba:
Pushed a fix for the read.

johnataylor:
:shipit:

PR #54, opened 2016-02-09
Boost a bit more on download rankings even if id is a full match

maartenba:
For example for q=json, Newtonsoft.Json now is first in the list, while package id "json" is second.


maartenba:
@johnataylor @joelverhagen @xavierdecoster @joyhui  Please review

johnataylor:
:shipit:
 (so does this change move newtonjson to the top result for "json"? because we think that is currently the "most relevant" result.)

maartenba:
It does. Not by much but it outranks "id:json" :-)

PR #55, opened 2016-02-10
Reduce the default segment size in lucene index

maartenba:
Fixes https://github.com/NuGet/NuGetGallery/issues/2445


maartenba:
/cc @johnataylor 

This is a potential fix that should get us to segment sizes of roughly 100 MB max. in the coming months.

johnataylor:
So you found there is a difference between setting the whole policy object vs setting values on the writer’s existing policy?

Because we did see this settings on the writer work just fine at some stage. This maybe coincidental but, when we started to do upserts (i.e. delete and add) on the index the merge factor seemed to get ignored. So did you test that?

Or was it the addition of:
                MinMergeMB = MinMergeMB,  
                MaxMergeMB = MaxMergeMB  

We had not set these before because the other settings were working?

johnataylor:
Oh yes, :shipit: regarding code review if this actually works following your testing tomorrow!

maartenba:
Day 2 of running this on my machine (have added a few extra delete + add operations in the code to do some IKEA-style durability testing) - so far biggest segment is still 45 MB. Will merge this when the good behaviour continues for a third day in row.

maartenba:
Another day in paradise! No big segments and stable sizes. I will give it until after the weekend but I feel we may have tackled the segments.

maartenba:
That's almost a week of good behaviour. Merging this one in. (we will need to reindex before go-live to reduce segments to start off)

PR #56, opened 2016-02-23
Added ability for catalog2registration to create compressed registrations

johnataylor:
added ability for catalog2registration to create a parallel set of compressed blobs.

NOTE - @maartenba this is a preliminary review - the lower level parts I've tested -i.e. the additions to the Azure storage layer.

However, the command line driver is not fully tested - probably a couple of hours tomorrow on testing that piece.

This turned out a little more complex (at runtime) than I'd hoped. The problem is the registration blob logic is too tied to the absolute. Ideally the whole logic could be based on relative addresses. Its only partially based on relative addresses.(For example, if we had the storage layer understand JSON-LD just a little it could write absolute paths when we give it relative - the client is actually expecting absolute right now.)


maartenba:
Looks good, indeed needs some more wiring.

Regarding the relative addresses, I'm not sure this "magic" of making URLs absolute should be in the lower level API. Perhaps the "aggregate storage" (that writes to both compressed and uncompressed) could have this logic and replace base URLs in every @id property of the JSON?

johnataylor:
Note: this is not required for this immediate fix.

But regarding the relative/absolute URLs. Basically what I was imagining was a layer above the raw storage that understood JSON-LD. For example, if you set an @base in the document as it wrote out the JSON it honored that and made @id relative to that. Or perhaps you could also support the opposite, have all URLs in the document relative and then as you saved them specify a base address that would be applied across them all.

I think it would be quiet appropriate to have these type of transformation embedded in a JSON readers and writers. The code can be pretty simple, for example the following removes nodes that match a particular Json Path. 

https://github.com/NuGet/Entropy/blob/master/MinimizerWebApplication/JsonSkipReader.cs

src/Ng/Catalog2Registration.cs
```diff
@@ -91,7 +96,68 @@ public void Run(string[] args, CancellationToken cancellationToken)
...
+        }
+
+        private void InterceptRegistrationIndex(Uri primaryResourceUri, Uri secondaryResourceUri, JToken content)
+        {
+            var tokens = content.SelectTokens("@id")
+                .Union(content.SelectTokens("items[*].@id"))
+                .Union(content.SelectTokens("items[*].catalogEntry.@id"))
+                .Union(content.SelectTokens("items[*].registration"))
+                .Union(content.SelectTokens("items[*].parent"))
+                .Union(content.SelectTokens("items.items[*].@id"))
```
johnataylor: wouldn't this be items[*].items[*].@id ? in other words wouldn't the first items also be an array reference?

src/Ng/Catalog2Registration.cs
```diff
@@ -91,7 +96,68 @@ public void Run(string[] args, CancellationToken cancellationToken)
...
+                tokens[i].Replace(
+                    ((string)tokens[i]).Replace(primaryResourceUri.ToString(), secondaryResourceUri.ToString()));
+            }
+        }
+
+        private void InterceptRegistrationIndex(Uri primaryResourceUri, Uri secondaryResourceUri, JToken content)
+        {
+            var tokens = content.SelectTokens("@id")
+                .Union(content.SelectTokens("items[*].@id"))
+                .Union(content.SelectTokens("items[*].catalogEntry.@id"))
```
johnataylor: we shouldn't have to rewrite the @id pointing to the Catalog Entry - that should remain constant right?
we are just hacking the @id within the scope of the registration blobs.

src/Catalog/Helpers/Utils.cs
```diff
@@ -166,28 +166,7 @@ public static JToken CreateJson2(IGraph graph, JToken frame = null)
                 return JsonSort.OrderJson(compacted);
             }
         }
-
-        public static string CreateJson(IGraph graph, JToken frame = null)
```
johnataylor: this seems like a confusing diff to me - the raw file looks OK.

maartenba:
@johnataylor Made a big update to this, feel free to have a look (it currently powers our v2v3 registrations)

If shipworthy, bring out those :shipit:! :smile:

src/Ng/Json/InterceptingJsonReader.cs
```diff
@@ -0,0 +1,119 @@
...
+        private static Regex MakeRegex(string jsonPath)
+        {
+            Regex expression = null;
+            if (!CachedRegexes.TryGetValue(jsonPath, out expression))
+            {
+                var pattern = jsonPath
+                    .Replace("[*]", @"\[\d+\]")
+                    .Replace(".", @"\.");
+
+                expression = new Regex("^" + pattern);
```
johnataylor: Just discussed this with Joel - and he points out that we should have closed this one down with a $ i.e. something like:
     expression = new Regex("^" + pattern + "$");

(Though we haven't tested that one.)

The reason we think this hasn't mattered is that the paths we happen to have all terminating - precisely to the property in question. In other words we think this works as is but is a little fragile.

johnataylor:
:shipit:

src/Ng/Json/InterceptingJsonReader.cs
maartenba: Good catch! Updated.

PR #57, opened 2016-03-02
Add response output caching to search service

maartenba:
* Memory limit is currently set to 100 MB (with a maximum of 10% of system memory in case there is memory pressure)
* Cache can easily be disabled: set Search.EnableResponseCaching setting to “false” in the portal
* I added the X-Cache: HIT header when a cache was hit – this is the same header the CDN provides us.
* Cache hit ratio can be viewed from /cache/diag. Note that this is intentionally not correct (hits may be missed as I did not add any locking mechanism to the hit counting to never block the request path)

It is quite opinionated for our scenario. Also, not sure if my "ShouldCache()" check is correct (e.g. we may want to exclude caches with cookies or certain header values)

@joyhui @johnataylor @xavierdecoster @skofman1 Please have a look

### DIFF
--- a/src/NuGet.Services.BasicSearch/ApplicationInsights.config
+++ b/src/NuGet.Services.BasicSearch/ApplicationInsights.config
@@ -11,8 +11,8 @@
 <Add Type="Microsoft.ApplicationInsights.Extensibility.Web.WebApplicationLifecycleModule, Microsoft.ApplicationInsights.Extensibility.Web"/>
 <Add Type="Microsoft.ApplicationInsights.Extensibility.Web.RequestTracking.TelemetryModules.WebRequestTrackingTelemetryModule, Microsoft.ApplicationInsights.Extensibility.Web"/>
 <Add Type="Microsoft.ApplicationInsights.Extensibility.Web.RequestTracking.TelemetryModules.WebExceptionTrackingTelemetryModule, Microsoft.ApplicationInsights.Extensibility.Web"/>
-<Add Type="Microsoft.ApplicationInsights.Extensibility.Web.RequestTracking.TelemetryModules.WebSessionTrackingTelemetryModule, Microsoft.ApplicationInsights.Extensibility.Web"/>
-<Add Type="Microsoft.ApplicationInsights.Extensibility.Web.RequestTracking.TelemetryModules.WebUserTrackingTelemetryModule, Microsoft.ApplicationInsights.Extensibility.Web"/>
+<!--<Add Type="Microsoft.ApplicationInsights.Extensibility.Web.RequestTracking.TelemetryModules.WebSessionTrackingTelemetryModule, Microsoft.ApplicationInsights.Extensibility.Web"/>
+<Add Type="Microsoft.ApplicationInsights.Extensibility.Web.RequestTracking.TelemetryModules.WebUserTrackingTelemetryModule, Microsoft.ApplicationInsights.Extensibility.Web"/>-->
 </TelemetryModules>
 <ContextInitializers>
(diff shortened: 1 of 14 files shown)


src/NuGet.Services.BasicSearch/Caching/MemoryCacheResponseBodyCache.cs
```diff
@@ -0,0 +1,108 @@
...
+                    contentBytes = Get(request);
+                    if (contentBytes != null)
+                    {
+                        Hits++;
+                        return true;
+                    }
+                }
+                catch
+                {
+                    // Ignore - we will assume no cache if this happens
```
skofman1: Log?

src/NuGet.Services.BasicSearch/ResponseHelpers.cs
```diff
@@ -8,38 +11,75 @@
...
+                 * exceptions. Another approach could also separate "do business logic" with "build response body". Since
+                 * most exceptions will likely happen during the "do business logic" step, this would reduce the change of
+                 * a 200 OK on service exception. However, this means that the whole result of the business logic is in
+                 * memory.
+                 */
+                content = new MemoryStream();
+                WriteToStream(content, writeContent);
+                content.Position = 0;
+
+                ResponseBodyCache.Add(context.Request, content.ToArray());
```
skofman1: do u write to cache even if allowResponseCache==false?

maartenba: Fixed, good catch!

maartenba:
Also made `ResponseHelpers` a non-static class so we can pass it around freely.

xavierdecoster:
@maartenba :+1: for the `ResponseHelpers`

src/NuGet.Services.BasicSearch/ApplicationInsights.config
```diff
@@ -11,8 +11,8 @@
 <Add Type="Microsoft.ApplicationInsights.Extensibility.Web.WebApplicationLifecycleModule, Microsoft.ApplicationInsights.Extensibility.Web"/>
 <Add Type="Microsoft.ApplicationInsights.Extensibility.Web.RequestTracking.TelemetryModules.WebRequestTrackingTelemetryModule, Microsoft.ApplicationInsights.Extensibility.Web"/>
 <Add Type="Microsoft.ApplicationInsights.Extensibility.Web.RequestTracking.TelemetryModules.WebExceptionTrackingTelemetryModule, Microsoft.ApplicationInsights.Extensibility.Web"/>
-<Add Type="Microsoft.ApplicationInsights.Extensibility.Web.RequestTracking.TelemetryModules.WebSessionTrackingTelemetryModule, Microsoft.ApplicationInsights.Extensibility.Web"/>
-<Add Type="Microsoft.ApplicationInsights.Extensibility.Web.RequestTracking.TelemetryModules.WebUserTrackingTelemetryModule, Microsoft.ApplicationInsights.Extensibility.Web"/>
+<!--<Add Type="Microsoft.ApplicationInsights.Extensibility.Web.RequestTracking.TelemetryModules.WebSessionTrackingTelemetryModule, Microsoft.ApplicationInsights.Extensibility.Web"/>
```
xavierdecoster: Yank 'm if we don't need them. Or maybe explain why we don't want them in the comments (I suspect cookies vs output caching?).

maartenba: Exactly, removed these these two to skip the cookies.

xavierdecoster:
Looks good to me!

PR #58, opened 2016-03-03
Refactor CatalogToDnx

skofman1:
Extract a class that save a package and a nuspec in flat container format


skofman1:
@johnataylor , @maartenba , @xavierdecoster :cake:

maartenba:
Yes! :shipit:

By means of housekeeping, can you also add a test (one for add, one for delete, one for delete all version to check no data is left behind) for it? (there is a MemoryStorage class you can use to fake storage so you can check which data was written (or not))

There are no tests for this currently so any test is better than none :-)

src/Catalog/Dnx/DnxCatalogCollector.cs
```diff
@@ -0,0 +1,121 @@
...
+                Trace.TraceInformation("package not found...");
+            }
+            else
+            {
+                httpResponseMessage.EnsureSuccessStatusCode();
+            }
+            return null;
+        }
+
+        private static string GetNuspec(Stream stream, string id)
```
xavierdecoster: I'm wondering if there's client code we could leverage here that may do the exact same thing? @emgarten

emgarten: That would be great to have the same logic in the server, use *PackageArchiveReader*
https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Packaging/PackageReaderBase.cs#L83

As you can see in the above method the client does not allow packages that have multiple nuspecs, but here the server just takes the first one and goes with it. I would expect the gallery to be even more strict on this. Bad packages can still come from other sources, but nuget.org shouldn't allow users to upload packages that cannot work.

I think users would appreciate the extra checks, I don't think anyone really wants to upload a package that gets indexed in this way since if the server handles it wrong they then cannot delete it.

skofman1: I added an issue to track this suggestion since it shouldn't be part of this work: https://github.com/NuGet/NuGetGallery/issues/2921

src/Catalog/Dnx/DnxCatalogCollector.cs
```diff
@@ -0,0 +1,121 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Newtonsoft.Json.Linq;
+using NuGet.Services.Metadata.Catalog.Persistence;
+
+namespace NuGet.Services.Metadata.Catalog.Dnx
+{
+    public class DnxCatalogCollector : CommitCollector
+    {
+        private DnxMaker _dnxMaker;
```
johnataylor: This is obviously private - none of the server code is like this.

src/Catalog/Dnx/DnxCatalogCollector.cs
skofman1: How about this test:
https://github.com/NuGet/NuGet.Services.Metadata/blob/master/NgTests/DnxCatalogCollectorTests.cs
What else should be covered?

src/Catalog/Dnx/DnxMaker.cs
```diff
@@ -0,0 +1,132 @@
...
+
+namespace NuGet.Services.Metadata.Catalog.Dnx
+{
+    public class DnxMaker
+    {
+        private readonly StorageFactory _storageFactory;
+
+        public DnxMaker(StorageFactory storageFactory)
+        {
+            if (storageFactory == null)
```
johnataylor: this is noise

src/Catalog/Dnx/DnxMaker.cs
```diff
@@ -0,0 +1,132 @@
...
+                if (storage.Exists(relativeAddress))
+                {
+                    await storage.Delete(resourceUri, cancellationToken);
+                }
+            }
+        }
+
+        private static HashSet<NuGetVersion> GetVersions(string json)
+        {
+            HashSet<NuGetVersion> result = new HashSet<NuGetVersion>();
```
johnataylor: its OK to use var in a narrow scope like this with an explicit assignment - there is no loss of readability

src/Catalog/Dnx/DnxCatalogCollector.cs
maartenba: Whoa! Did not even know this existed :-) Ignore my test remark then.

johnataylor:
:shipit:

skofman1:
Pushed

PR #59, opened 2016-03-22
Improve catalog2registration processing performance

maartenba:
Given a package that has 1000 version leafs + an index.json, the current catalog2registration process will read all 1000 version leafs + index.json, and writes them all again as well. Adding a package thus means 2x 1000 + 2x index.json + 1x new package I/O operations.

This PR reduces the I/O to 2x index.json + 1x new package I/O operations:

1. For each registration, only read index.json (and pages on large
registration). Leaf nodes are not read.
2. Only write new registration leaf blobs (don't re-write leafs that
already exist as it's a waste of effort).

Based on https://github.com/NuGet/NuGet.Services.Metadata/pull/47

### DIFF
--- a/src/Catalog/Registration/RegistrationMakerCatalogItem.cs
+++ b/src/Catalog/Registration/RegistrationMakerCatalogItem.cs
@@ -22,7 +22,6 @@ namespace NuGet.Services.Metadata.Catalog.Registration
         private Uri _registrationAddress;
         private DateTime _publishedDate;
         private bool _listed;
-        private readonly bool _isExistingItem;
 
         public RegistrationMakerCatalogItem(Uri catalogUri, IGraph catalogItem, Uri registrationBaseAddress, bool isExistingItem, Uri packageContentBaseAddress = null)
         {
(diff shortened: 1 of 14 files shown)


maartenba:
The magic is in https://github.com/NuGet/NuGet.Services.Metadata/pull/59/files#diff-58b5390f3179863436ad1915f9c01f7bR40 - returning null for content that already exists means the existing file is left alone.

Please review @skofman1 @xavierdecoster @joelverhagen @johnataylor

src/Catalog/Registration/RegistrationMakerCatalogItem.cs
```diff
@@ -14,30 +13,37 @@ namespace NuGet.Services.Metadata.Catalog.Registration
...
+            _isExistingItem = isExistingItem;
         }
 
         public override StorageContent CreateContent(CatalogContext context)
         {
+            if (_isExistingItem)
+            {
+                return null;
+            }
+
```
skofman1: This behavior is pretty counter intuitive, and makes the code more complex. Is it possible to make a change in a higher layer?

maartenba: :+1:

joelverhagen:
:shipit:

maartenba:
Reworked this a bit. The magic potion is now a decision for the registration maker to make - https://github.com/NuGet/NuGet.Services.Metadata/pull/59/files#diff-4ed2d1d2e3a3fa5104e200c1b6d39ec2R81

xavierdecoster:
:shipit:

skofman1:
:ship:

maartenba:
Thanks!

PR #62, opened 2016-03-24
Fixed test and test run in VS

skofman1:
@maartenba , @xavierdecoster


maartenba:
Retroactive :shipit: ;-)

PR #64, opened 2016-04-14
Consolidated search improvements

maartenba:
Apologies for making this a big commit. This has been an ongoing effort where all of the below sort of link together and were very hard to commit separately.

**Technical changes**

* Introduce cloud service (instead of using Antares)
* Install IIS AppInit on the web role and make sure the index is warm before the machine gets into load balancer

**Performance improvements**

* Remove all user code lock statements where possible
* AzureDirectory is used as source store, which is synced with an in-memory index every 5 minutes (no more waits for locks from AzureDirectory)
* Index load time on cold start ~2 minutes, warm start (with a cache on disk) ~1 minute
* Rewrite V2 "id:xyz" queries that originate from the NuGet Gallery to "packageid:xyz" to increase matching performance
* Rewrite ranking result scoring to use a map of Lucene document id's (array index lookup vs. dictionary find)
* Rewrite package owners to be merged in search results based on a simple dictionary instead of indexing them on the fly (reduced startup time by ~5 minutes)
* Rewrite queries on package owners to use a filter (reduces the document result set before running tehactual query, improving query performance)
* String interning for package id's and package versions to reduce GC overhead (also reduced managed memory footprint by ~100 MB)
* After the above changes, the response cache we introduced earlier became the bottleneck - so I removed it

**Functional improvements**

* V3 compatibility (don't include unlisted packages in result by default)
* V2 compatibility (don't filter listed/unlisted and prerelease as current V2 also does not do this)

**Remarks**

* The unit tests are still failing (intentional) - I have to redo these as the shape of our queries has changed by this commit. Want to first gather feedback on the PR before fixing the tests and re-doing the tests.
* Performance profiling on an Azure D3v2 machine shows that this code can take 1.700 requests per second on one machine. Current V2 search handles ~500/sec on average, on 6 machines.
* I expect us to have to run on 4 machines in cloud services for redundancy + ensuring spare capacity.

@skofman1 @xavierdecoster @joelverhagen @johnataylor Care to review? :-)


src/NuGet.Services.BasicSearch/WebRole.cs
```diff
@@ -0,0 +1,60 @@
...
+                {
+                    using (var reader = new StreamReader(response.GetResponseStream()))
+                    {
+                        reader.ReadToEnd();
+                    }
+                }
+            }
+            catch
+            {
+                // ignore
```
skofman1: log?

src/NuGet.Indexing/GalleryServiceImpl.cs
```diff
@@ -25,22 +25,49 @@ public static void Search(JsonWriter jsonWriter, NuGetSearcherManager searcherMa
             var searcher = searcherManager.Get();
             try
             {
-                bool includeUnlisted = ignoreFilter;
-                includePrerelease = ignoreFilter || includePrerelease;
+                // The old V2 search service would tread "id:" queries (~match) in the same way as it did "packageid:" (==match).
```
skofman1: treat

src/NuGet.Indexing/GalleryServiceImpl.cs
```diff
@@ -25,22 +25,49 @@ public static void Search(JsonWriter jsonWriter, NuGetSearcherManager searcherMa
...
+                // If "id:" is in the query, replace it.
+                if (luceneQuery && !string.IsNullOrEmpty(q) && q.StartsWith("id:", StringComparison.OrdinalIgnoreCase))
+                {
+                    q = "packageid:" + q.Substring(3);
```
skofman1: any reason not to use string.Replace?

src/NuGet.Indexing/GalleryServiceImpl.cs
```diff
@@ -25,22 +25,49 @@ public static void Search(JsonWriter jsonWriter, NuGetSearcherManager searcherMa
...
+                // the data you'd expect looking at the method parameters:
+                //   bool includeUnlisted = ignoreFilter;
+                //   includePrerelease = ignoreFilter || includePrerelease;
+                //   feed = ignoreFilter ? null : feed;
+                //   Filter filter = searcher.GetFilter(includeUnlisted, includePrerelease, feed);
+                //
+                // But hey, we'll do it the old V2 way instead and happily ignore the correct way.
                 feed = ignoreFilter ? null : feed;
-                Filter filter = searcher.GetFilter(includeUnlisted, includePrerelease, feed);
+                Filter filter = searcher.GetFilter(feed);
```
skofman1: it confusing that you pass a null to GetFilter, and expect it to return a correct value. A clearer approach would be:
Filter filter = ignoreFilter ? null : searcher.GetFilter(feed);

src/NuGet.Indexing/GalleryServiceImpl.cs
```diff
@@ -25,22 +25,49 @@ public static void Search(JsonWriter jsonWriter, NuGetSearcherManager searcherMa
...
 
-                Query query = NuGetQuery.MakeQuery(q);
+                Query query = NuGetQuery.MakeQuery(q, searcher);
+
+                if (filter != null)
+                {
+                    query = new FilteredQuery(query, filter);
+                    filter = null;
```
skofman1: why?

src/NuGet.Indexing/GalleryServiceImpl.cs
```diff
@@ -25,22 +25,49 @@ public static void Search(JsonWriter jsonWriter, NuGetSearcherManager searcherMa
...
+                }
 
                 if (countOnly)
                 {
                     DocumentCountImpl(jsonWriter, searcher, query, filter);
```
skofman1: you always pass null here, so just pass null explicitly

src/NuGet.Indexing/GalleryServiceImpl.cs
```diff
@@ -25,22 +25,49 @@ public static void Search(JsonWriter jsonWriter, NuGetSearcherManager searcherMa
...
                 }
                 else
                 {
-                    IDictionary<string, int> rankings = searcher.Rankings;
-
-                    ListDocumentsImpl(jsonWriter, searcher, query, rankings, filter, sortBy, skip, take, searcherManager);
+                    ListDocumentsImpl(jsonWriter, searcher, query, filter, sortBy, skip, take, searcherManager);
```
skofman1: same, just pass null explicitly

src/NuGet.Services.BasicSearch/WebRole.cs
maartenba: not important here, it's mainly to do a first request to the local host when just started

src/NuGet.Indexing/RankingsScoreQuery.cs
```diff
@@ -39,41 +38,54 @@ public override string ToString()
...
             {
                 _rankings = rankings;
-                _ids = FieldCache_Fields.DEFAULT.GetStrings(reader, "Id");
                 _baseBoost = baseBoost;
+
+                SegmentReader segmentReader = reader as SegmentReader;
+
+                _readerName = segmentReader != null
+                    ? segmentReader.SegmentName
+                    : string.Empty;
```
skofman1: this is unclear. Can you please add a comment explaining the behavior here?

skofman1: And in general, can you add some documentation to this file..

---
In reply to: [59756131](https://github.com/NuGet/NuGet.Services.Metadata/pull/64#discussion_r59756131) [](ancestors = 59756131)

src/NuGet.Indexing/GalleryServiceImpl.cs
maartenba: Fixed

src/NuGet.Indexing/GalleryServiceImpl.cs
maartenba: Agreed. Updated.

src/NuGet.Indexing/GalleryServiceImpl.cs
maartenba: This filters before running the query, reducing the search set (and the amount of results to search a score/ranking for). Adding a comment.

src/NuGet.Indexing/GalleryServiceImpl.cs
maartenba: Fixed, thanks!

src/NuGet.Indexing/GalleryServiceImpl.cs
maartenba: Fixed

src/NuGet.Indexing/RankingsScoreQuery.cs
maartenba: Would it make sense to do a quick meeting around the entire thing?

maartenba: Added some comments, how does it look?

src/NuGet.Indexing/NuGetSearcherManager.cs
```diff
@@ -102,11 +115,21 @@ public class NuGetSearcherManager : SearcherManager<NuGetIndexSearcher>
...
+
+                    azureDirectorySynchronizer = new AzureDirectorySynchronizer(sourceDirectory, directory);
+                }
+
                 searcherManager = new NuGetSearcherManager(
                     indexContainer,
                     loggerFactory.CreateLogger<NuGetSearcherManager>(),
-                    directory ?? new AzureDirectory(storageAccount, indexContainer),
-                    loader ?? new StorageLoader(storageAccount, dataContainer, loggerFactory.CreateLogger<StorageLoader>()));
+                    directory ?? new AzureDirectory(storageAccount, indexContainer, new RAMDirectory()),
```
skofman1: directory always has value at this point

src/NuGet.Indexing/NuGetSearcherManager.cs
```diff
@@ -117,20 +140,37 @@ public class NuGetSearcherManager : SearcherManager<NuGetIndexSearcher>
...
 
             _logger.LogInformation("NuGetSearcherManager.Reopen: refreshed original IndexReader in {IndexReaderReopenDuration} seconds.", stopwatch.Elapsed.TotalSeconds);
 
             return indexReader;
         }
 
+        protected override bool RequiresNewSearcher(IndexReader newReader, IndexSearcher currentSearcher)
+        {
+            return newReader != currentSearcher.IndexReader;
+        }
```
skofman1: it's the same implementation everywhere, so why do u need to override?

src/NuGet.Indexing/NuGetSearcherManager.cs
maartenba: 🙇

src/NuGet.Indexing/NuGetSearcherManager.cs
maartenba: Fixed

src/NuGet.Indexing/NuGetSearcherManager.cs
```diff
@@ -200,48 +220,68 @@ protected override NuGetIndexSearcher CreateSearcher(IndexReader reader)
...
-                    rankings,
+                    rankingsHandler.Result,
                     latestBitSet,
-                    latestStableBitSet);
+                    latestStableBitSet,
+                    ownersHandler.Result);
             }
             catch (Exception e)
             {
                 _logger.LogError("Internal server error", e);
```
skofman1: more details like method name, etc.

src/NuGet.Indexing/NuGetIndexSearcher.cs
```diff
@@ -12,43 +13,59 @@ namespace NuGet.Indexing
...
-        public IDictionary<string, int> Rankings { get; private set; }
+        public RankingsHandler.RankingResult Rankings { get; private set; }
         public OpenBitSet LatestBitSet { get; private set; }
         public OpenBitSet LatestStableBitSet { get; private set; }
+        public OwnersHandler.OwnersResult Owners { get; private set; }
         public DateTime LastReopen { get; private set; }
 
+        public Filter GetFilter(string curatedFeed)
+        {
+            Filter curatedFeedFilter;
```
skofman1: this should throw id curatedFeed is empty

skofman1: Or change it to:
bool TryGetFilter(string curatedFeed, out Filter filter)

---
In reply to: [59766156](https://github.com/NuGet/NuGet.Services.Metadata/pull/64#discussion_r59766156) [](ancestors = 59766156)

src/NuGet.Indexing/AzureDirectorySynchronizer.cs
```diff
@@ -0,0 +1,138 @@
...
+                    }
+                    finally
+                    {
+                        try
+                        {
+                            indexOutput?.Close();
+                        }
+                        finally
+                        {
+                            indexInput?.Close();
```
skofman1: can this throw? if so wrap it in try catch

src/NuGet.Indexing/CuratedFeedHandler.cs
```diff
@@ -59,13 +59,13 @@ public void End(IndexReader indexReader)
             }
         }
 
-        public void Process(IndexReader indexReader, string readerName, int n, Document document, string id, NuGetVersion version)
+        public void Process(IndexReader indexReader, string readerName, int documentNumber, Document document, string id, NuGetVersion version)
```
skofman1: good one :)

src/NuGet.Indexing/RankingsScoreQuery.cs
```diff
@@ -39,41 +39,59 @@ public override string ToString()
...
 
-            public RankingScoreProvider(IndexReader reader, IDictionary<string, int> rankings, double baseBoost)
+            public RankingScoreProvider(IndexReader reader, RankingsHandler.RankingResult rankings, double baseBoost)
                 : base(reader)
             {
                 _rankings = rankings;
-                _ids = FieldCache_Fields.DEFAULT.GetStrings(reader, "Id");
                 _baseBoost = baseBoost;
+
+                // We need the reader name: Lucene *may* has multiple segments (which are smaller indices)
```
skofman1: have

skofman1:
LGTM

src/NuGet.Indexing/NuGetSearcherManager.cs
maartenba: Fixed, also added some more granularity in things that could go wrong in that method.

src/NuGet.Indexing/NuGetIndexSearcher.cs
maartenba: Updated.

src/NuGet.Indexing/AzureDirectorySynchronizer.cs
maartenba: Caught in NuGetSearcherManager - IndexReader Reopen(IndexSearcher searcher)

src/NuGet.Indexing/CuratedFeedHandler.cs
maartenba: Clarity for the win :-)

src/NuGet.Indexing/RankingsScoreQuery.cs
maartenba: Fixed, thanks!

PR #65, opened 2016-04-18
Catalog2Lucene: process all known batches before committing index

maartenba:
Catalog2Lucene: process all known batches before committing index

Right now, catalog2lucene does a commit to the Lucene index for each batch that is read from the catalog.
Typically, such batch contains few packages that were added or deleted, which means that there are many cases where we commit to Lucene for just 2-3 packages. This is wasteful, as each commit means uploading the changed segmenst to storage (often a few hundred MB).

This behaviour becomes problematic when multiple packages come in. Imagine 100 packages flow into NuGet. These will usually be distributed over 10-20 batches depending on the package owner's upload speed. For catalog2lucene that would mean 10-20 commits, and thus uploading 10-20 times the updated segment. For a 100 MB segment that would mean 1GB-2GB of uploading, during which catalog2lucene can not process next batches. This results in index lag.

To overcome this issue, this PR reads all new batches from the catalog, processes them, and then does only one commit at the end of the run, similar to what the V2 search service did. This would mean in the above example that the changes of the 10-20 batches are applied to Lucene but only committed once, meaning only one upload has to happen to the Azure directory. This will result in more throughput for catalog2lucene, reducing index lag.

This change does not affect smaller batches / calm moments. If only e.g. 3 packages are uploaded in an hour, these 3 will be committed fairly quick (no difference from what we have in production right now).

Please review, @skofman1 @xavierdecoster @joelverhagen


xavierdecoster:
LGTM! :ship:

src/Catalog/Test/PrintCommitCollector.cs
```diff
@@ -16,7 +16,7 @@ public PrintCommitCollector(Uri index, Func<HttpMessageHandler> handlerFunc = nu
         {
         }
 
-        protected override Task<bool> OnProcessBatch(CollectorHttpClient client, IEnumerable<JToken> items, JToken context, DateTime commitTimeStamp, CancellationToken cancellationToken)
+        protected override Task<bool> OnProcessBatch(CollectorHttpClient client, IEnumerable<JToken> items, JToken context, DateTime commitTimeStamp, bool isLastBatch, CancellationToken cancellationToken)
         {
             Console.WriteLine("------------------------------------------------");
             Console.WriteLine("COMMIT: {0}", commitTimeStamp.ToString("O"));
```
skofman1: Can you add isLastBatch to the printout, for fullness

skofman1:
:sheep: :it:

src/Catalog/Test/PrintCommitCollector.cs
maartenba: Fixed, thanks!

PR #66, opened 2016-04-18
Switch Serilog sink to AppInsights

maartenba:
And yes, this is a PR on top of a PR.

@skofman1 @xavierdecoster


src/Ng/packages.config
```diff
@@ -1,11 +1,11 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="dotNetRDF" version="1.0.8.3533" targetFramework="net452" />
-  <package id="Elasticsearch.Net" version="1.7.1" targetFramework="net452" />
   <package id="HtmlAgilityPack" version="1.4.9" targetFramework="net452" />
   <package id="Lucene.Net" version="3.0.3" targetFramework="net452" />
   <package id="Lucene.Net.Contrib" version="3.0.3" targetFramework="net452" />
   <package id="Lucene.Net.Store.Azure" version="3.0.5553.21100" targetFramework="net452" />
+  <package id="Microsoft.ApplicationInsights" version="1.2.3" targetFramework="net452" />
```
xavierdecoster: Does it work with latest Microsoft.ApplicationInsights 2.0.1 too? (I believe we get e.g. sampling capabilities starting in 2.x)

maartenba: Nope

xavierdecoster:
Changes look good to me. :shipit:

src/Ng/packages.config
skofman1: why not? what breaks it?

---
In reply to: [60061135](https://github.com/NuGet/NuGet.Services.Metadata/pull/66#discussion_r60061135) [](ancestors = 60061135)

skofman1:
LGTM

src/Ng/packages.config
maartenba: The Serilog component. I'll see if I can do the update though, may be okay to do so.

maartenba: Updated, seems to work - http://maballia-v2v3.cloudapp.net

PR #67, opened 2016-04-20
Expand acronyms in search 

maartenba:
Recognizes and expands known acronyms in search.

Contains a known list of acronyms and their expansions. Examples:

* Tokenizes XAMLBehaviors into XAML;Behaviors;XAMLBehaviors
* Tokenizes EF into EF;Entity;Entity Framework

This is only applied to the Id field currently, not sure it would make sense to also tokenize in e.g. description field.

@skofman1 @xavierdecoster @joelverhagen @yishaigalatzer


src/NuGet.Indexing/ExpandAcronymsFilter.cs
```diff
@@ -0,0 +1,118 @@
...
+            }
+
+            try
+            {
+                if (!input.IncrementToken()) // end of stream; no more tokens on input stream
+                {
+                    return false;
+                }
+            }
+            catch (Exception ex)
```
xavierdecoster: What exception could be thrown here and why? (is there a more specific exception we can catch instead?)

src/NuGet.Indexing/ExpandAcronymsFilter.cs
```diff
@@ -0,0 +1,118 @@
...
+                return original;
+            }
+
+            var result = new StringBuilder(original.Length);
+
+            int substringLength = substring.Length;
+            int substringStartIndex = -1;
+            int lastCharacterIndex = 0;
+            
+            while (true)
```
xavierdecoster: `while (true) ... if (condition == true) break;` is perhaps more clear when rewritten to use `while (condition == false)`?
e.g. `while (substringStartIndex >= 0)`

src/NuGet.Indexing/ExpandAcronymsFilter.cs
maartenba: This is the guidance Lucene provides :-)

src/NuGet.Indexing/ExpandAcronymsFilter.cs
maartenba: Not sure if that makes sense, there are two variations of adding the string in there. Doing the conditional while would duplicate code for adding the last entry.

src/NuGet.Indexing/NuGetAcronymExpansionProvider.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+    public class NuGetAcronymExpansionProvider : IAcronymExpansionProvider
+    {
+        public static readonly IAcronymExpansionProvider Instance = new NuGetAcronymExpansionProvider();
+
+        private static readonly Dictionary<string, string[]> Acronyms = new Dictionary<string, string[]>(StringComparer.OrdinalIgnoreCase);
+
+        static NuGetAcronymExpansionProvider()
+        {
+            Acronyms.Add("xaml",  new[] { "xaml" });
```
skofman1: perhaps not for this PR, but I would have this list somewhere where it can be easily modified, without the need to redeploy the service (like a json in a blob).

skofman1:
LGTM

src/NuGet.Indexing/NuGetAcronymExpansionProvider.cs
maartenba: Totally agree ☺

yishaigalatzer: I wouldn't put it in a json blob, I too thought this was a good idea, but it will create a bunch of issues when trying to do A/B testing, and add another connectivity pitfall.

I think deployment is fine, I would make it a resource file that is easy to edit, and add a test to validate that the values are valid (e.g. the xml or json loads fine).

I would make it part of this PR

src/NuGet.Indexing/CamelCaseFilter.cs
```diff
@@ -1,19 +1,19 @@
...
 using System.Collections.Generic;
 
 namespace NuGet.Indexing
 {
     public class CamelCaseFilter : TokenFilter
     {
-        ITermAttribute _termAttribute;
-        IOffsetAttribute _offsetAttribute;
-        IPositionIncrementAttribute _positionIncrementAttribute;
+        private readonly ITermAttribute _termAttribute;
```
yishaigalatzer: :+1:

src/NuGet.Indexing/ExpandAcronymsFilter.cs
```diff
@@ -0,0 +1,117 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Lucene.Net.Analysis;
+using Lucene.Net.Analysis.Tokenattributes;
+
+namespace NuGet.Indexing
+{
+    public sealed class ExpandAcronymsFilter : TokenFilter
```
yishaigalatzer: we don't really need sealed classes. The general guidance is to avoid sealing in all our code

src/NuGet.Indexing/ExpandAcronymsFilter.cs
```diff
@@ -0,0 +1,117 @@
...
+            _tokenSet = new HashSet<string>();
+        }
+
+        public override bool IncrementToken()
+        {
+            if (_tokenSet.Count > 0)
+            {
+                RestoreState(_currentState);
+
+                var tokenAtHead = _tokenSet.First();
```
yishaigalatzer: what does first mean in a hashset?

src/NuGet.Indexing/ExpandAcronymsFilter.cs
```diff
@@ -0,0 +1,117 @@
...
+                foreach (var acronym in acronyms)
+                {
+                    // Add expanded acronym (ef => entity;framework)
+                    foreach (var expansion in _acronymExpansionProvider.Expand(acronym))
+                    {
+                        _tokenSet.Add(expansion);
+                    }
+
+                    // Add original term without the acronym (xamlbehaviors with xaml acronym => behaviors)
+                    var termWithoutAcronym = RemoveSubstring(_termAttribute.Term, acronym);
```
yishaigalatzer: measure to see if this gets hot when processing lots of data, you can theoretically do the math without allocation, but I don't know if it matters here perf wise or not without measuring

src/NuGet.Indexing/ExpandAcronymsFilter.cs
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        private static string RemoveSubstring(string original, string substring)
+        {
+            if (string.IsNullOrEmpty(original) || string.IsNullOrEmpty(substring))
+            {
+                return original;
+            }
+
+            var result = new StringBuilder(original.Length);
```
yishaigalatzer: you don't need to optimize here, let it grow if it needs to

src/NuGet.Indexing/ExpandAcronymsFilter.cs
```diff
@@ -0,0 +1,117 @@
...
+                        _tokenSet.Add(termWithoutAcronym);
+                    }
+                }
+            }
+
+            _currentState = CaptureState();
+            return true;
+        }
+
+        private static string RemoveSubstring(string original, string substring)
```
yishaigalatzer: Add comments explaining exactly what this method does, and write tests to cover it

src/NuGet.Indexing/NuGetAcronymExpansionProvider.cs
```diff
@@ -0,0 +1,44 @@
...
+            Acronyms.Add("mvc",   new[] { "mvc" });
+            Acronyms.Add("wpf",   new[] { "wpf" });
+            Acronyms.Add("ef",    new[] { "ef", "entity", "framework", "entity framework" });
```
yishaigalatzer: Do we need both entity framework and entity framework? Wouldn't the other tokenizer break it down itself later?

src/NuGet.Indexing/NuGetAcronymExpansionProvider.cs
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NuGet.Indexing
+{
+    public class NuGetAcronymExpansionProvider : IAcronymExpansionProvider
```
yishaigalatzer: create a private constructor so this class cannot be initialized in any other way

yishaigalatzer:
:watch:

src/NuGet.Indexing/NuGetAcronymExpansionProvider.cs
maartenba: Thanks, fixed.

src/NuGet.Indexing/ExpandAcronymsFilter.cs
maartenba: Fixed, thanks.

src/NuGet.Indexing/ExpandAcronymsFilter.cs
maartenba: Made it a queue. (and a Hashset to skip duplicates)

src/NuGet.Indexing/ExpandAcronymsFilter.cs
maartenba: Should only be hit when an acronym is detected. Added some unit tests and will profile under load.

src/NuGet.Indexing/ExpandAcronymsFilter.cs
maartenba: It will only shrink in this case (or stay same length)

src/NuGet.Indexing/ExpandAcronymsFilter.cs
maartenba: Fixed

src/NuGet.Indexing/NuGetAcronymExpansionProvider.cs
maartenba: Good catch!

src/NuGet.Indexing/NuGetAcronymExpansionProvider.cs
maartenba: Fixed.

maartenba:
Made some updates, please have another look.

skofman1:
Still looks good

src/NuGet.Indexing/Properties/Strings.resx
```diff
@@ -0,0 +1,161 @@
...
+
+  "http": [
+    "http"
+  ],
+
+  "https": [
+    "https"
+  ],
+
+  "ftp": [
```
joelverhagen: Is there a case where the token itself will not be in the list? Is not, why not make the code as self as an expanded token?

PR #68, opened 2016-04-29
Add ApplicationInsights OWIN Middleware

maartenba:
Open questions:

* Do we need to catch a possible exception [here](https://github.com/NuGet/NuGet.Services.Metadata/pull/68/files#diff-4aba68f7a5514fc37f056085fc7f7818R53), track it and rethrow to ensure exceptions and requests are correlated?
* Do we need a custom `ITelemetryInitializer` that sets the `telemetry.Context.Operation.Id = ...` to the correlation ID from OWIN so that we can correlate any custom event/metric we send elsewhere in the codebase?

@xavierdecoster @skofman1 

Do not merge this in yet, we are awaiting feedback from Application Insights team.


skofman1:
Did you see this code: https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/master/src/Microsoft.ApplicationInsights.AspNet/RequestTrackingMiddleware.cs

I think the only thing missing in your code is passing sdk verison in TrackRequest.
In addition, I would change the name to RequestTrackingMiddleware (again, like the offical one), since you might want to add exception tracking as well.

maartenba:
SDK version seems to be internal. Open questions remain for AI team.

maartenba:
Did some changes based on AI team feedback. Awaiting their next review pass. (@skofman1 @xavierdecoster feel free to also do a pass)

src/NuGet.Services.BasicSearch/OwinRequestIdTelemetryInitializer.cs
```diff
@@ -0,0 +1,20 @@
...
+using Microsoft.ApplicationInsights.Extensibility;
+
+namespace NuGet.Services.BasicSearch
+{
+    public class OwinRequestIdTelemetryInitializer
+        : ITelemetryInitializer
+    {
+        public void Initialize(ITelemetry telemetry)
+        {
+            if (OwinRequestIdContext.Get() != null)
```
xavierdecoster: Maybe store the result of `OwinRequestIdContext.Get()` in a local field, instead of calling it twice?

xavierdecoster:
LGTM

src/NuGet.Services.BasicSearch/OwinRequestIdTelemetryInitializer.cs
maartenba: Fixed :) Good catch!

abaranch:
LGTM

src/NuGet.Services.BasicSearch/ApplicationInsightsMiddleware.cs
```diff
@@ -0,0 +1,97 @@
...
+            : this(next, null)
+        {
+        }
+
+        public ApplicationInsightsMiddleware(OwinMiddleware next, TelemetryConfiguration telemetryConfiguration) 
+            : base(next)
+        {
+            if (next == null)
+            {
+                throw new ArgumentNullException(nameof(next));
```
dnduffy: What if this is the last one?  Shouldn't next be legitimately null for the last middleware?

src/NuGet.Services.BasicSearch/ApplicationInsightsMiddleware.cs
```diff
@@ -0,0 +1,97 @@
...
+            var name = $"{requestMethod} {path}";
+
+            var telemetry = new RequestTelemetry
+            {
+                Id = requestId,
+                Name = name,
+                StartTime = requestStartDate,
+                Duration = duration,
+                ResponseCode = responseCode.ToString(),
+                Success = responseCode < 400 || !requestFailed,
```
dnduffy: This seems like a bit of a stretch since not everything below 400 is truly "success".

dnduffy:
The null check seems fragile, otherwise ship it.

src/NuGet.Services.BasicSearch/ApplicationInsightsMiddleware.cs
maartenba: Removed the check. Ideally this should be (one of) the first middlewares registered. Good catch!

src/NuGet.Services.BasicSearch/ApplicationInsightsMiddleware.cs
maartenba: Switched to >= 200 && <= 299 (as per https://msdn.microsoft.com/en-us/library/windows/apps/windows.web.http.httpresponsemessage.issuccessstatuscode)

src/NuGet.ApplicationInsights.Owin/ApplicationInsightsMiddleware.cs
```diff
@@ -0,0 +1,95 @@
...
+        public const string OwinRequestIdKey = "owin.RequestId";
+
+        private readonly TelemetryClient _telemetryClient;
+
+        public ApplicationInsightsMiddleware(OwinMiddleware next)
+            : this(next, null)
+        {
+        }
+
+        public ApplicationInsightsMiddleware(OwinMiddleware next, TelemetryConfiguration telemetryConfiguration)
```
kartang: Minor: Can we enable the scope for telemetry client as opposed to telemetryconfiguration? I don't see using telemetryConfiguration anywhere. Why not just get telemetry client?

skofman1:
Regarding my previous comment: how about changing the name to RequestTrackingMiddleware (like the offical one), since you might want to add exception tracking as well. What do you think?

src/NuGet.ApplicationInsights.Owin/ApplicationInsightsMiddleware.cs
```diff
@@ -0,0 +1,95 @@
...
+                OwinRequestIdContext.Set(requestId);
+
+                if (Next != null)
+                {
+                    await Next.Invoke(context);
+                }
+            }
+            catch (Exception)
+            {
+                requestFailed = true;
```
kartang: Addition: Would be good to have a similar exception middleware that tracks all the failures. We might miss all the failures in this case.

maartenba: Would just tracking the exception here help?

kartang: That will help. However, if we have a different middleware that explicitly handles exceptions, it might be a concern tomorrow. At that point, we might want to consider getting the details that the handler throws. It might be ideal to split it as two middleware's.
Adding @dnduffy

src/NuGet.ApplicationInsights.Owin/OwinRequestIdContext.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    public static class OwinRequestIdContext
+    {
+        public static void Set(string value)
+        {
+            if (!string.IsNullOrEmpty(value))
+            {
+                CallContext.LogicalSetData(ApplicationInsightsMiddleware.OwinRequestIdKey, value);
+            }
+        }
```
kartang: This is interesting. Out of curiosity, could you please explain the reason behind using CallContext to store the id? Just to understand the working scenario. Thanks!

maartenba: Seemed like the easiest way of setting the context and sharing it between the OWIN pipeline and the telemetry initializer. If there's a better way I'd love to learn :)

kartang: That makes sense. I totally missed that point of initializing the id. So, all the telemetry transmitted within the scope of this request will have the same operationId. Thanks!

PR #77, opened 2016-07-13
Adding KeyVault Integration for Jobs

ryuyu:
This PR makes the following changes:
 * Adds support for pulling credentials from the Azure KeyVault
 * Updates usage text to indicate new Azure KeyVault options
 * Updates the run scripts to pull values from Azure KeyVault

Note that KeyVault usage is optional and will not be used if vaultName option is not specified.

@maartenba @skofman1


src/Ng/CommandHelpers.cs
```diff
@@ -501,6 +551,46 @@ public static Lucene.Net.Store.Directory GetLuceneDirectoryImpl(IDictionary<stri
                     return null;
                 }
 
+                string vaultName;
+                if (!arguments.TryGetValue("-vaultName", out vaultName))
+                {
+                    useKeyVault = false;
+                }
+
+                if (useKeyVault)
```
scottbommarito: There appears to be a lot of shared keyvault initialization code between here and the `if (useKeyVault)` above...could we make somehow reduce this?

src/Ng/CommandHelpers.cs
```diff
@@ -284,24 +285,69 @@ public static StorageFactory CreateStorageFactory(IDictionary<string, string> ar
             else if (storageType.Equals("Azure", StringComparison.InvariantCultureIgnoreCase))
             {
                 string storageAccountName;
-                if (!arguments.TryGetValue("-storageAccountName", out storageAccountName))
+                string storageKeyValue;
+
+                bool useKeyVault = true;
+
+                string storageContainer;
+                if (!arguments.TryGetValue("-storageContainer", out storageContainer))
```
maartenba: Can we refactor this construct a bit? E.g. something like:

```
string storageContainer;
TryGetArgument(arguments, "-storageContainer", out storageContainer, required: true);
```

Makes the entire code block a bit more readable?

src/Ng/CommandHelpers.cs
```diff
@@ -284,24 +285,69 @@ public static StorageFactory CreateStorageFactory(IDictionary<string, string> ar
...
+                    if (!arguments.TryGetValue("-certificateThumbprint", out certificateThumbprint))
+                    {
+                        TraceRequiredArgument("-certificateThumbprint");
+                        return null;
+                    }
+
+                    string validateCertificate = "false";
+                    arguments.TryGetValue("-validateCertificate", out validateCertificate);
+
+                    bool shouldValidateCertificate = validateCertificate == null ? false : validateCertificate.Equals("true", StringComparison.InvariantCultureIgnoreCase);
```
maartenba: Nit: Add some line breaks, e.g.: 

```
bool result = condition
    ? true
    : false
```

Rule of thumb is that we should be able to read the code on GitHub without having to scroll right :-)

src/Ng/CommandHelpers.cs
```diff
@@ -480,15 +526,19 @@ public static Lucene.Net.Store.Directory GetLuceneDirectoryImpl(IDictionary<stri
             }
             else if (luceneDirectoryType.Equals("Azure", StringComparison.InvariantCultureIgnoreCase))
             {
-                string luceneStorageAccountName;
-                if (!arguments.TryGetValue(names["storageAccountName"], out luceneStorageAccountName))
+                bool useKeyVault = true;
+                string luceneStorageAccountName,
```
maartenba: Initialize both on a separate line ending in `;`

```
string a = "";
string b= "";
```

src/Ng/Ng.csproj
```diff
@@ -61,6 +61,10 @@
       <HintPath>..\..\packages\HtmlAgilityPack.1.4.9\lib\Net45\HtmlAgilityPack.dll</HintPath>
       <Private>True</Private>
     </Reference>
+    <Reference Include="Hyak.Common, Version=1.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
```
maartenba: Where does this new reference come from?

maartenba:
A few remarks, other than that LGTM

src/Ng/Catalog2Dnx.cs
```diff
@@ -49,7 +49,7 @@ public Catalog2Dnx(ILoggerFactory loggerFactory)
 
         private static void PrintUsage()
         {
-            Console.WriteLine("Usage: ng catalog2dnx -source <catalog> -contentBaseAddress <content-address> -storageBaseAddress <storage-base-address> -storageType file|azure [-storagePath <path>]|[-storageAccountName <azure-acc> -storageKeyValue <azure-key> -storageContainer <azure-container> -storagePath <path>] [-verbose true|false] [-interval <seconds>]");
+            Console.WriteLine("Usage: ng catalog2dnx -source <catalog> -contentBaseAddress <content-address> -storageBaseAddress <storage-base-address> -storageType file|azure [-storagePath <path>]|[-storageAccountName <azure-acc> -storageKeyValue <azure-key> -storageContainer <azure-container> -storagePath <path> [-vaultName <keyvault-name> -clientId <keyvault-client-id> -certificateThumbprint <keyvault-certificate-thumbprint> [-validateCertificate true|false]]] [-verbose true|false] [-interval <seconds>]");
```
skofman1: you can remove -validateCertificate. Lets have it always as true

src/Ng/Catalog2Lucene.cs
```diff
@@ -74,7 +74,7 @@ internal static IndexWriter CreateIndexWriter(Lucene.Net.Store.Directory directo
 
         static void PrintUsage()
         {
-            Console.WriteLine("Usage: ng catalog2lucene -source <catalog> [-registration <registration-root>] -luceneDirectoryType file|azure [-lucenePath <file-path>] | [-luceneStorageAccountName <azure-acc> -luceneStorageKeyValue <azure-key> -luceneStorageContainer <azure-container>] [-verbose true|false] [-interval <seconds>]");
+            Console.WriteLine("Usage: ng catalog2lucene -source <catalog> [-registration <registration-root>] -luceneDirectoryType file|azure [-lucenePath <file-path>] | [-luceneStorageAccountName <azure-acc> -luceneStorageKeyValue <azure-key> -luceneStorageContainer <azure-container> [-vaultName <keyvault-name> -clientId <keyvault-client-id> -certificateThumbprint <keyvault-certificate-thumbprint> [-validateCertificate true|false]]] [-verbose true|false] [-interval <seconds>]");
```
skofman1: same

src/Ng/CommandHelpers.cs
skofman1: And lets use consts for argument names

src/Ng/CommandHelpers.cs
```diff
@@ -284,24 +285,69 @@ public static StorageFactory CreateStorageFactory(IDictionary<string, string> ar
...
+                        TraceRequiredArgument("-certificateThumbprint");
+                        return null;
+                    }
+
+                    string validateCertificate = "false";
+                    arguments.TryGetValue("-validateCertificate", out validateCertificate);
+
+                    bool shouldValidateCertificate = validateCertificate == null ? false : validateCertificate.Equals("true", StringComparison.InvariantCultureIgnoreCase);
+
+                    KeyVaultConfiguration keyVaultConfiguration = new KeyVaultConfiguration(vaultName, clientId, certificateThumbprint, shouldValidateCertificate);
```
skofman1: var

src/Ng/CommandHelpers.cs
```diff
@@ -284,24 +285,69 @@ public static StorageFactory CreateStorageFactory(IDictionary<string, string> ar
...
+                string storageKeyValueFrame;
+                if (!arguments.TryGetValue("-storageKeyValue", out storageKeyValueFrame))
                 {
                     TraceRequiredArgument("-storageKeyValue");
                     return null;
                 }
 
-                string storageContainer;
-                if (!arguments.TryGetValue("-storageContainer", out storageContainer))
+                string vaultName;
```
skofman1: please move all the KeyVault related initializations and arg parsing to a separate method

src/Ng/CommandHelpers.cs
```diff
@@ -284,24 +285,69 @@ public static StorageFactory CreateStorageFactory(IDictionary<string, string> ar
...
+                string storageAccountNameFrame;
+                if (!arguments.TryGetValue("-storageAccountName", out storageAccountNameFrame))
                 {
                     TraceRequiredArgument("-storageAccountName");
                     return null;
                 }
 
-                string storageKeyValue;
-                if (!arguments.TryGetValue("-storageKeyValue", out storageKeyValue))
+                string storageKeyValueFrame;
```
skofman1: why Frame?

src/Ng/Ng.csproj
skofman1: KeyVault..

---
In reply to: [70754613](https://github.com/NuGet/NuGet.Services.Metadata/pull/77#discussion_r70754613) [](ancestors = 70754613)

src/Ng/CommandHelpers.cs
ryuyu: done

src/Ng/CommandHelpers.cs
```diff
@@ -370,99 +399,78 @@ public static StorageFactory CreateSuffixedStorageFactory(string suffix, IDictio
...
                 throw new ArgumentNullException("suffix");
             }
-
-            Uri storageBaseAddress = null;
-            string storageBaseAddressStr;
-            if (arguments.TryGetValue("-storageBaseAddress" + suffix, out storageBaseAddressStr))
+            try
             {
-                storageBaseAddressStr = storageBaseAddressStr.TrimEnd('/') + "/";
+                Uri storageBaseAddress = null;
```
scottbommarito: I've seen this `storageBaseAddress` chunk a couple times, you could probably reduce this into a function you call in all these `StorageFactory` factory methods.

src/Ng/CommandHelpers.cs
```diff
@@ -370,99 +399,78 @@ public static StorageFactory CreateSuffixedStorageFactory(string suffix, IDictio
...
-                return null;
-            }
+                string storageType;
+                TryGetArgument(arguments, StorageType, out storageType, required: true);
 
-            if (storageType.Equals("File", StringComparison.InvariantCultureIgnoreCase))
-            {
-                string storagePath;
-                if (!arguments.TryGetValue("-storagePath" + suffix, out storagePath))
+                if (storageType.Equals("File", StringComparison.InvariantCultureIgnoreCase))
```
scottbommarito: This `if`, and the `Azure` `if` below, appear to be repeated in `CreateStorageFactory` as well. You may want to put it in a function.

scottbommarito:
This is like a million times cleaner now, I love it.

src/Ng/Catalog2Dnx.cs
```diff
@@ -49,7 +49,23 @@ public Catalog2Dnx(ILoggerFactory loggerFactory)
 
         private static void PrintUsage()
         {
-            Console.WriteLine("Usage: ng catalog2dnx -source <catalog> -contentBaseAddress <content-address> -storageBaseAddress <storage-base-address> -storageType file|azure [-storagePath <path>]|[-storageAccountName <azure-acc> -storageKeyValue <azure-key> -storageContainer <azure-container> -storagePath <path> [-vaultName <keyvault-name> -clientId <keyvault-client-id> -certificateThumbprint <keyvault-certificate-thumbprint> [-validateCertificate true|false]]] [-verbose true|false] [-interval <seconds>]");
+            Console.WriteLine("Usage: ng catalog2dnx "
```
maartenba: WHOA! READABILITY! Big 👍

maartenba:
Love how clean it became, LGTM!

src/Ng/Catalog2Dnx.cs
```diff
@@ -49,7 +49,23 @@ public Catalog2Dnx(ILoggerFactory loggerFactory)
 
         private static void PrintUsage()
         {
-            Console.WriteLine("Usage: ng catalog2dnx -source <catalog> -contentBaseAddress <content-address> -storageBaseAddress <storage-base-address> -storageType file|azure [-storagePath <path>]|[-storageAccountName <azure-acc> -storageKeyValue <azure-key> -storageContainer <azure-container> -storagePath <path>] [-verbose true|false] [-interval <seconds>]");
+            Console.WriteLine("Usage: ng catalog2dnx "
+                + "-"  + CommandHelpers.Source             + " <catalog> "
+                + "-"  + CommandHelpers.ContentBaseAddress + " <content-address> "
```
skofman1: nit: use C# 6 formatting. That way the string creation will look even cleaner:
$"-{CommandHelpers.Source} <...>"

src/Ng/CommandHelpers.cs
```diff
@@ -16,9 +17,66 @@ namespace Ng
 {
     static class CommandHelpers
     {
+        #region ArgumentStrings
+        public const string CatalogBaseAddress = "catalogBaseAddress";
```
skofman1: >CatalogBaseAddress [](start = 28, length = 18)

please move to a separate static class in a separate file.

src/Ng/CommandHelpers.cs
```diff
@@ -16,9 +17,66 @@ namespace Ng
...
+        public const string Verbose = "verbose";
+        public const string Version = "version";
+        public const string UnlistShouldDelete = "unlistShouldDelete";
+        public const string UseCompressedStorage = "useCompressedStorage"; 
+        #endregion
+
         public static IDictionary<string, string> GetArguments(string[] args, int start)
         {
             IDictionary<string, string> result = new Dictionary<string, string>();
+            List<string> inputArgs = new List<string>();
```
skofman1: >< [](start = 16, length = 1)

var

src/Ng/CommandHelpers.cs
```diff
@@ -16,9 +17,66 @@ namespace Ng
...
+        public const string ValidateCertificate = "validateCertificate";
+        public const string Verbose = "verbose";
+        public const string Version = "version";
+        public const string UnlistShouldDelete = "unlistShouldDelete";
+        public const string UseCompressedStorage = "useCompressedStorage"; 
+        #endregion
+
         public static IDictionary<string, string> GetArguments(string[] args, int start)
         {
             IDictionary<string, string> result = new Dictionary<string, string>();
```
skofman1: >string [](start = 24, length = 6)

var

src/Ng/CommandHelpers.cs
```diff
@@ -34,6 +92,17 @@ static class CommandHelpers
             for (int i = 1; i < args.Length; i += 2)
             {
                 result.Add(args[i], args[i + 1]);
+                inputArgs.Add(args[i]);
+            }
+
+            var secretInjector = GetSecretInjector(result);
+
+            if (secretInjector != null)
```
skofman1: no need for this null check. you can have GetSecretInjector to return EmptySecretReader, that will just return the same string.

src/Ng/CommandHelpers.cs
```diff
@@ -34,6 +92,17 @@ static class CommandHelpers
             for (int i = 1; i < args.Length; i += 2)
             {
                 result.Add(args[i], args[i + 1]);
+                inputArgs.Add(args[i]);
```
skofman1: >inputArgs [](start = 16, length = 9)

we don't need a second data structure. You can just go over the original dictionary, and return a new dictionary.

src/Ng/CommandHelpers.cs
```diff
@@ -45,86 +114,113 @@ private static void TraceRequiredArgument(string name)
...
         {
-            string value;
-            if (!arguments.TryGetValue("-source", out value))
+            Console.WriteLine("Argument \"{0}\" not provided", name);
+            Trace.TraceWarning("Argument \"{0}\" not provided", name);
+        }
+
+        private static void TryGetArgument(IDictionary<string, string> arguments, string searchArg, out string value, bool required = false)
+        {
+            if (!arguments.TryGetValue("-" + searchArg, out value))
```
skofman1: >"-" [](start = 39, length = 3)

this is used a lot in the code. Make a const out of it.

src/Ng/CommandHelpers.cs
```diff
@@ -45,86 +114,113 @@ private static void TraceRequiredArgument(string name)
...
+            {
+                if (required)
+                {
+                    TraceRequiredArgument(searchArg);
+                    throw new ArgumentException();
+                }
+                else
+                {
+                    TraceMissingArgument(searchArg);
```
skofman1: >TraceMissingArgument [](start = 20, length = 20)

I don't think this is necessary. Not all arguments are required in all scenarios, and printing every missing one just creates noise.

src/Ng/CommandHelpers.cs
```diff
@@ -45,86 +114,113 @@ private static void TraceRequiredArgument(string name)
...
+        }
+
+        private static void TryGetArgument(IDictionary<string, string> arguments, string searchArg, out string value, bool required = false)
+        {
+            if (!arguments.TryGetValue("-" + searchArg, out value))
+            {
+                if (required)
+                {
+                    TraceRequiredArgument(searchArg);
+                    throw new ArgumentException();
```
skofman1: >ArgumentException [](start = 30, length = 17)

Add info to the exception, always a good practice

src/Ng/CommandHelpers.cs
```diff
@@ -45,86 +114,113 @@ private static void TraceRequiredArgument(string name)
...
+        {
+            try
+            {
+                string vaultName;
+                TryGetArgument(arguments, VaultName, out vaultName, required: true);
+
+                string clientId;
+                TryGetArgument(arguments, ClientId, out clientId, required: true);
+
+                string certificateThumbprint = "";
```
skofman1: >""; [](start = 47, length = 3)

why the initialization?

src/Ng/CommandHelpers.cs
```diff
@@ -45,86 +114,113 @@ private static void TraceRequiredArgument(string name)
...
-        public static string GetSource(IDictionary<string, string> arguments)
+        private static void TraceMissingArgument(string name)
         {
-            string value;
-            if (!arguments.TryGetValue("-source", out value))
+            Console.WriteLine("Argument \"{0}\" not provided", name);
+            Trace.TraceWarning("Argument \"{0}\" not provided", name);
+        }
+
+        private static void TryGetArgument(IDictionary<string, string> arguments, string searchArg, out string value, bool required = false)
```
skofman1: >void [](start = 23, length = 4)

in general, when a method is called TryGetXYZ we expect it to have a certain structure. Meaning, it returns a value using out parameter, and returns a bool, saying if it was successful or not. Although we do something a bit different here, with the throw, I still think it's a good idea to return a bool in case we dont throw.

src/Ng/CommandHelpers.cs
```diff
@@ -45,86 +114,113 @@ private static void TraceRequiredArgument(string name)
...
+                string vaultName;
+                TryGetArgument(arguments, VaultName, out vaultName, required: true);
+
+                string clientId;
+                TryGetArgument(arguments, ClientId, out clientId, required: true);
+
+                string certificateThumbprint = "";
+                TryGetArgument(arguments, CertificateThumbprint, out certificateThumbprint, required: true);
+
+                string validateCertificate = "false";
```
skofman1: >false [](start = 46, length = 5)

no need to set it here, will be overridden anyways

src/Ng/CommandHelpers.cs
```diff
@@ -45,86 +114,113 @@ private static void TraceRequiredArgument(string name)
...
+                arguments.TryGetValue(ValidateCertificate, out validateCertificate);
+
+                bool shouldValidateCertificate = validateCertificate == null
+                    ? false
+                    : validateCertificate.Equals("true", StringComparison.InvariantCultureIgnoreCase);
```
skofman1: you can use bool.TryParse

src/Ng/CommandHelpers.cs
```diff
@@ -45,86 +114,113 @@ private static void TraceRequiredArgument(string name)
...
+                else
+                {
+                    TraceMissingArgument(searchArg);
+                }
+
+                value = null;
+            }
+        }
+
+        public static SecretInjector GetSecretInjector(IDictionary<string, string> arguments)
```
skofman1: >GetSecretInjector [](start = 37, length = 17)

In this method you don't distinguish between two cases:
1. We don't want to use KeyVault, so we didn't provide any arguments for it
2. We WANT to use KeyVault, but we provided partial arguments for it.
In the first case, the method should succeed, and the flow should continue
In the second case. we want to stop the flow, because we got bad arguments

src/Ng/CommandHelpers.cs
```diff
@@ -45,86 +114,113 @@ private static void TraceRequiredArgument(string name)
...
         public static bool GetUnlistShouldDelete(IDictionary<string, string> arguments)
         {
-            string unlistShouldDeleteStr = null;
-            if (arguments.TryGetValue("-unlistShouldDelete", out unlistShouldDeleteStr))
-            {
-                return unlistShouldDeleteStr.Equals("true", StringComparison.InvariantCultureIgnoreCase);
-            }
-
-            return false;
+            return GetBool(arguments, UnlistShouldDelete, false);
```
skofman1: >false [](start = 58, length = 5)

use named parameter

skofman1: >GetBool [](start = 19, length = 7)

use GetBool in the KeyVault argument parsing code as well

src/Ng/CommandHelpers.cs
```diff
@@ -140,11 +236,7 @@ public static bool GetVerbose(IDictionary<string, string> arguments)
         public static bool GetBool(IDictionary<string, string> arguments, string argumentName, bool defaultValue)
         {
             string argumentValue;
-            if (!arguments.TryGetValue("-" + argumentName, out argumentValue))
-            {
-                return defaultValue;
-            }
-
+            TryGetArgument(arguments, argumentName, out argumentValue);
             return string.IsNullOrEmpty(argumentValue) ? defaultValue : argumentValue.Equals("true", StringComparison.InvariantCultureIgnoreCase);
```
skofman1: >argumentValue.Equals("true", StringComparison.InvariantCultureIgnoreCase); [](start = 72, length = 74)

bool.Parse. In case the user misspelled falce, we want to fail

skofman1:
>         public static int GetInterval(IDictionary<string, string> arguments)

LOL.. hardcoded interval in CommandHelpers?! Can you please make this be passed explicitly? I know it's not a part of your change, but it would be a nice fix

---
Refers to: src/Ng/CommandHelpers.cs:243 in c74e96f. [](commit_id = c74e96f54c1a359213f51e2af0762922cafac3da, deletion_comment = False)

src/Ng/CommandHelpers.cs
```diff
@@ -153,167 +245,150 @@ public static int GetInterval(IDictionary<string, string> arguments)
...
 
-                storageBaseAddress = new Uri(storageBaseAddressStr);
+                StorageCredentials credentials = new StorageCredentials(storageAccountName, storageKeyValue);
+                CloudStorageAccount account = new CloudStorageAccount(credentials, true);
+                return new AzureStorageFactory(account, storageContainer, storagePath, storageBaseAddress) { Verbose = verbose, CompressContent = true };
             }
-
-            string storageType;
-            if (!arguments.TryGetValue("-storageType", out storageType))
+            catch (ArgumentException)
```
skofman1: >ArgumentException [](start = 19, length = 17)

please check this, but I think that when a required argument is missing, you can just throw the exception and catch it higher in the code. From what I saw, the method that calls CreateCompressedStorageFactory, checks if the returned value is null and fails. It might as well catch an exception.

src/Ng/CommandHelpers.cs
```diff
@@ -153,167 +245,150 @@ public static int GetInterval(IDictionary<string, string> arguments)
...
+        public static StorageFactory CreateStorageFactory(IDictionary<string, string> arguments, bool verbose)
+        {
+            try
             {
-                string storagePath;
-                if (!arguments.TryGetValue("-storagePath", out storagePath))
+                Uri storageBaseAddress = null;
+                string storageBaseAddressStr;
+                TryGetArgument(arguments, StorageBaseAddress, out storageBaseAddressStr);
+                if (!string.IsNullOrEmpty(storageBaseAddressStr))
```
skofman1: >if (!string.IsNullOrEmpty(storageBaseAddressStr)) [](start = 16, length = 49)

if TryGetArgument returns true if it succeeds, you won't need this line

skofman1:
:clock1:

src/Ng/CommandHelpers.cs
ryuyu: Leftover from a different pattern that wouldn't compile without it. Cleaned up.

src/Ng/CommandHelpers.cs
ryuyu: Same comment as above, leftover from a previous pattern. Cleaned up.

src/Ng/CommandHelpers.cs
```diff
@@ -45,275 +55,269 @@ private static void TraceRequiredArgument(string name)
...
+                    bool shouldValidateCertificate = GetBool(arguments, Constants.ValidateCertificate, defaultValue: false);
+
+                    var keyVaultConfiguration = new KeyVaultConfiguration(vaultName, clientId, certificateThumbprint, shouldValidateCertificate);
+                    secretReader = new KeyVaultReader(keyVaultConfiguration);
+                }
+            }
+            catch (ArgumentException)
+            {
+                secretReader = new EmptySecretReader();
             }
```
skofman1: this is still not the desired behavior, since this method never fails. If VaultName is provided, but ClientId is not, we will succeed and return EmptySecretReader to the caller, instead of failing all together and printing usage help string.

src/Ng/CommandHelpers.cs
```diff
@@ -45,275 +55,269 @@ private static void TraceRequiredArgument(string name)
...
-                {
-                    TraceRequiredArgument("-storageAccountName");
-                    return null;
-                }
+                string storageType;
+                TryGetArgument(arguments, Constants.StorageType, out storageType, required: true);
 
-                string storageKeyValue;
-                if (!arguments.TryGetValue("-storageKeyValue", out storageKeyValue))
+                if (storageType.Equals("File", StringComparison.InvariantCultureIgnoreCase))
```
skofman1: >File [](start = 40, length = 4)

const

src/Ng/CommandHelpers.cs
```diff
@@ -45,275 +55,269 @@ private static void TraceRequiredArgument(string name)
...
-                    TraceRequiredArgument("-storageKeyValue");
-                    return null;
+                    string storagePath;
+                    TryGetArgument(arguments, Constants.StoragePath, out storagePath, required: true);
+
+                    if (storageBaseAddress == null)
+                    {
+                        TraceRequiredArgument("-storageBaseAddress");
+                        return null;
+                    }
```
skofman1: this is not needed because the previous line is going to trace an throw

src/Ng/CommandHelpers.cs
```diff
@@ -45,275 +55,269 @@ private static void TraceRequiredArgument(string name)
...
+
+                    return new FileStorageFactory(storageBaseAddress, storagePath) { Verbose = verbose };
                 }
+                else if (storageType.Equals("Azure", StringComparison.InvariantCultureIgnoreCase))
```
skofman1: >Azure [](start = 45, length = 5)

const

src/Ng/CommandHelpers.cs
```diff
@@ -324,99 +328,78 @@ public static StorageFactory CreateSuffixedStorageFactory(string suffix, IDictio
...
-                return null;
-            }
+                string storageType;
+                TryGetArgument(arguments, Constants.StorageType, out storageType, required: true);
 
-            if (storageType.Equals("File", StringComparison.InvariantCultureIgnoreCase))
-            {
-                string storagePath;
-                if (!arguments.TryGetValue("-storagePath" + suffix, out storagePath))
+                if (storageType.Equals("File", StringComparison.InvariantCultureIgnoreCase))
```
skofman1: >File [](start = 40, length = 4)

const

src/Ng/CommandHelpers.cs
```diff
@@ -324,99 +328,78 @@ public static StorageFactory CreateSuffixedStorageFactory(string suffix, IDictio
...
-                if (storageBaseAddress == null)
-                {
-                    TraceRequiredArgument("-storageBaseAddress" + suffix);
-                    return null;
-                }
+                    if (storageBaseAddress == null)
+                    {
+                        TraceRequiredArgument("-storageBaseAddress" + suffix);
+                        return null;
+                    }
```
skofman1: same as above, this code is not needed here

src/Ng/CommandHelpers.cs
```diff
@@ -324,99 +328,78 @@ public static StorageFactory CreateSuffixedStorageFactory(string suffix, IDictio
...
-                if (!arguments.TryGetValue("-storageAccountName" + suffix, out storageAccountName))
-                {
-                    TraceRequiredArgument("-storageAccountName" + suffix);
-                    return null;
+                    return new FileStorageFactory(storageBaseAddress, storagePath) { Verbose = verbose };
                 }
-
-                string storageKeyValue;
-                if (!arguments.TryGetValue("-storageKeyValue" + suffix, out storageKeyValue))
+                else if (storageType.Equals("Azure", StringComparison.InvariantCultureIgnoreCase))
```
skofman1: >Azure [](start = 45, length = 5)

const

src/Ng/CommandHelpers.cs
```diff
@@ -324,99 +328,78 @@ public static StorageFactory CreateSuffixedStorageFactory(string suffix, IDictio
...
-            string luceneResetStr = "false";
-            if (arguments.TryGetValue("-luceneReset", out luceneResetStr))
-            {
-                return (luceneResetStr.Equals("true", StringComparison.InvariantCultureIgnoreCase));
-            }
-            else
-            {
-                return false;
-            }
+            return GetBool(arguments, Constants.LuceneReset, false);
```
skofman1: >false [](start = 61, length = 5)

named parameter

src/Ng/CommandHelpers.cs
```diff
@@ -440,74 +423,61 @@ public static Lucene.Net.Store.Directory GetCopyDestLuceneDirectory(IDictionary<
...
-                return null;
-            }
+                string luceneDirectoryType;
+                TryGetArgument(arguments, names[Constants.DirectoryType], out luceneDirectoryType, required: true);
 
-            if (luceneDirectoryType.Equals("File", StringComparison.InvariantCultureIgnoreCase))
-            {
-                string lucenePath;
-                if (!arguments.TryGetValue(names["path"], out lucenePath))
+                if (luceneDirectoryType.Equals("File", StringComparison.InvariantCultureIgnoreCase))
```
skofman1: >File [](start = 48, length = 4)

const

src/Ng/CommandHelpers.cs
```diff
@@ -440,74 +423,61 @@ public static Lucene.Net.Store.Directory GetCopyDestLuceneDirectory(IDictionary<
...
-                if (!arguments.TryGetValue(names["storageAccountName"], out luceneStorageAccountName))
-                {
-                    TraceRequiredArgument(names["storageAccountName"]);
-                    return null;
+                    return new SimpleFSDirectory(directoryInfo);
                 }
-
-                string luceneStorageKeyValue;
-                if (!arguments.TryGetValue(names["storageKeyValue"], out luceneStorageKeyValue))
+                else if (luceneDirectoryType.Equals("Azure", StringComparison.InvariantCultureIgnoreCase))
```
skofman1: >Azure [](start = 53, length = 5)

azure

src/Ng/CommandHelpers.cs
```diff
@@ -45,275 +55,269 @@ private static void TraceRequiredArgument(string name)
...
+                }
+
+                value = null;
+                return false;
+            }
+
+            return true;
+        }
+
+        public static SecretInjector GetSecretInjector(IDictionary<string, string> arguments)
```
skofman1: should be private

src/Ng/CommandHelpers.cs
ryuyu: storageBaseAddress is not the var checked directly above. it is reused and required for File type, but not for Azure type. So we null check here.

---
In reply to: [71207829](https://github.com/NuGet/NuGet.Services.Metadata/pull/77#discussion_r71207829) [](ancestors = 71207829)

src/Ng/CommandHelpers.cs
ryuyu: storageBaseAddress is not the var checked directly above. it is reused and required for File type, but not for Azure type. So we null check here

---
In reply to: [71208452](https://github.com/NuGet/NuGet.Services.Metadata/pull/77#discussion_r71208452) [](ancestors = 71208452)

src/Ng/CommandHelpers.cs
ryuyu: It seems like there are 2 cases here:
1. Parameter is not specified. In this case, we should fall back to the default value.
2. Parameter is specified, but specified value is not a valid value. In this case, we should fail.

Is this the behaviour that is expected?

---
In reply to: [71022687](https://github.com/NuGet/NuGet.Services.Metadata/pull/77#discussion_r71022687) [](ancestors = 71022687)

skofman1: yes

---
In reply to: [71233002](https://github.com/NuGet/NuGet.Services.Metadata/pull/77#discussion_r71233002) [](ancestors = 71233002,71022687)

src/Ng/CommandHelpers.cs
ryuyu: It looks like a lot of places that call these functions (not specifically this one) use the null value to indicate that it should do a printUsage call. I am reluctant to change this for all of these with this changeset. We should re-evaluate how/when we should fail and when printUsage should be called for this globally.

---
In reply to: [71024082](https://github.com/NuGet/NuGet.Services.Metadata/pull/77#discussion_r71024082) [](ancestors = 71024082)

src/Ng/Program.cs
```diff
@@ -27,22 +27,22 @@ public static void Main(string[] args)
...
             _logger = loggerFactory.CreateLogger<Program>();
 
-            var cancellationTokenSource = new CancellationTokenSource();
             try
             {
+                // Get arguments
+                var arguments = CommandHelpers.GetArguments(args, 0);
+
+                // Configure ApplicationInsights
+                ApplicationInsights.Initialize(CommandHelpers.GetApplicationInsightsInstrumentationKey(arguments));
```
skofman1: >Initialize [](start = 36, length = 10)

I suspect that this line has to run before CreateLoggerFactory. U can verify this by testing the code and using this AI app: [https://ms.portal.azure.com/?resourceMenuPerf=true#resource/subscriptions/fa8b3229-dc0f-4633-a285-ad597076921d/resourceGroups/NugetDevGroup/providers/microsoft.insights/components/StageManager/overview](https://ms.portal.azure.com/?resourceMenuPerf=true#resource/subscriptions/fa8b3229-dc0f-4633-a285-ad597076921d/resourceGroups/NugetDevGroup/providers/microsoft.insights/components/StageManager/overview)

src/Ng/Program.cs
```diff
@@ -27,22 +27,22 @@ public static void Main(string[] args)
...
-
-            // Configure ApplicationInsights
-            ApplicationInsights.Initialize(CommandHelpers.GetApplicationInsightsInstrumentationKey(arguments));
-
             // Create an ILoggerFactory
             var loggerConfiguration = LoggingSetup.CreateDefaultLoggerConfiguration(withConsoleLogger: true);
             var loggerFactory = LoggingSetup.CreateLoggerFactory(loggerConfiguration);
 
             // Create a logger that is scoped to this class (only)
             _logger = loggerFactory.CreateLogger<Program>();
```
skofman1: put everything in try catch.. always a good practice

src/Ng/Program.cs
```diff
@@ -53,43 +53,47 @@ public static void Main(string[] args)
...
                     case "lightning":
-                        Lightning.Run(args, cancellationTokenSource.Token);
+                        Lightning.Run(arguments, cancellationTokenSource.Token);
                         break;
                     default:
                         PrintUsage();
                         break;
                 }
             }
+            catch (ArgumentException ae)
```
skofman1: >ArgumentException [](start = 19, length = 17)

also print usage in this case

skofman1:
>             Console.WriteLine("Usage: ng [package2catalog|feed2catalog|catalog2registration|catalog2lucene|catalog2dnx|frameworkcompatibility|copylucene|checklucene|clearlucene|db2lucene|lightning]");

update the usage string to include KeyVault arguments

---
Refers to: src/Ng/Program.cs:19 in ad059df. [](commit_id = ad059df7da9dfa29d18f68b5c9a75ac2edc2eded, deletion_comment = False)

skofman1:
> 	set NUGETJOBS_STORAGE_PRIMARY=#{Jobs.ngcatalog2dnx.Storage.Primary}

remove those everywhere

---
Refers to: src/Ng/Scripts/Catalog2DnxV3.cmd:8 in ad059df. [](commit_id = ad059df7da9dfa29d18f68b5c9a75ac2edc2eded, deletion_comment = True)

src/Ng/Program.cs
ryuyu: The reason this wasn't done was because the catch statements of the try block below both try to log error using this logger. It looks like we might have to change that if this require that AI initialize before it works correctly.

---
In reply to: [71253697](https://github.com/NuGet/NuGet.Services.Metadata/pull/77#discussion_r71253697) [](ancestors = 71253697)

skofman1: Then a separate try catch with command line logging

---
In reply to: [71394658](https://github.com/NuGet/NuGet.Services.Metadata/pull/77#discussion_r71394658) [](ancestors = 71394658,71253697)

src/Ng/Program.cs
ryuyu: I have moved a few things around and put this back where it was before after app insights initialization.

---
In reply to: [71253386](https://github.com/NuGet/NuGet.Services.Metadata/pull/77#discussion_r71253386) [](ancestors = 71253386)

skofman1:
Good job :) :sheep: :ship: :shipit:

PR #78, opened 2016-07-19
KeyVault integration

skofman1:
@maartenba @xavierdecoster @scottbommarito @ryuyu


src/NuGet.Services.BasicSearch/ConfigurationService.cs
```diff
@@ -35,27 +59,40 @@ public bool TryGet(string key, out string value)
                 {
                     var cloudKey = key.Replace(':', '-'); // no ':' supported in cloud services
                     value = SafeRoleEnvironment.GetConfigurationSettingValue(cloudKey);
-                    return true;
+                    setValue = true;
```
scottbommarito: Are you sure you don't want a
```
if (value != null)
{
    setValue = true;
}
```
here?

Is it impossible for `GetConfigurationSettingValue(cloudKey)` to return null?

src/NuGet.Services.BasicSearch/SecretReader/ISecretReaderFactory.cs
```diff
@@ -0,0 +1,15 @@
...
+using NuGet.Indexing;
+using NuGet.Services.KeyVault;
+
+namespace NuGet.Services.BasicSearch
+{
+    internal interface ISecretReaderFactory
+    {
+        ISecretInjector CreateSecretInjector(ISecretReader secretReader);
+
+        ISecretReader CreateSecterReader(IConfiguration configuration);
```
scottbommarito: `Sector` instead of `Secter`?

src/NuGet.Services.BasicSearch/SecretReader/SecretReaderFactory.cs
```diff
@@ -0,0 +1,56 @@
...
+
+using System;
+using NuGet.Indexing;
+using NuGet.Services.KeyVault;
+
+namespace NuGet.Services.BasicSearch
+{
+    public class SecretReaderFactory : ISecretReaderFactory
+    {
+        public const string VaultName = "keyVault:VaultName";
```
scottbommarito: I think it might be better to give these constants a name that implies that they are strings used to access values in the config rather than the values themselves, e.g. `public const string ConfigVaultName = "keyVault:VaultName";`

scottbommarito:
A couple pretty minor comments, LGTM.

src/NuGet.Services.BasicSearch/ConfigurationService.cs
skofman1: Keeping the same behavior. If we are running in Azure, the RoleEnvironment configuration is the ultimate truth. 

---
In reply to: [71400834](https://github.com/NuGet/NuGet.Services.Metadata/pull/78#discussion_r71400834) [](ancestors = 71400834)

src/NuGet.Services.BasicSearch/SecretReader/ISecretReaderFactory.cs
skofman1: good catch. Thanks :)

---
In reply to: [71401074](https://github.com/NuGet/NuGet.Services.Metadata/pull/78#discussion_r71401074) [](ancestors = 71401074)

src/NuGet.Services.BasicSearch/ConfigurationService.cs
maartenba: Can we rename it to `hasSetValue` or something like that? `setValue` got my brain into thinking that the variable means we have to set a value later on.

src/NuGet.Services.BasicSearch/ConfigurationService.cs
```diff
@@ -26,7 +46,13 @@ public string Get(string key)
         /// <returns>True if the value is found. False, otherwise.</returns>
         public bool TryGet(string key, out string value)
         {
-            if (key == null) throw new ArgumentNullException("key");
+            if (key == null)
+            {
+                throw new ArgumentNullException("key");
```
maartenba: `nameof`

maartenba:
Few nits

src/NuGet.Services.BasicSearch/ConfigurationService.cs
```diff
@@ -35,27 +61,40 @@ public bool TryGet(string key, out string value)
...
+                }
+                else
+                {
+                    // Get value from AppSettings
+                    value = ConfigurationManager.AppSettings[key];
+                    if (value != null)
+                    {
+                        setValue = true;
+                    }
+                }
```
shishirx34: How about:
```
value = Environment.GetEnvironmentVariable(key);
value = value ?? ConfigurationManager.AppSettings[key];
setValue = value != null;
```

shishirx34:
One minor comment, Looks Good! :shipit:

PR #79, opened 2016-07-22
Resolve "invalid portable framework" exception on INT

maartenba:
* Ignoring ArgumentException that denotes invalid portable framework on
purpose
* Update NuGet.* dependencies to support newer TFM's

@skofman1 @xavierdecoster


xavierdecoster:
LGTM :shipit:

src/Catalog/NuGet.Services.Metadata.Catalog.csproj
```diff
@@ -89,7 +89,7 @@
       <Private>True</Private>
     </Reference>
     <Reference Include="NuGet.Versioning, Version=3.5.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\NuGet.Versioning.3.5.0-beta-final\lib\net45\NuGet.Versioning.dll</HintPath>
+      <HintPath>..\..\packages\NuGet.Versioning.3.5.0-beta2-1484\lib\net45\NuGet.Versioning.dll</HintPath>
```
skofman1: How r the client version and server version correlate? Should we support in the server versions that weren't released yet? This my cause the server to succeed in upload but the client to fail in install..

maartenba: They shouldn't correlate, although vLatest is always good to have all TFM's supported.

skofman1:
:shipit:

PR #82, opened 2016-07-28
Append querystring to package binary download to bust CDN cache

maartenba:
Partial fix for https://github.com/NuGet/NuGetGallery/issues/3168
Adds a query string parameter to the package binary download so we fetch
a fresh cache version from the CDN.

@scottbommarito @skofman1 @ryuyu


src/Ng/Feed2Catalog.cs
```diff
@@ -304,7 +304,10 @@ static DateTime ForceUtc(DateTime date)
             {
                 foreach (var packageItem in entry.Value)
                 {
-                    var response = await client.GetAsync(packageItem.ContentUri, cancellationToken);
+                    var packageUriBuilder = new UriBuilder(packageItem.ContentUri);
```
skofman1: >packageUriBuilder [](start = 24, length = 17)

can u please add a comment explaining this change (for future generations)

skofman1:
:shipit:

src/Ng/Feed2Catalog.cs
maartenba: Added :-)

scottbommarito:
LGTM!

ryuyu:
LGTM

shishirx34:
Looks Good.

PR #83, opened 2016-08-02
Run BinSkim as part of the build

maartenba:
@scottbommarito @skofman1


scottbommarito:
seems good to me, will test locally now

skofman1:
:shipit:

PR #85, opened 2016-08-10
Improve logging for V3 jobs

skofman1:
1. Added flush of ApplicationInsights buffer to make sure we get the last exception message
2. Reduced severity of AI logging to make sure we get traces
3. Propagated ILogger in all ng.exe code to get prettier logs (packages still use Trace)
4. Changed NG.exe to be console application instead of windows application, so we can see console errors if needed (shouldn't impact job run)
5. Next step: add job name to AI events using serilog context: https://github.com/NuGet/NuGet.Staging/pull/70


skofman1:
@maartenba @xavierdecoster @shishirx34 @ryuyu @scottbommarito

src/Ng/Catalog2Dnx.cs
```diff
@@ -43,7 +42,6 @@ public Catalog2Dnx(ILoggerFactory loggerFactory)
                 while (run);
```
maartenba: Can we add
```
{
// ...
}```

File: src/Ng/Catalog2Lucene.cs [hunk 2]
```diff
@@ -91,35 +100,35 @@ public static void PrintUsage()
...
             string storageBaseAddress = CommandHelpers.GetStorageBaseAddress(arguments, required: false);
 
-            Trace.TraceInformation("CONFIG source: \"{0}\" registration: \"{1}\" catalogBaseAddress: \"{2}\" storageBaseAddress: \"{3}\" interval: {4} seconds", source, registration ?? "(null)", catalogBaseAddress ?? "(null)", storageBaseAddress ?? "(null)", interval);
+            _logger.LogInformation("CONFIG source: \"{ConfigSource}\" registration: \"{Registration}\"" +
+                                   " catalogBaseAddress: \"{CatalogBaseAddress}\" storageBaseAddress: \"{StorageBaseAddress}\"" +
+                                   " interval: {Interval} seconds",
+                                   source, 
+                                   registration ?? "(null)",
+                                   catalogBaseAddress ?? "(null)",
+                                   storageBaseAddress ?? "(null)", interval);
```
maartenba: `interval` on newline for consistency?

maartenba:
Few minor nits, otherwise LGTM

xavierdecoster:
Less statics #ftw! LGTM. 
Wondering: do we still need the separate Trace statements?

src/Ng/Catalog2Dnx.cs
skofman1: This is a do-while...

maartenba: Missed that in the diff. Nevermind :-)

skofman1:
not sure what you mean? Trace statements?

---
In reply to: [238814191](https://github.com/NuGet/NuGet.Services.Metadata/pull/85#issuecomment-238814191) [](ancestors = 238814191)

xavierdecoster:
I meant calls into System.Diagnostics.Trace, e.g. https://github.com/NuGet/NuGet.Services.Metadata/pull/85/files#diff-7ac5a7fdf140953e0b3ad305de87bc58R125

shishirx34:
Looks good.

scottbommarito:
There's a couple places in this where you use `string.Format` in the Logs and a couple places where you don't in this file. I personally prefer `string.Format` better, because I'm under the impression that it shows the formatted message in the name column of AI and not the customDimensions column, but I think the established idiom throughout our logging is to not use `string.Format` so we should be consistent.

skofman1:
There is a listener that writes from Trace to a log file. I wanted to make sure that if the logger fails to initialize, or we dont see the error in AI for some reason, it would appear in the log file.

---
In reply to: [238951316](https://github.com/NuGet/NuGet.Services.Metadata/pull/85#issuecomment-238951316) [](ancestors = 238951316)

PR #87, opened 2016-08-19
Search by PackageType API Proposal

ryuyu:
This is the preliminary implementation of search by packageType for the search service, detailed at https://github.com/NuGet/Home/wiki/Search-by-Package-Type-and-Query-Language-Surfacing

@maartenba @joelverhagen @skofman1 @yishaigalatzer


joelverhagen:
This seems to have commits that are not yours -- is this intentional?

src/Catalog/context/Catalog.json
```diff
@@ -9,7 +9,8 @@
         "categories": { "@container" : "@set" },
         "entries": { "@container" : "@set" },
         "links": { "@container" : "@set" },
-        "tags": { "@container": "@set" },
+      "tags": { "@container": "@set" },
+      "packageTypes": {"@container":  "@set" },
```
shishirx34: nit: alignment.

src/Catalog/context/Registration.json
```diff
@@ -16,7 +16,15 @@
...
-        "dependencies": { "@container" : "@set", "@id": "dependency" },
+      "dependencies": {
+        "@container": "@set",
+        "@id": "dependency"
+      },
+
+      "packageTypes": {
+        "@container": "@set",
+        "@id": "packageType"
+      },
```
shishirx34: nit: alignment

ryuyu:
Oh that's strange. I wonder where these are from. I looked over the changes and it looks like there isn't anything included that wasn't a change that I made. I'll see if i can clean up history.

scottbommarito:
Did you branch this from master and then try to merge into dev?

src/NuGet.Indexing/Extraction/DocumentCreator.cs
```diff
@@ -261,6 +263,56 @@ private static void AddRequiresLicenseAcceptance(Document document, IDictionary<
...
+            StringBuilder searchIndex = new StringBuilder();
+            if (package.TryGetValue("flattenedPackageTypes", out value))
+            {
+                if (!string.IsNullOrWhiteSpace(value))
+                {
+                    AddField(document, "FlattenedPackageTypes", value, Field.Index.NOT_ANALYZED);
+
+                    using (var textWriter = new StringWriter())
+                    {
+                        using (var jsonWriter = new JsonTextWriter(textWriter))
```
shishirx34: not super important, but I think you can avoid scoping for single lines.

```
using (var textWriter = new StringWriter())
using (var jsonWriter = new JsonTextWriter(textWriter)) {....}
```

src/NuGet.Indexing/Extraction/DocumentCreator.cs
```diff
@@ -261,6 +263,56 @@ private static void AddRequiresLicenseAcceptance(Document document, IDictionary<
...
+
+                                    jsonWriter.WriteEndObject();
+                                }
+                            }
+
+                            jsonWriter.WriteEndArray();
+                            jsonWriter.Flush();
+                            textWriter.Flush();
+
+                            var packageTypes = textWriter.ToString();
```
shishirx34: I think you should be able to optimize similar thing here that AddFlattenedPackageTypes did, that we talked offline.

```
var searchIndex = string.Join(" ", value.Split("|").Select(packageType => packageType.Split(':')[0]).ToList());
```

Might need to add a length check in there as well I guess. If that affects readability a lot, I am fine with the way it is.

src/Ng/SearchIndexFromCatalogCollector.cs
```diff
@@ -135,7 +135,7 @@ private void ProcessCatalogItems(IndexWriter indexWriter, IEnumerable<JObject> c
 
             foreach (JObject catalogItem in catalogItems)
             {
-                _logger.LogInformation("Process CatalogItem {CatalogItem}", catalogItem["@id"]);
+                _logger.LogInformation($"Process CatalogItem {catalogItem["@id"]}", catalogItem["@id"]);
```
skofman1: >gInformation($"Process CatalogIte [](start = 26, length = 33)

dont change this, it's in Serilog format

src/NuGet.Indexing/ResponseFormatter.cs
```diff
@@ -79,6 +79,8 @@ private static void WriteData(JsonWriter jsonWriter, NuGetIndexSearcher searcher
                 WriteDocumentValue(jsonWriter, "licenseUrl", document, "LicenseUrl");
                 WriteDocumentValue(jsonWriter, "projectUrl", document, "ProjectUrl");
                 WriteDocumentValueAsArray(jsonWriter, "tags", document, "Tags");
+                // WriteDocumentValueAsArray(jsonWriter, "packageTypes", document, "PackageTypes");
```
skofman1: > // Wr [](start = 15, length = 6)

?

src/NuGet.Indexing/Extraction/DocumentCreator.cs
```diff
@@ -45,6 +45,8 @@ public static Document CreateDocument(IDictionary<string, string> package)
             AddField(document, "Summary", package, "summary", Field.Index.ANALYZED);
             AddField(document, "Tags", package, "tags", Field.Index.ANALYZED, 2.0f);
             AddField(document, "Authors", package, "authors", Field.Index.ANALYZED);
+            // AddField(document, "PackageTypes", package, "packageTypes", Field.Index.ANALYZED);
```
skofman1: >// AddF [](start = 12, length = 7)

?

src/NuGet.Indexing/Querys/DownloadsScoreProvider.cs
```diff
@@ -81,7 +81,7 @@ private float AdjustByDownloads(int docId)
         public static float DownloadScore(long totalDownloads,
             QueryBoostingContext context)
         {
-            if (!context.BoostByDownloads)
+            if (context==null || !context.BoostByDownloads)
```
skofman1: >==n [](start = 23, length = 3)

spaces

src/NuGet.Services.BasicSearch/Web.config
```diff
@@ -14,10 +14,10 @@
     <sessionState mode="Off" />
   </system.web>
   <appSettings>
-    <!--
-    <add key="Local.Lucene.Directory" value="C:\nuget\SearchData\Lucene-v2v3" />
+
+    <add key="Local.Lucene.Directory" value="C:\nuget\newLucene" /><!--"C:\nuget\SearchData\Lucene-v2v3" /-->
```
skofman1: >:\nuget\newLuce [](start = 46, length = 15)

undo

skofman1:
> namespace NuGet.IndexingTests

u need a LOT of UTs

---
Refers to: tests/NuGet.IndexingTests/Extraction/DocumentCreatorTests.cs:11 in 45a04ff. [](commit_id = 45a04ff898f4117c43b73960df7b6d7b8fce4438, deletion_comment = False)

joelverhagen:
> I'll see if i can clean up history.

If your branch is hanging off of `master` in Git, you need to select that as the base branch when creating the PR. Right now, GitHub thinks your PR should be compared to `dev`.

scottbommarito:
Yeah, I think the fastest way to fix this is to create a branch off dev then cherry-pick all the commits you made from this branch. You might also want to include Hari's update readme commit because that looks like it was made to master directly instead of going through dev first.

src/Ng/SearchIndexFromCatalogCollector.cs
```diff
@@ -44,7 +44,7 @@ public SearchIndexFromCatalogCollector(ILogger logger, Uri index, IndexWriter in
             IEnumerable<JObject> catalogItems = await FetchCatalogItems(client, items, cancellationToken);
 
             var numDocs = _indexWriter.NumDocs();
-            _logger.LogInformation(string.Format("Index contains {0} documents.", _indexWriter.NumDocs()));
+            _logger.LogInformation(string.Format("Index contains {0} documents.", numDocs));
```
scottbommarito: I think this is supposed to be

`_logger.LogInformation($"Index contains {NumDocs} documents.", numDocs);`

to follow Serilog format, but I don't think it's that big of a deal.

src/NuGet.Indexing/Extraction/DocumentCreator.cs
```diff
@@ -261,6 +263,56 @@ private static void AddRequiresLicenseAcceptance(Document document, IDictionary<
             }
         }
 
+        private static void AddPackageTypes(Document document, IDictionary<string, string> package)
+        {
+            string value;
```
scottbommarito: maybe rename this to `flattenedPackageTypes` so it's more clear what it is at a glance?

src/NuGet.Indexing/Extraction/DocumentCreator.cs
```diff
@@ -261,6 +263,56 @@ private static void AddRequiresLicenseAcceptance(Document document, IDictionary<
...
+                                {
+                                    jsonWriter.WriteStartObject();
+                                    jsonWriter.WritePropertyName("Type");
+                                    jsonWriter.WriteValue(fields[0]);
+                                    searchIndex.Append(fields[0]);
+                                    searchIndex.Append(" ");
+                                    if (fields.Length > 1)
+                                    {
+                                        jsonWriter.WritePropertyName("Version");
+                                        jsonWriter.WriteValue(fields[1]);
```
scottbommarito: So we only care about the first two fields? Is it possible for there to be more than two?

src/NuGet.Indexing/NuGetQuery.cs
```diff
@@ -194,6 +197,11 @@ private static void AuthorClause(BooleanQuery query, Analyzer analyzer, IEnumera
             query.Add(ConstructClauseQuery(analyzer, "Authors", values), occur);
         }
 
+        private static void PackageTypeClause(BooleanQuery query, Analyzer analyzer, IEnumerable<string> values, Occur occur)
+        {
+            query.Add(ConstructClauseQuery(analyzer, "PackageTypesIndex", values), occur);
```
scottbommarito: General question about this file...
Should we put the `"PackageTypesIndex"`, `"Authors"`, and similar query names into constants? Is there already a constant with these values?

src/NuGet.Indexing/ResponseFormatter.cs
scottbommarito: Actually I think is because of the `WritePackageTypes` method he created below.

src/NuGet.Services.BasicSearch/Web.config
```diff
@@ -14,10 +14,10 @@
     <sessionState mode="Off" />
   </system.web>
   <appSettings>
-    <!--
-    <add key="Local.Lucene.Directory" value="C:\nuget\SearchData\Lucene-v2v3" />
+
```
scottbommarito: Why was this commented out before? Why are we uncommenting it?

src/Ng/SearchIndexFromCatalogCollector.cs
ryuyu: reverted. Was for me to get local logging as this line loges Process CatalogItem [] to console logging.

---
In reply to: [75550188](https://github.com/NuGet/NuGet.Services.Metadata/pull/87#discussion_r75550188) [](ancestors = 75550188)

src/NuGet.Indexing/ResponseFormatter.cs
ryuyu: This was an older implementation that didn't save the packageTypes as an object, but as a simple array. Removed.

---
In reply to: [75550543](https://github.com/NuGet/NuGet.Services.Metadata/pull/87#discussion_r75550543) [](ancestors = 75550543)

src/NuGet.Indexing/Extraction/DocumentCreator.cs
ryuyu: As of now, we only care about Type and Version. If we decide to add more metadata here, we can add it here, but as of now, it would do nothing.

---
In reply to: [75555637](https://github.com/NuGet/NuGet.Services.Metadata/pull/87#discussion_r75555637) [](ancestors = 75555637)

src/NuGet.Services.BasicSearch/Web.config
ryuyu: Reverted.

---
In reply to: [75552446](https://github.com/NuGet/NuGet.Services.Metadata/pull/87#discussion_r75552446) [](ancestors = 75552446)

src/NuGet.Indexing/Extraction/DocumentCreator.cs
ryuyu: Same as other. Leftover from an older, less flexible implementation. deleted.

---
In reply to: [75551979](https://github.com/NuGet/NuGet.Services.Metadata/pull/87#discussion_r75551979) [](ancestors = 75551979)

src/NuGet.Services.BasicSearch/Web.config
ryuyu: Commented because it isn't used in deployed env. Uncommented to test locally. Reverted.

---
In reply to: [75556589](https://github.com/NuGet/NuGet.Services.Metadata/pull/87#discussion_r75556589) [](ancestors = 75556589)

src/NuGet.Indexing/Extraction/DocumentCreator.cs
scottbommarito: Might be valuable to put a comment about this there...typically when you see a split like that you assume there's a loop so it's easy to overlook.

src/Ng/SearchIndexFromCatalogCollector.cs
ryuyu: Svetlana, any followup here? i only changed the call to _indexWriter.NumDocs() to the actual value. What should this string actually be?

---
In reply to: [75553808](https://github.com/NuGet/NuGet.Services.Metadata/pull/87#discussion_r75553808) [](ancestors = 75553808)

src/NuGet.Indexing/Extraction/DocumentCreator.cs
ryuyu: done.

---
In reply to: [75554919](https://github.com/NuGet/NuGet.Services.Metadata/pull/87#discussion_r75554919) [](ancestors = 75554919)

src/Ng/SearchIndexFromCatalogCollector.cs
scottbommarito: paging dr. @skofman1

src/NuGet.Indexing/Querys/DownloadsScoreProvider.cs
ryuyu: Done.

---
In reply to: [75552402](https://github.com/NuGet/NuGet.Services.Metadata/pull/87#discussion_r75552402) [](ancestors = 75552402)

src/NuGet.Indexing/Extraction/DocumentCreator.cs
ryuyu: processing flattenedDependencies below does exactly the same thing. And assuming there's a loop is fine and all, but it is fairly obvious from even a cursory glance at the code that this is not being looped on. I would rather not clutter with comments things that are obvious from examination.

src/NuGet.Indexing/Extraction/DocumentCreator.cs
```diff
@@ -261,6 +262,57 @@ private static void AddRequiresLicenseAcceptance(Document document, IDictionary<
...
+                            foreach (var packageType in flattenedPackageTypes.Split('|'))
+                            {
+                                string[] fields = packageType.Split(':');
+                                if (fields.Length > 0)
+                                {
+                                    jsonWriter.WriteStartObject();
+                                    jsonWriter.WritePropertyName("Type");
+                                    jsonWriter.WriteValue(fields[0]);
+                                    searchIndex.Append(fields[0]);
+                                    searchIndex.Append(" ");
```
shishirx34: I think in general it would be a good idea to create an array list of the required fields for `searchIndex` and then convert it to a string delimited by <space> using `string.Join`. Personally I think that is a more readable format, no need to remove the added ending space in StringBuilder below.

To each his own, will leave it up to you.

src/NuGet.Indexing/Extraction/DocumentCreator.cs
```diff
@@ -261,6 +262,61 @@ private static void AddRequiresLicenseAcceptance(Document document, IDictionary<
             }
         }
 
+        private static void AddPackageTypes(Document document, IDictionary<string, string> package)
+        {
+            string flattenedPackageTypes;
+            StringBuilder searchIndex = new StringBuilder();
+            if (package.TryGetValue("flattenedPackageTypes", out flattenedPackageTypes))
```
joelverhagen: This `IDictionary<string, string>` is just an intermediate data structure, right? Why do that `flattenedPackageTypes` here? In fact, why have a flattened format at all? This is a new field everywhere. I think we should have a consistent model unless we have a very compelling reason.

src/NuGet.Indexing/PackageTypesAnalyzer.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Collections.Generic;
+using System.IO;
+
+namespace NuGet.Indexing
+{
+    class PackageTypesAnalyzer : Analyzer
+    {
+        public override TokenStream TokenStream(string fieldName, TextReader reader)
+        {
+            return new LowerCaseFilter(new DotTokenizer(reader));
```
joelverhagen: Why `DotTokenizer`? These should be verbatim strings right?

src/NuGet.Indexing/Extraction/DocumentCreator.cs
ryuyu: This was to have consistency with a flattened format that the dependencies also seemed to use, and was for the V2 API. As we discussed in the other thread, I'm actually okay with getting rid of this completely if we don't need this in the V2 API.

---
In reply to: [75567354](https://github.com/NuGet/NuGet.Services.Metadata/pull/87#discussion_r75567354) [](ancestors = 75567354)

src/NuGet.Indexing/PackageTypesAnalyzer.cs
ryuyu: Excellent point. Fixed.

src/NuGet.Indexing/Extraction/CatalogPackageReader.cs
```diff
@@ -46,6 +46,23 @@ public override IEnumerable<string> GetFiles()
...
+            var packageTypes = new List<PackageType>();
+            var array = _catalogItem.GetJArray("packageTypes");
+            if (array == null)
+            {
+                return packageTypes;
+            }
+
+            foreach (var entry in array)
+            {
+                packageTypes.Add(new PackageType((string)entry["type"], new Version((string)entry["version"])));
```
maartenba: Is version always defined or can it be null?

src/NuGet.Indexing/Extraction/DocumentCreator.cs
```diff
@@ -261,6 +262,61 @@ private static void AddRequiresLicenseAcceptance(Document document, IDictionary<
             }
         }
 
+        private static void AddPackageTypes(Document document, IDictionary<string, string> package)
```
maartenba: Not sure we want to store JSON in the search index. Can we just store the raw values and have the API generate JSON if needed? All JSON formatting is additional cruft that will make the index grow in size.

src/NuGet.Indexing/Extraction/DocumentCreator.cs
```diff
@@ -261,6 +262,61 @@ private static void AddRequiresLicenseAcceptance(Document document, IDictionary<
...
+                }
+            }
+
+            // remove the trailing space from the index before saving
+            if (searchIndex.Length > 0)
+            {
+                searchIndex.Length--;
+            }
+
+            AddField(document, "PackageTypesIndex", searchIndex.ToString(), Field.Index.ANALYZED);
```
maartenba: In addition, analyzing the JSON data is a lot of work for Lucene, both during index and during query. Might want to make this field contain just the values instead of full-blwon JSON.

src/NuGet.Indexing/NuGetQuery.cs
```diff
@@ -194,6 +197,11 @@ private static void AuthorClause(BooleanQuery query, Analyzer analyzer, IEnumera
             query.Add(ConstructClauseQuery(analyzer, "Authors", values), occur);
         }
 
+        private static void PackageTypeClause(BooleanQuery query, Analyzer analyzer, IEnumerable<string> values, Occur occur)
```
maartenba: In terms of query performance, you may want to make the package types behave like the owners (make it a filtered query).

If it's a clause, Lucene will have more documents to verify. If it's a filtered query, result set is made very small at the start of the query and Lucene will only have a few documents to check.

maartenba:
Will package types surface in registration blobs?

maartenba:
Will package types surface in V3 search only or also on V2 feeds? (e.g. in gallery?)

src/NuGet.Indexing/Extraction/CatalogPackageReader.cs
ryuyu: Version should always be defined. It will either be defined by the package metadata, or filled in with a default value.

src/NuGet.Indexing/Extraction/DocumentCreator.cs
ryuyu: The PackageTypesIndex actually is just space separated values. At this point, the generated JSON goes into the packageTypes and is not Analyzed.

ryuyu:
I think the idea is that package types should surface in both searches. @harikmenon Any input on this?

src/NuGet.Indexing/Extraction/DocumentCreator.cs
```diff
@@ -272,37 +272,13 @@ private static void AddPackageTypes(Document document, IDictionary<string, strin
                 {
                     AddField(document, "FlattenedPackageTypes", flattenedPackageTypes, Field.Index.NOT_ANALYZED);
 
-                    using (var textWriter = new StringWriter())
+                    foreach (var packageType in flattenedPackageTypes.Split('|'))
                     {
-                        using (var jsonWriter = new JsonTextWriter(textWriter))
+                        string[] fields = packageType.Split(':');
```
maartenba: `var` all the things :-)

src/NuGet.Indexing/NuGetQuery.cs
```diff
@@ -89,6 +93,15 @@ private static Query ConstructQuery(Dictionary<QueryField, HashSet<string>> clau
                             }
                         }
 
+                        if (packageTypes != null)
```
maartenba: I don't think you need this, unless we want to support querying the package type in free-form queries? (e.g. when searching for "dotnettools", do we want to also search our package type field? If yes, keep this. If no, drop the entire if and its body :-))

src/NuGet.Indexing/NuGetQuery.cs
```diff
@@ -75,7 +75,11 @@ private static Query ConstructQuery(Dictionary<QueryField, HashSet<string>> clau
                         }
                         break;
                     case QueryField.PackageType:
-                        PackageTypeClause(booleanQuery, analyzer, clause.Value, Occur.MUST);
+                        // PackageTypeClause(booleanQuery, analyzer, clause.Value, Occur.MUST);
```
maartenba: See e-mail conversation

```
                case QueryField.PackageType:
                    filters.Add(
                        new QueryWrapperFilter(
                            ConstructClauseQuery(analyzer, "FlattenedPackageTypes", clause.Value, Occur.MUST)));
                    break;
```

maartenba:
Probably the added classes for package types are not needed (cbd624b), see e-mail conversation.

maartenba:
I just thought of a very important thing: regenerating the search index using `Sql2Lucene`. We typically do this every month/2 months to generate a fresh, optimized search index. If we want to have package types in our index, we'll need to store package types in the database, too, so that `Sql2Lucene` can include the package types in the index as well. If we don't, then regenerating the index will basically drop all package type info from the index and render it useless.

src/NuGet.Indexing/Handlers/PackageTypesHandler.cs
```diff
@@ -0,0 +1,83 @@
...
+        public bool SkipDeletes => true;
+
+        public void Begin(IndexReader indexReader)
+        {
+            _knownPackageTypes = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            _packageTypeTuples = new Dictionary<string, IDictionary<string, DynamicDocIdSet>>();
+
+            if (indexReader.GetSequentialSubReaders() != null)
+            {
+                foreach (SegmentReader segmentReader in indexReader.GetSequentialSubReaders())
```
shishirx34: nit: may be extract subreaders before:

```
var sequentialSubReaders = indexReader.GetSequentialSubReaders();
if (sequentialSubReaders != null)
{
    foreach (SegmentReader segmentReader in sequentialSubReaders) {...}
}
```

src/NuGet.Indexing/NuGetQuery.cs
```diff
@@ -215,6 +228,19 @@ private static void PackageTypeClause(BooleanQuery query, Analyzer analyzer, IEn
             }
         }
 
+        private static IEnumerable<Filter> PackageTypeFilters(
+            PackageTypesResult packageTypes,
+            HashSet<string> value)
```
shishirx34: some better name than `value` may be?

maartenba:
Starts to look good :-) Can you also add the plumbing in Sql2Lucene? (add the package types field from database)

maartenba:
Do we have any sample packages? (want to run some tests with this one :))

harikmenon:
@yishaigalatzer whats your thoughts on exposing this in v2 for gallery?

src/NuGet.Indexing/ResponseFormatter.cs
```diff
@@ -96,6 +97,35 @@ private static void WriteData(JsonWriter jsonWriter, NuGetIndexSearcher searcher
...
+            {
+                jsonWriter.WritePropertyName("packageTypes");
+                jsonWriter.WriteStartArray();
+                foreach (var packageType in flattenedPackageTypes.Split('|'))
+                {
+                    string[] fields = packageType.Split(':');
+                    if (fields.Length > 0)
+                    {
+                        jsonWriter.WriteStartObject();
+                        jsonWriter.WritePropertyName("Type");
```
maartenba: Can we lowercase this? And perhaps name it `name` like it is in the .nuspec?

src/NuGet.Indexing/ResponseFormatter.cs
```diff
@@ -96,6 +97,35 @@ private static void WriteData(JsonWriter jsonWriter, NuGetIndexSearcher searcher
...
+                        jsonWriter.WriteValue(fields[0]);
+                        if (fields.Length > 1)
+                        {
+                            jsonWriter.WritePropertyName("Version");
```
maartenba: Can we lowercase this?

src/NuGet.Indexing/ResponseFormatter.cs
ryuyu: done

src/NuGet.Indexing/ResponseFormatter.cs
ryuyu: done

ryuyu:
Ping. Any further comments? I would like to get this merged and testing asap.

maartenba:
Looks good from my side. Others?

src/NuGet.Indexing/app.config
```diff
@@ -67,4 +71,4 @@
       <provider invariantName="System.Data.SqlClient" type="System.Data.Entity.SqlServer.SqlProviderServices, EntityFramework.SqlServer" />
     </providers>
   </entityFramework>
-</configuration>
\ No newline at end of file
+<startup><supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5.2" /></startup></configuration>
```
skofman1: ><supportedRunt [](start = 9, length = 14)

alignment

src/NuGet.Indexing/NuGetQuery.cs
skofman1: ?

---
In reply to: [75626531](https://github.com/NuGet/NuGet.Services.Metadata/pull/87#discussion_r75626531) [](ancestors = 75626531)

src/NuGet.Indexing/PerfEventTracker.cs
```diff
@@ -7,6 +7,7 @@
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
+using NuGetGallery;
```
skofman1: >NuGetGallery [](start = 6, length = 12)

why is the search service depend on the gallery?

src/NuGet.Indexing/ResponseFormatter.cs
```diff
@@ -96,6 +97,35 @@ private static void WriteData(JsonWriter jsonWriter, NuGetIndexSearcher searcher
             jsonWriter.WriteEndArray();
         }
 
+        private static void WritePackageTypes(JsonWriter jsonWriter, Document document)
```
skofman1: >WritePackageTypes [](start = 28, length = 17)

are there UTs for this?

src/NuGet.Indexing/Extraction/CatalogPackageReader.cs
```diff
@@ -46,6 +46,24 @@ public override IEnumerable<string> GetFiles()
             }
         }
 
+        public override IReadOnlyList<PackageType> GetPackageTypes()
```
skofman1: >GetPackageTypes [](start = 51, length = 15)

UTs?

src/NuGet.Indexing/PerfEventTracker.cs
ryuyu: I think this was to allow access to the disposableAction below.
Might it make more sense to just copy the disposableAction into Nuget.Indexing?
Also, i'm not 100% sure how the reference to disposableAction was being resolved before, as the current one is referecning NuGetGallery.DisposableAction.

Maybe it had to do with the reference to NuGet.Core that i think i dropped.

In any case, what do you think we should do here?

src/NuGet.Indexing/NuGetQuery.cs
ryuyu: It's actually not used anymore. I removed it.

---
In reply to: [78817495](https://github.com/NuGet/NuGet.Services.Metadata/pull/87#discussion_r78817495) [](ancestors = 78817495,75626531)

src/NuGet.Indexing/PerfEventTracker.cs
ryuyu: So after looking at this, it looks like NuGet.Indexing has a dependency on NuGetGallery.Core for many things. This just opens up that reference to allow us to use the disposableAction here.

src/NuGet.Indexing/ResponseFormatter.cs
ryuyu: There are currently no UTs (as far as i can tell) for the responseFormatter in general other than the V2/V3 functional tests. I propose that we add these as part of a different PR expanding our testing of this class in general.

---
In reply to: [78818136](https://github.com/NuGet/NuGet.Services.Metadata/pull/87#discussion_r78818136) [](ancestors = 78818136)

src/NuGet.Indexing/Extraction/CatalogPackageReader.cs
ryuyu: There are tests in the NuGet.IndexingTests/Extraction/CatalogPackageMetadataExtractorTests.cs that test this code path.

---
In reply to: [78818424](https://github.com/NuGet/NuGet.Services.Metadata/pull/87#discussion_r78818424) [](ancestors = 78818424)

src/NuGet.Indexing/PerfEventTracker.cs
ryuyu: Repeating here: Indexing already depends on NuGetGallery.Core, and NuGetGallery namespace gives us access to the DisposableAction class used below.

---
In reply to: [78817894](https://github.com/NuGet/NuGet.Services.Metadata/pull/87#discussion_r78817894) [](ancestors = 78817894)

yishaigalatzer:
I think it's fine to add unit tests as a separate PR, but please don't merge this without unit tests merged first (so do another PR, get it signed off and merge it, then rebase this one on top of it)

maartenba:
Is this going in at some point or is it moot?

joelverhagen:
This will not be implemented on the old search service. When the need arises in the future, we can re-implement this on the new Azure Search-based service.

PR #88, opened 2016-08-25
Cache busting Catalog2Dnx

shishirx34:
- Added support to Catalog2Dnx to always load the package's index.json the latest available
- Added support to Catalog2Dnx to load latest .nupkg as well.
- Fixed test to ignore query strings.

/cc @maartenba @xavierdecoster @ryuyu @scottbommarito 

I will verify this change on Dev environment to ensure it is working as expected.


src/Catalog/CommonUtility.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace NuGet.Services.Metadata.Catalog
+{
+    public static class CommonUtility
+    {
+        public static Uri GetNugetCacheBustingUri(Uri originalUri)
+        {
+            return GetNugetCacheBustingUri(originalUri, DateTime.Now.ToString());
+        }
+
+        // Adding the timestamp to the URI as a query string ensures
```
scottbommarito: I think there's a specific form for function comments.

Something like...
```
/// <summary>
/// Adding the timestamp to the URI as a query string ensures a cache miss with the CDN
/// </summary>
/// <returns>Returns a URI which ensures a cache miss with the CDN</returns>
```

src/Catalog/CommonUtility.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace NuGet.Services.Metadata.Catalog
+{
+    public static class CommonUtility
```
scottbommarito: I think you might be able to get away with calling this either `Common` or `Utility` instead of the entire `CommonUtility`

tests/NgTests/Infrastructure/MemoryStorage.cs
```diff
@@ -46,7 +46,8 @@ protected override Task OnSave(Uri resourceUri, StorageContent content, Cancella
         protected override Task<StorageContent> OnLoad(Uri resourceUri, CancellationToken cancellationToken)
         {
             StorageContent content;
-            Content.TryGetValue(resourceUri, out content);
+            // Ignore query strings
```
scottbommarito: This comment is kinda confusing without the context that it used to be just `resourceUri` instead of `new Uri(resourceUri.GetLeftPart(UriPartial.Path))`. I think this segment would be much clearer if you had it as...

```
var resourceUriWithoutQuery = new Uri(resourceUri.GetLeftPart(UriPartial.Path));
Content.TryGetValue(resourceUriWithoutQuery, out content);
```

scottbommarito:
A couple nitpicks. LGTM.

src/Catalog/CommonUtility.cs
shishirx34: It was more of a behavioral comment than a functional, but I might as well add it.

src/Catalog/CommonUtility.cs
shishirx34: I would rather have a common namespace to define utilities, but that would be an overkill. Will rename this to `Utilities`

tests/NgTests/Infrastructure/MemoryStorage.cs
shishirx34: That is an unnecessary assignment, I updated the comment to be more clearer. #wontfix

src/Catalog/Utilities.cs
```diff
@@ -0,0 +1,26 @@
...
+
+using System;
+
+namespace NuGet.Services.Metadata.Catalog
+{
+    public static class Utilities
+    {
+        public static Uri GetNugetCacheBustingUri(Uri originalUri)
+        {
+            return GetNugetCacheBustingUri(originalUri, DateTime.Now.ToString());
```
maartenba: .Now is evil. --> `.UtcNow`.

maartenba:
Oe more nit, other than that LGTM

PR #89, opened 2016-08-30
Fix Cache busting for Catalog2Dnx

shishirx34:
In my previous PR I added cache busting URI changes mistakenly to one of the URI that is loaded from the blob using Azure SDK. This results in a failed load of the blob, as it tries to load a blob with name something like `index.json?nuget-cache=...`.

This PR fixes it to not add cache busting URI for blob load. As well as set `CacheControl` to `no-store` for all the subsequent blobs saved by the Catalog2Dnx job, which ensures no blobs are cached and we get latest data from the blob for `nuspec` and `nupkg` @xavierdecoster - does this seem fine to you?

/cc @xavierdecoster @skofman1 @ryuyu @maartenba 

Edit: It turns out, the CDN also provides gzip compression for accept-encoding headers, setting `no-store` to nupkg and nuspec would break that behavior, so keeping it 2min TTL on those blobs. Since we use HttpClient.GetAsync to access it, using cache busting for it would just work.


ryuyu:
:shipit:

xavierdecoster:
LGTM. :shipit:

shishirx34:
I verified the changes and they look good. Merging changes in.

PR #90, opened 2016-09-01
 Ng job restructuring and key rotation logic

scottbommarito:
Restructured jobs to all inherit from a common base class, `NgJob`. Jobs that are intended to loop inherit from `LoopingNgJob`.

`LoopingNgJob` has logic to reinitialize during its sleep interval, thus refreshing the config values from KeyVault. Restructuring the jobs allowed this to be done easily and without duplication. It also allowed standardization of the job loop, which currently was duplicated in each looping job.

This PR is huge but it's mostly due to removing entire files and restructuring them elsewhere. I was able to cut out a lot of duplicated code. `CommandHelpers` was a mess and the use of the new `IArgumentsDictionary` allowed many of its methods to be removed entirely.

WIP:
- [x] link to a release version of NuGet.Services.KeyVault
- [x] fix use of `CancellationToken`
- [x] run the full job pipeline and verify everything works
- [x] verify no new code is overwritten
- [x] update to newer version of extensions


src/Ng/Arguments.cs
```diff
@@ -1,64 +1,102 @@
 ﻿namespace Ng
 {
-    static class Constants
+    static class Arguments
     {
+
+        #region Shared
```
scottbommarito: I sorted everything by what job they were in, using regions, which is much prettier than the alphanumerically sorted blob of arguments before. If you think of a better way to do this let me know.

src/Ng/packages.config
```diff
@@ -27,7 +27,7 @@
   <package id="Newtonsoft.Json" version="8.0.3" targetFramework="net452" />
   <package id="NuGet.Common" version="3.5.0-beta2-1484" targetFramework="net452" />
   <package id="NuGet.Core" version="2.8.5" targetFramework="net452" />
-  <package id="NuGet.Services.KeyVault" version="1.0.0.0" targetFramework="net452" />
+  <package id="NuGet.Services.KeyVault" version="1.0.14-r-scottbom-keyrot" targetFramework="net452" />
```
scottbommarito: (placeholder package until the package is finalized)

src/NuGet.Services.BasicSearch/packages.config
```diff
@@ -38,7 +38,7 @@
   <package id="Microsoft.Owin.StaticFiles" version="3.0.1" targetFramework="net452" />
   <package id="Microsoft.WindowsAzure.ConfigurationManager" version="3.1.0" targetFramework="net452" />
   <package id="Newtonsoft.Json" version="8.0.3" targetFramework="net452" />
-  <package id="NuGet.Services.KeyVault" version="1.0.0.0" targetFramework="net452" />
+  <package id="NuGet.Services.KeyVault" version="1.0.14-r-scottbom-keyrot" targetFramework="net452" />
```
scottbommarito: (placeholder package until the package is finalized)

src/Ng/App.config
```diff
@@ -10,7 +10,7 @@
     <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
       <dependentAssembly>
         <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
-        <bindingRedirect oldVersion="0.0.0.0-8.0.0.0" newVersion="8.0.0.0" />
+        <bindingRedirect oldVersion="0.0.0.0-9.0.0.0" newVersion="9.0.0.0" />
```
ryuyu: This isn't strictly related to this PR, but do we want to move this package (newtonsoft.json) to the 9.x release for all of our packages as we update them? I think it would be best to be consistent one way or the other, and I would prefer to lean towards updating if it's not too much of as hassle (which you've already done here).

scottbommarito: I don't think it's a very high priority, but if we're making changes in a project we might as well do so. The main reason I did this was because the compiler was telling me I needed to do this to fix a dependency--I think `NuGet.Services.Configuration` package required 9.x.

src/Ng/Program.cs
```diff
@@ -57,74 +56,36 @@ public static void Main(string[] args)
...
-                        printToolUsage = Lightning.PrintUsage;
-                        Lightning.Run(arguments, cancellationTokenSource.Token);
-                        break;
-                    default:
-                        printToolUsage();
-                        break;
+                    job = new NgJob(loggerFactory);
+
+                    // Missing or invalid job name, so throw an ArgumentNullException!
+                    throw new ArgumentNullException();
```
skofman1: why?

src/Ng/Jobs/NgJob.cs
```diff
@@ -0,0 +1,74 @@
...
+        protected virtual void Init(IDictionary<string, string> arguments, CancellationToken cancellationToken)
+        {
+            // no-op
+        }
+
+        protected virtual async Task RunInternal(CancellationToken cancellationToken)
+        {
+            // no-op
+            await Task.FromResult<string>(null);
+        }
```
skofman1: abstract

src/Ng/Jobs/LoopingNgJob.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+        }
+
+        public override async Task Run(IDictionary<string, string> arguments, CancellationToken cancellationToken)
+        {
+            var intervalSec = arguments.GetOrNull<int>(Arguments.Interval) ?? Arguments.DefaultInterval;
+
+            Logger.LogInformation("Looping job at interval {Interval} seconds", intervalSec);
+
+            Init(arguments, cancellationToken);
```
skofman1: Initializing very frequently will be a huge overhead for a job that runs continuosly

src/Ng/Jobs/LoopingNgJob.cs
```diff
@@ -0,0 +1,35 @@
...
+            while (true)
```
skofman1: >hile (true) [](start = 13, length = 11)

do.. while will eliminate the need to Init before the loop

src/Ng/Jobs/Catalog2DnxJob.cs
```diff
@@ -0,0 +1,78 @@
...
+        private ReadCursor _back;
+
+        public Catalog2DnxJob(ILoggerFactory loggerFactory)
+            : base(loggerFactory)
+        {
+        }
+
+        public override void PrintUsage()
+        {
+            Console.WriteLine("Usage: ng catalog2dnx "
```
skofman1: >Console.WriteLine [](start = 12, length = 17)

if the logger was initialized, we should log there (please verify the logger was setup to write to Console).

src/Ng/Jobs/CatalogUtility.cs
```diff
@@ -0,0 +1,250 @@
...
+                if (obj.TryGetValue(propertyName, out token))
+                {
+                    return token.ToObject<DateTime>().ToUniversalTime();
+                }
+            }
+
+            return null;
+        }
+
+        public static async Task<SortedList<DateTime, IList<PackageDetails>>> GetPackages(HttpClient client, Uri uri, string keyDateProperty)
```
skofman1: >GetPackages [](start = 78, length = 11)

this is used only by Feed2Catalog. Why move to a shared class?

scottbommarito: It is also used by `Package2Catalog`.

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/3310

src/Ng/CommandHelpers.cs
```diff
@@ -12,381 +13,170 @@
...
+                var shouldValidateCert = arguments.GetOrThrow<bool>(Arguments.ValidateCertificate);
+
+                var keyVaultConfig = new KeyVaultConfiguration(vaultName, clientId, certificateThumbprint, storeName, storeLocation, shouldValidateCert);
+
+                secretReader = new CachingSecretReader(new KeyVaultReader(keyVaultConfig),
+                    arguments.GetOrNull<int>(Arguments.RefreshIntervalSec) ??
+                    CachingSecretReader.DefaultRefreshIntervalSec);
             }
-            else
+            catch (Exception)
```
skofman1: > (Exception) [](start = 17, length = 12)

this hides all exceptions from  the user. If there was an initialization error, I want to know about it. Please leave the previous logic as is.

src/Ng/CommandHelpers.cs
```diff
@@ -12,381 +13,170 @@
...
-            if (!int.TryParse(intervalStr, out interval))
-            {
-                interval = defaultInterval;
-            }
-
-            return interval;
-        }
-
-        public static DateTime GetStartDate(IDictionary<string, string> arguments)
+        public static StorageFactory CreateStorageFactory(IDictionary<string, string> arguments, bool verbose)
```
skofman1: >CreateStorageFactory [](start = 37, length = 20)

I hope all the methods bellow dont have logic changes. This is not covered by UTs, and it's going to be hard to find how slightly different behavior affects things in runtime. My recommendation: undo this.

src/Ng/CommandHelpers.cs
```diff
@@ -410,64 +200,58 @@ public static Lucene.Net.Store.Directory GetCopyDestLuceneDirectory(IDictionary<
...
         }
 
-        public static Lucene.Net.Store.Directory GetLuceneDirectoryImpl(IDictionary<string, string> arguments, IDictionary<string, string> names, bool required = true)
+        public static Lucene.Net.Store.Directory GetLuceneDirectoryImpl(IDictionary<string, string> arguments, IDictionary<string, string> argumentNameMap, bool required = true)
         {
             try
             {
-                string luceneDirectoryType;
-                TryGetArgument(arguments, names[Constants.DirectoryType], out luceneDirectoryType, required: true);
+                var luceneDirectoryType = arguments.GetOrNull(argumentNameMap[Arguments.DirectoryType]);
```
skofman1: >GetOrNull [](start = 52, length = 9)

you changed the behavior of this method. Previously, it would throw in the DirectoryType wasn't specified

src/Ng/Program.cs
```diff
@@ -10,39 +10,38 @@
...
 
             try
             {
                 // Get arguments
-                var arguments = CommandHelpers.GetArguments(args, 0);
+                var arguments = CommandHelpers.GetArguments(args, 1);
 
                 // Configure ApplicationInsights
-                ApplicationInsights.Initialize(CommandHelpers.GetApplicationInsightsInstrumentationKey(arguments));
+                ApplicationInsights.Initialize(arguments.GetOrNull(Arguments.InstrumentationKey));
```
skofman1: >GetOrNull [](start = 57, length = 9)

PR for this?

src/Ng/Jobs/LoopingNgJob.cs
```diff
@@ -0,0 +1,47 @@
...
+    public abstract class LoopingNgJob : NgJob
+    {
+        protected LoopingNgJob(ILoggerFactory loggerFactory)
+            : base(loggerFactory)
+        {
+        }
+
+        public override async Task Run(IDictionary<string, string> arguments, CancellationToken cancellationToken)
+        {
+            var intervalSec = arguments.GetOrNull<int>(Arguments.Interval) ?? Arguments.DefaultInterval;
```
skofman1: >GetOrNull [](start = 40, length = 9)

GetOrDefault

src/Ng/Jobs/NgJob.cs
```diff
@@ -0,0 +1,59 @@
...
+        protected ILoggerFactory LoggerFactory;
+        protected ILogger Logger;
+
+        protected NgJob(ILoggerFactory loggerFactory)
+        {
+            LoggerFactory = loggerFactory;
+
+            // We want to make a logger using the subclass of this job.
+            // Use reflection to find the type that is being constructed and create a logger with its type.
+            var derivedType = GetType();
```
skofman1: >GetType [](start = 30, length = 7)

check that it returns the correct type

src/Ng/Jobs/NgJob.cs
```diff
@@ -0,0 +1,59 @@
...
+            derivedType.GetMethod(nameof(InitializeLogger))
+                .MakeGenericMethod(derivedType)
```
skofman1: do u really want to keep this? :)

src/Ng/Jobs/NgJob.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            var message = "Usage: ng [" + string.Join("|", NgJobFactory.JobMap.Keys) + "] "
+                          + $"[-{Arguments.VaultName} <keyvault-name> "
+                          + $"-{Arguments.ClientId} <keyvault-client-id> "
+                          + $"-{Arguments.CertificateThumbprint} <keyvault-certificate-thumbprint> "
+                          + $"[-{Arguments.ValidateCertificate} true|false]]";
+
+            logger.LogError(message);
+        }
+
```
skofman1: log to console

src/Ng/Jobs/NgJob.cs
```diff
@@ -0,0 +1,59 @@
...
+                .MakeGenericMethod(derivedType)
+                .Invoke(this, new object[] { loggerFactory });
+        }
+        
+        public void InitializeLogger<T>(ILoggerFactory loggerFactory)
+        {
+            Logger = loggerFactory.CreateLogger<T>();
+        }
+
+        public static void PrintUsage(ILogger logger)
```
skofman1: >PrintUsage [](start = 27, length = 10)

return a string and the consumer will decide on how to print it

src/Ng/Jobs/NgJob.cs
scottbommarito: discovered that `ILoggerFactory` has a `CreateLogger(Type t)` so i will use it instead of `CreateLogger<T>()` which gets rid of the reflection!

src/Ng/CommandHelpers.cs
scottbommarito: this actually still does (the if statement below, with `luceneDirectoryType.Equals`, will crash if the directory type is null/undefined) but i will browse over the file and verify that every `required` is handled properly.

src/Ng/CommandHelpers.cs
scottbommarito: Double-checked this file. Everything looks good.

skofman1:
:shipit:

scottbommarito:
@ryuyu @shishirx34

PR #91, opened 2016-09-06
Secret rotation for SearchService

scottbommarito:
Configuration isn't widely used in the SearchService, and is mostly consumed in `NuGetSearcherManager`. I restructured the class along with a couple other places so that it no longer caches a `StorageAccount` or anything else that could lead to stale credentials in the future.
- Removed `ConfigurationService` and related code.
- Integrated with `IConfigurationFactory` and `IConfigurationProvider`. Also moved all configuration keys and default values to `IndexingSettings` and `BasicSearchSettings` as opposed to having them hard-coded for every use.
- Restructured `StorageLoader` to store a `StorageAccount`.
- Created `IIndexDirectoryProvider` with `GetDirectory`, `GetIndexContainerName`, `GetSynchronizer`, and `Reload` methods. This interface is designed so that `NuGetSearcherManager` no longer has to manage any of the information--it is all managed through an `IIndexDirectoryProvider`.
- Added `Reload` to `ILoader`. For `StorageLoader`, this call reloads the `StorageAccount` from configuration. For the other loaders, it no-ops.
- Removed the `AzureDirectorySynchronizer`, `IndexName`, and `Directory` references from `NuGetSearcherManager` and replaced them with accesses to an `IIndexDirectoryProvider`.
- Added a `Search.IndexReloadRateSec` configuration value. The `NuGetSearcherManager` will not attempt to reload the index or its storage account unless this much time has passed.
- Added a `Search.AuxiliaryDataRefreshRateSec` for consistency with `Search.IndexReloadRateSec`--previously, the value was hard-coded.
- Added `ReloadIndexAndLoaderIfExpired` to `NuGetSearcherManager`. This method checks if the amount of time specified by `Search.IndexReloadRateSec` has passed. If so, it calls `Reload` on both the `ILoader` and `IIndexDirectoryProvider`.
- Added a call to `ReloadIndexAndLoaderIfExpired` in `NuGetSearcherManager.Reopen`. Therefore, every time the `IndexReader` is reopened, the index is reloaded beforehand if it has expired.

- [x] Test more thoroughly
- [ ] Update references to a release version of `NuGet.Services.KeyVault`


scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/3311

src/NuGet.Indexing/ILoader.cs
```diff
@@ -2,11 +2,16 @@
...
 using Newtonsoft.Json;
+using System.Threading.Tasks;
 
 namespace NuGet.Indexing
 {
     public interface ILoader
     {
         JsonReader GetReader(string name);
+
+        AzureDirectorySynchronizer GetSynchronizer();
```
ryuyu: I don't think this method should live on this interface.

scottbommarito: I think the reason I put it here was because I thought it somewhat made sense to have it live with the [`StorageLoader`](https://github.com/NuGet/NuGet.Services.Metadata/pull/91/files#diff-abd3dac6b87987c81b2d39954087afdaR20). Would you prefer it to be in its own class?

src/NuGet.Indexing/StorageLoader.cs
```diff
@@ -7,21 +7,34 @@
...
-        private readonly CloudStorageAccount _storageAccount;
         private readonly string _containerName;
         private readonly FrameworkLogger _logger;
 
-        public StorageLoader(CloudStorageAccount storageAccount, string containerName, FrameworkLogger logger)
+        private CloudStorageAccount _storageAccount;
+        private AzureDirectorySynchronizer _synchronizer;
+        private Func<Task<Tuple<CloudStorageAccount, AzureDirectorySynchronizer>>> _createCloud;
+
+        public static async Task<StorageLoader> CreateStorageLoader(Func<Task<Tuple<CloudStorageAccount, AzureDirectorySynchronizer>>> createCloud, string containerName, FrameworkLogger logger)
```
ryuyu: I'm not sure I like passing in a function here. I feel like StorageLoader itself should have the implementation to generate what it needs, and we should be passing in parameters that it needs (or the configuration). Is there a particular reason this is implemented in this manner?

scottbommarito: I think passing in parameters sounds like a much better implementation. I will fix.

src/NuGet.Indexing/NuGetSearcherManager.cs
```diff
@@ -8,74 +8,80 @@
...
+            _indexProvider = indexProvider;
             _loader = loader;
-            _azureDirectorySynchronizer = azureDirectorySynchronizer;
+
+            _indexReloadRate = TimeSpan.FromSeconds(indexReloadRateSec);
+            _auxiliaryDataRefreshRate = TimeSpan.FromSeconds(auxiliaryDataRefreshRateSec);
         }
 
-        public string IndexName { get; private set; }
+        public virtual string IndexName => _indexProvider.GetIndexContainerName();
```
scottbommarito: (I made this virtual so that it can be mocked in a test case.)

src/Ng/packages.config
```diff
@@ -24,10 +24,10 @@
   <package id="Microsoft.Net.Http" version="2.2.22" targetFramework="net452" />
   <package id="Microsoft.Web.Xdt" version="2.1.1" targetFramework="net452" />
   <package id="Microsoft.WindowsAzure.ConfigurationManager" version="3.1.0" targetFramework="net452" />
-  <package id="Newtonsoft.Json" version="8.0.3" targetFramework="net452" />
+  <package id="Newtonsoft.Json" version="9.0.1" targetFramework="net452" />
   <package id="NuGet.Common" version="3.5.0-beta2-1484" targetFramework="net452" />
   <package id="NuGet.Core" version="2.8.5" targetFramework="net452" />
-  <package id="NuGet.Services.KeyVault" version="1.0.0.0" targetFramework="net452" />
+  <package id="NuGet.Services.KeyVault" version="1.0.14-r-scottbom-keyrot" targetFramework="net452" />
```
ryuyu: Do we really want to depend on this prerelease package right now? Are we getting close to doing an updated release of this?

scottbommarito: This is the wrong package. When I fix the merge conflicts, this will go to `1.0.75+`. We already depend on pre-release packages in many places (they are just `-r-develop` branches).

src/NuGet.Services.BasicSearch/SecretReader/SecretReaderFactory.cs
```diff
@@ -2,25 +2,26 @@
...
 {
-    internal class SecretReaderFactory : ISecretReaderFactory
+    public class SecretReaderFactory : ISecretReaderFactory
     {
-        public const string VaultNameKey = "keyVault:VaultName";
-        public const string ClientIdKey = "keyVault:ClientId";
-        public const string CertificateThumbprintKey = "keyVault:CertificateThumbprint";
-
-        public ISecretReader CreateSecretReader(IConfiguration configuration)
+        public ISecretReader CreateSecretReader()
```
joelverhagen: >CreateSecretReader [](start = 29, length = 18)

Write an async helper method to this method can have a single .Result.

src/NuGet.Indexing/IndexDirectoryProvider.cs
```diff
@@ -0,0 +1,75 @@
...
+            var newIndexContainerName = await _settings.GetOrDefault(IndexingSettings.IndexContainer, IndexingSettings.IndexContainerDefault);
+            if (_directory != null && newIndexContainerName == _indexContainerName)
+            {
+                return false;
+            }
+
+            _indexContainerName = newIndexContainerName;
+            _logger.LogInformation("StorageLoader index container: {IndexContainerName}", _indexContainerName);
+            
+            var storageAccount = CloudStorageAccount.Parse(await _settings.GetOrThrow<string>(IndexingSettings.StoragePrimary));
```
ryuyu: Does the setting for IndexingSettings.StoragePrimary contain a full connection string? If it's just the storage key, CloudStorageAccount.Parse I don't think can create a valid connection. Can anybody confirm?

src/NuGet.Indexing/SearcherManagerT.cs
```diff
@@ -16,22 +17,17 @@ public abstract class SearcherManager<TIndexSearcher>
...
-        public Directory Directory { get; private set; }
+        protected abstract Directory GetDirectory();
 
-        public SearcherManager(Directory directory)
-        {
-            Directory = directory;
-        }
-
-        public void Open()
+        public async Task Open()
```
joelverhagen: >Open [](start = 26, length = 4)

OpenAsync?

src/NuGet.Indexing/IndexDirectoryProvider.cs
scottbommarito: It is a full connection string that has the key encoded in `$$`s. (You can double-check this by looking at the variables in Octopus.)

src/NuGet.Indexing/StorageLoader.cs
```diff
@@ -42,5 +52,23 @@ public JsonReader GetReader(string name)
                 throw;
             }
         }
+
+        public async Task<bool> Reload()
```
ryuyu: Cool. Looks like we can leverage this to dynamically update the container name from another source.

src/NuGet.Services.BasicSearch/SecretReader/SecretReaderFactory.cs
```diff
@@ -2,25 +2,26 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using NuGet.Indexing;
+using System.Security.Cryptography.X509Certificates;
 using NuGet.Services.KeyVault;
 
-namespace NuGet.Services.BasicSearch
+namespace NuGet.Services.BasicSearch.SecretReader
```
ryuyu: Does this really need to be in its own namespace?

scottbommarito: It shares the namespace with `EnvironmentSettingsProvider`. I have renamed the namespace to `Configuration` because I think it makes more sense.

src/NuGet.Services.BasicSearch/SecretReader/EmptySecretReaderFactory.cs
```diff
@@ -3,17 +3,18 @@
 
 using NuGet.Indexing;
 using NuGet.Services.KeyVault;
+using System.Threading.Tasks;
 
 namespace NuGet.Services.BasicSearch
 {
-    internal class EmptySecretReaderFactory : ISecretReaderFactory
+    public class EmptySecretReaderFactory : ISecretReaderFactory
```
skofman1: why the change?

src/NuGet.Services.BasicSearch/SecretReader/SecretReaderFactory.cs
```diff
@@ -2,25 +2,26 @@
...
 {
-    internal class SecretReaderFactory : ISecretReaderFactory
+    public class SecretReaderFactory : ISecretReaderFactory
```
skofman1: why not internal?

src/NuGet.Indexing/NuGetSearcherManager.cs
```diff
@@ -143,13 +138,35 @@ public class NuGetSearcherManager : SearcherManager<NuGetIndexSearcher>
...
+        }
+        
+        private async Task<bool> ReloadIndexAndLoaderIfExpired()
         {
-            if (_azureDirectorySynchronizer != null)
+            var hasReloaded = false;
+
+            if (_lastTimeIndexReloaded < DateTime.UtcNow - _indexReloadRate)
+            {
+                hasReloaded = await _indexProvider.Reload() || await _loader.Reload();
```
scottbommarito: LOL

src/NuGet.Indexing/NuGetSearcherManager.cs
```diff
@@ -84,57 +90,46 @@ public class NuGetSearcherManager : SearcherManager<NuGetIndexSearcher>
...
+            }
+
+            IIndexDirectoryProvider indexProvider;
+            if (directory == null)
+            {
+                // If no directory has been provided, create an IndexDirectoryProvider from the configuration.
+                indexProvider = await IndexDirectoryProvider.Create(settings, logger);
             }
             else
             {
```
scottbommarito: Unfortunately, with the addition of the `IIndexDirectoryProvider` interface, splitting the scenarios would now have far more redundant code than before. I think I'll keep it as is but I'll take a look to see if I can make it easier to read.

src/NuGet.Indexing/NuGetSearcherManager.cs
scottbommarito: [Not really.](https://github.com/NuGet/NuGet.Services.Metadata/pull/91/files#diff-76f506fbfd233fe530713bf9da2104fdR36)

I thought this was the least code impact of the possible options since you cannot pass in a `IndexDirectoryProvider` into `Create`.

scottbommarito:
Merge conflicts now fixed!

src/NuGet.Services.BasicSearch/Startup.cs
```diff
@@ -172,19 +173,22 @@ private void ReopenCallback(object state)
...
-                    () =>
+                await Retry.IncrementalAsync(
+                    async () =>
                     {
                         var stopwatch = Stopwatch.StartNew();
 
-                        _searcherManager = NuGetSearcherManager.Create(configuration, loggerFactory, directory, loader);
-                        _searcherManager.Open();
+                        var searcherManager = await NuGetSearcherManager.Create(settings, loggerFactory, directory, loader);
+                        await _searcherManager.OpenAsync();
```
scottbommarito: yes, thanks

src/NuGet.Services.BasicSearch/Startup.cs
```diff
@@ -172,19 +173,22 @@ private void ReopenCallback(object state)
...
+
+                        // Guarantee that the searcher manager has been opened before we set it.
```
skofman1: >/ Guarantee that t [](start = 25, length = 18)

why is this important?

src/NuGet.Services.BasicSearch/Web.config
```diff
@@ -1,116 +1,119 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <!--
   For more information on how to configure your ASP.NET application, please visit
   http://go.microsoft.com/fwlink/?LinkId=169433
   -->
 <configuration>
   <system.web>
-    <compilation debug="true" targetFramework="4.5.2" />
-    <httpRuntime targetFramework="4.5.2" />
```
skofman1: >5 [](start = 36, length = 1)

have you changed something non cosmetic here?

scottbommarito: As far as I know the only thing I added was `keyVault:ValidateCertificate`.

src/NuGet.Services.BasicSearch/Startup.cs
scottbommarito: It's probably not, but I did it because previously the two lines above ran synchronously, but now this is an asynchronous context. I want to make the operations above atomic to reduce the likelihood of a threading problem, so I moved the assignment of `_searcherManager`. Not sure if it's necessary.

src/NuGet.Indexing/IndexDirectoryProvider/CloudIndexDirectoryProvider.cs
```diff
@@ -0,0 +1,92 @@
...
+using Lucene.Net.Store.Azure;
+using Microsoft.Extensions.Logging;
+using Microsoft.WindowsAzure.Storage;
+using NuGet.Services.Configuration;
+using FrameworkLogger = Microsoft.Extensions.Logging.ILogger;
+
+namespace NuGet.Indexing.IndexDirectoryProvider
+{
+    /// <summary>
+    /// Maintains an index on the cloud. Provides a
```
scottbommarito: LOL will fix

skofman1:
>         public static void Incremental(Action runLogic, Func<Exception, bool> shouldRetry, int maxRetries, TimeSpan waitIncrement)

is this one still used? if not, you can delete it

---
Refers to: src/NuGet.Indexing/Retry.cs:12 in 18c1155. [](commit_id = 18c11551f68ecea6651ebe8ddcfac0b3e06bd06d, deletion_comment = False)

scottbommarito:
@skofman1 Both `Retry.Incremental` are used.

src/NuGet.Services.BasicSearch/SecretReader/SecretReaderFactory.cs
skofman1: Read about this: https://msdn.microsoft.com/en-us/library/system.runtime.compilerservices.internalsvisibletoattribute(v=vs.110).aspx

This is the correct approach to make stuff visible to test code

src/NuGet.Indexing/IndexDirectoryProvider/CloudIndexDirectoryProvider.cs
```diff
@@ -0,0 +1,83 @@
...
+        private readonly FrameworkLogger _logger;
+
+        private Directory _directory;
+        private string _indexContainerName;
+        private string _storageAccountConnectionString;
+        private AzureDirectorySynchronizer _synchronizer;
+
+        public CloudIndexDirectoryProvider(IndexingConfiguration config, FrameworkLogger logger)
+        {
+            _logger = logger;
```
skofman1: null checks

scottbommarito:
I have closed this PR because https://github.com/NuGet/NuGet.Services.Metadata/pull/111 extends it.

PR #92, opened 2016-09-15
Adding Basic Unit Tests for ResponseFormatter

ryuyu:
This PR adds a framework and some basic tests for validating the ResponseFormatter.

@maartenba @xavierdecoster @skofman1 @shishirx34 @scottbommarito


tests/NuGet.IndexingTests/TestSupport/MockIndexReader.cs
```diff
@@ -0,0 +1,136 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Lucene.Net.Documents;
+using Lucene.Net.Index;
+
+namespace NuGet.IndexingTests.TestSupport
+{
+    public class MockIndexReader : IndexReader
```
scottbommarito: could we use `Mock<IndexReader>` for this instead?

```
public static Mock<IndexReader> CreateMockIndexReader(int numberOfDocs)
{
    var mockIndexReader = new Mock<IndexReader>(MockBehavior.Strict);
    var numDocs = numberOfDocs;
    
    mockIndexReader.Setup(x => x.MaxDoc).Returns(numDocs);
    mockIndexReader.Setup(x => x.TermDocs()).ReturnsNull();
    mockIndexReader.Setup(x => x.NumDocs()).Returns(numDocs);
}
```

tests/NuGet.IndexingTests/TestSupport/MockLogger.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace NuGet.IndexingTests.TestSupport
+{
+    public class MockLogger : ILogger
```
scottbommarito: `new Mock<Logger>()`?

tests/NuGet.IndexingTests/TestSupport/MockTopDocs.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Lucene.Net.Search;
+
+namespace NuGet.IndexingTests.TestSupport
+{
+    public class MockTopDocs : TopDocs
+    {
+        public MockTopDocs(int totalHits, ScoreDoc[] scoreDocs, float maxScore) :
+            base(totalHits, scoreDocs, maxScore)
```
scottbommarito: Do you need this class if all it does is pass the values to its superclass?

tests/NuGet.IndexingTests/TestSupport/MockSearcherManager.cs
```diff
@@ -0,0 +1,19 @@
...
+using System;
+using NuGet.Indexing;
+
+namespace NuGet.IndexingTests.TestSupport
+{
+    public class MockSearcherManager : NuGetSearcherManager
+    {
+        private string _indexName;
+
+        public MockSearcherManager(string indexName) : base(indexName, new MockLogger(), null, null, null)
```
scottbommarito: you could maybe just put a new constructor in `NuGetSearchManager` and call that in the unit test

ryuyu: I would prefer not to touch the existing code without a production need, especially just to enable test scenarios.

However, I'm willing to change it if the majority opinion is different.

tests/NuGet.IndexingTests/TestSupport/MockTopDocs.cs
ryuyu: Good question. This was originally done with the expectation that things would have to be overridden, but that doesn't seem to be the case and it can probably safely be removed.

That being said, I do like the clear denomination between the "mock" objects and the real objects, and this makes it easier to override if we need to later.

Either way, I do not feel particularly strongly about it, and will change it if we think it is the right thing to do.

tests/NuGet.IndexingTests/TestSupport/MockIndexReader.cs
ryuyu: done. Thanks for the suggestion.

tests/NuGet.IndexingTests/TestSupport/MockLogger.cs
ryuyu: Done.

tests/NuGet.IndexingTests/TestSupport/MockTopDocs.cs
scottbommarito: I think it would be trivial to switch to a "Mock" class when the need arises, so for now we can just get rid of it.

tests/NuGet.IndexingTests/TestSupport/MockSearcherManager.cs
scottbommarito: I think given that it's just a small constructor that would be calling another constructor I don't see an issue with touching the code. I've had to add constructors for testing before--sometimes there's no other way to do it due to how the language works.

tests/NuGet.IndexingTests/TestSupport/MockTopDocs.cs
ryuyu: After thinking about it, this sufficiently works around having to add a new constructor, and allows me to make some setup for testing purposes as well, which I am actually against adding in non-test code. I will leave this as is.

tests/NuGet.IndexingTests/ResponseFormatterTests.cs
```diff
@@ -0,0 +1,484 @@
...
+namespace NuGet.IndexingTests
+{
+    public class ResponseFormatterTests
+    {
+
+        [Theory]
+        [MemberData(nameof(V2CountResultData))]
+        public void WriteV2CountResultTest(int totalHits, string expected)
+        {
+            StringBuilder sb = new StringBuilder();
```
maartenba: Var all the things

tests/NuGet.IndexingTests/ResponseFormatterTests.cs
```diff
@@ -0,0 +1,484 @@
...
+                    Constants.LucenePropertyId.ToLower()), sb.ToString());
+            }
+        }
+
+        [Fact]
+        public void WriteRankingsResultTest()
+        {
+            var rankingsSegment = new RankingBySegment();
+
+            rankingsSegment[Constants.RankingsSegmentName] = generateRankings(numRankings: 5, idPrefix: Constants.RankingsIdPrefix);
```
maartenba: Uppercase G + add some newlines)

tests/NuGet.IndexingTests/TestSupport/MockSearcherManager.cs
maartenba: I like the overload for testing

tests/NuGet.IndexingTests/ResponseFormatterTests.cs
```diff
@@ -0,0 +1,484 @@
...
+                    100,  // Num docs in index
+                    new Dictionary<string, string> {
+                        { "user1", "value1" },
+                        { "user2", "value2" },
+                        { "commitTimeStamp", "1/1/2000" }
+                    },
+                    1,   // Top Docs Total Hits
+                    1.0,  // Top Docs Max Score
+                    0,    // skip
+                    0,    // take
```
skofman1: add more values or total hits, score, skip, take

tests/NuGet.IndexingTests/ResponseFormatterTests.cs
```diff
@@ -0,0 +1,484 @@
...
+            int take,
+            bool includeExplanation,
+            string expected)
+        {
+            var searcher = new MockSearcher(indexName, numDocs, commitUserData);
+            var topDocs = new TopDocs(topDocsTotalHits, Constants.ScoreDocs, topDocsMaxScore);
+
+            var sb = new StringBuilder();
+            var sw = new StringWriter(sb);
+            using (var writer = new JsonTextWriter(sw))
```
skofman1: >us [](start = 12, length = 2)

newline

tests/NuGet.IndexingTests/TestSupport/Constants.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Lucene.Net.Search;
+using NuGet.Indexing;
+
+namespace NuGet.IndexingTests.TestSupport
```
skofman1: >TestSupport [](start = 30, length = 11)

nit: No need to have many namespaces, they just add unnecessary usings. Can we leave this in NuGet.IndexingTests?

tests/NuGet.IndexingTests/TestSupport/MockObjectFactory.cs
```diff
@@ -0,0 +1,51 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Lucene.Net.Documents;
+using Lucene.Net.Index;
+using Microsoft.Extensions.Logging;
+using Moq;
+
+namespace NuGet.IndexingTests.TestSupport
+{
+    public class MockObjectFactory
```
skofman1: >MockObjectFactory [](start = 17, length = 17)

you can make this static

tests/NuGet.IndexingTests/TestSupport/MockSearcherManager.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        public MockSearcherManager(string indexName) 
+            : base(indexName: indexName, 
+                  logger: MockObjectFactory.CreateMockLogger().Object, 
+                  directory: null, 
+                  loader: null, 
+                  azureDirectorySynchronizer: null)
+        {
+            _indexName = indexName;
+            RegistrationBaseAddress[Constants.SchemeName] = new Uri(Constants.BaseUri);
```
skofman1: why do we need this class? it contains no logic.. Can we write the UTs to use the original class, or to use a Mock?

tests/NuGet.IndexingTests/ResponseFormatterTests.cs
ryuyu: I thought I got all of these. Looks like I missed one. Done.

tests/NuGet.IndexingTests/ResponseFormatterTests.cs
ryuyu: done

tests/NuGet.IndexingTests/ResponseFormatterTests.cs
ryuyu: done

tests/NuGet.IndexingTests/TestSupport/MockObjectFactory.cs
ryuyu: Done

tests/NuGet.IndexingTests/TestSupport/Constants.cs
ryuyu: I don't mind moving these into the NuGet.IndexingTests namespace, but the other files in the TestSupport folder are already in the TestSupport namespace. Do we just want to get rid of that entirely and put everything in the INdexingTests namespace?

tests/NuGet.IndexingTests/TestSupport/MockSearcherManager.cs
ryuyu: I want the constructor for this object to hide the fact that we need more stuff since we don't care about it in the test. This constructor also sets the RegistrationBaseAddress at construction time, which I also don't want the tests to explicitly set (just know what it is supposed to be given a scheme). I think @scottbommarito commented on this as well. I think there should be a discussion above.

ryuyu: Removed this and changed to directly us it in the Searcher since it isn't used anywhere else.

tests/NuGet.IndexingTests/TestSupport/MockSearcher.cs
```diff
@@ -0,0 +1,54 @@
...
+                  rankings: null,
+                  context: null,
+                  latestBitSet: Constants.LatestBitSet,
+                  latestStableBitSet: Constants.LatestStableBitSet,
+                  owners: Constants.EmptyOwnersResult)
+        {
+            MockObjectFactory.MockPrefix = Constants.MockBase;
+        }
+
+        private static NuGetSearcherManager InitNuGetSearcherManagerMock(string indexName)
```
skofman1: >NuGetSearcherManager [](start = 23, length = 20)

like :)

skofman1: >Mock [](start = 68, length = 4)

it's not a Mock, please fix name

skofman1:
:shipit:

PR #93, opened 2016-09-27
Fix CreateLogger failing in sql2lucene

ryuyu:
For some reason, the call to CreateLogger was throwing a MissingMethod exception when we went cross assembly. This PR adds reference to Microsoft.Extensions.Logging in NuGet.Indexing to remedy this.

Note that I'm not 100% sure this is the right thing to do here, since as far as i can tell, we shouldn't need an explicit reference.

@skofman1 @xavierdecoster @shishirx34 @scottbommarito @maartenba


src/NuGet.Indexing/Sql2Lucene.cs
```diff
@@ -225,12 +225,12 @@ public static void Export(string sourceConnectionString, string destinationPath,
...
 
             var batches = CalculateBatches(sourceConnectionString);
-            logger.LogVerbose("Calculated {BatchCount} batches (took {BatchCalculationTime} seconds)", batches.Count, stopwatch.Elapsed.TotalSeconds);
+            LoggerExtensions.LogInformation(logger, "Calculated {BatchCount} batches (took {BatchCalculationTime} seconds)", batches.Count, stopwatch.Elapsed.TotalSeconds);
 
             stopwatch.Restart();
 
             var packageFrameworks = LoadPackageFrameworks(sourceConnectionString);
-            logger.LogVerbose("Loaded package frameworks (took {PackageFrameworksLoadTime} seconds)", stopwatch.Elapsed.TotalSeconds);
+            LoggerExtensions.LogInformation(logger, "Loaded package frameworks (took {PackageFrameworksLoadTime} seconds)", stopwatch.Elapsed.TotalSeconds);
```
skofman1: >LogInformation [](start = 29, length = 14)

logger.LogInformation.. it's an extension method..

ryuyu: Done

scottbommarito:
Any reason why we haven't merged this?

@ryuyu

PR #94, opened 2016-09-28
Search service functional tests

shishirx34:
- Created a self contained project for search service functional tests. Tests can be run from Visual studio test explorer too against INT environment. 
- Adds functional test coverage for all endpoints of search service.
- Adds scripts needed to build and run tests from team city. These tests can be run locally against any hosted search service. Takes two urls as parameters, SearchServiceUrl and IndexBaseUrl for all service endpoints for nuget gallery.

I have build configuration in TC running the functional tests listed here. 
/cc @xavierdecoster @maartenba @skofman1 @ryuyu


tests/BasicSearchTests.FunctionalTests.Core/StartupFunctionalTests.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.Net.Http;
+using BasicSearchTests.FunctionalTests.Core.Models;
+using Newtonsoft.Json;
+
+namespace BasicSearchTests.FunctionalTests.Core
+{
+    public class StartupFunctionalTests
+    {
+        private HttpClient _client;
+        private static string RegistrationBaseUrl = "RegistrationsBaseUrl";
```
skofman1: >RegistrationsBaseUrl [](start = 53, length = 20)

can you please check if this const already exists in a package we can use?

tests/BasicSearchTests.FunctionalTests.Core/StartupFunctionalTests.cs
```diff
@@ -0,0 +1,90 @@
...
+            var response = await _client.GetAsync("/");
+            var content = await response.Content.ReadAsStringAsync();
+
+            // Assert
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            Assert.Equal(content, "READY");
+        }
+
+        [Fact]
+        public async Task InvalidEndpoint()
```
skofman1: >InvalidEndpoint [](start = 26, length = 15)

do we need this test?  Why is this functionality important?

tests/BasicSearchTests.FunctionalTests.Core/StartupFunctionalTests.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            Assert.False(registrationBaseUrl == null, "Failed to get registration base url. Please check that index url is correct.");
+
+            var cursorUri = new Uri(new Uri(registrationBaseUrl), "cursor.json");
+            var registrationResponse = await httpClient.GetAsync(cursorUri);
+            registrationResponse.EnsureSuccessStatusCode();
+
+            var registrationCursorContent = await registrationResponse.Content.ReadAsStringAsync();
+            var lastRegistrationCommitTime = JsonConvert.DeserializeObject<RegistrationCursor>(registrationCursorContent).Value;
+            Assert.NotNull(lastRegistrationCommitTime);
```
skofman1: please separate to a different method

tests/BasicSearchTests.FunctionalTests.Core/StartupFunctionalTests.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            // Assert
+            Assert.Equal(HttpStatusCode.NotFound, response.StatusCode);
+            Assert.Equal(content, "UNRECOGNIZED");
+        }
+
+        [Fact]
+        public async Task IndexIsFresh()
+        {
+            var httpClient = new HttpClient();
```
skofman1: >HttpClient [](start = 33, length = 10)

something that concerns me about this code in general: requests are not retried. Since we run this code during deployment, a failure of a request will cause the deployment to fail. NuGet.Protocol package contains a class called HttpRetryHandler. Using it will allow you to easily add retries to your code.

tests/BasicSearchTests.FunctionalTests.Core/StartupFunctionalTests.cs
```diff
@@ -0,0 +1,90 @@
...
+            Assert.NotNull(lastRegistrationCommitTime);
+
+            var response = await _client.GetAsync("/search/diag");
+            var content = await response.Content.ReadAsAsync<SearchDiagResult>();
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            Assert.NotNull(content);
+
+            var diffTimes = lastRegistrationCommitTime.Subtract(content.CommitUserData.CommitTimeStamp).TotalHours;
+            //Last CommitTimeStamp for search service shouldn't be far off from the last registration timestamp.
+            Assert.True(diffTimes >= 0 && diffTimes <= 1, "Search index is ahead of last registration timestamp");
```
skofman1: >1 [](start = 55, length = 1)

const

tests/BasicSearchTests.FunctionalTests.Core/StartupFunctionalTests.cs
shishirx34: Kept it for completion. Not super important, but it kind of validates that incorrect endpoints return NotFound

tests/BasicSearchTests.FunctionalTests.Core/V2SearchFunctionalTests.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Threading.Tasks;
+using BasicSearchTests.FunctionalTests.Core.Models;
+using BasicSearchTests.FunctionalTests.Core.TestSupport;
+using Xunit;
+using System;
+using System.Net;
+
+namespace BasicSearchTests.FunctionalTests.Core
+{
+    public class V2SearchSkipFunctionalTests
```
skofman1: >Skip [](start = 25, length = 4)

skip?

tests/BasicSearchTests.FunctionalTests.Core/StartupFunctionalTests.cs
```diff
@@ -0,0 +1,90 @@
...
+{
+    public class StartupFunctionalTests
+    {
+        private HttpClient _client;
+        private static string RegistrationBaseUrl = "RegistrationsBaseUrl";
+
+        public StartupFunctionalTests()
+        {
+            // Arrange
+            _client = new HttpClient { BaseAddress = new Uri(EnvironmentSettings.SearchServiceBaseUrl) };
```
scottbommarito: It appears that all of your function test classes have this same constructor and line.

I suggest making a base class with a protected/public `HttpClient` that is initialized in its constructor.

tests/BasicSearchTests.FunctionalTests.Core/V2SearchFunctionalTests.cs
shishirx34: whoops.. legacy from copying over existing tests. Will fix this.

tests/BasicSearchTests.FunctionalTests.Core/StartupFunctionalTests.cs
scottbommarito: Does this fail if a job is down? I would be a little careful about introducing a functional test that relies on the jobs being functional.

tests/BasicSearchTests.FunctionalTests.Core/StartupFunctionalTests.cs
shishirx34: Good point. As a side note, gallery's functional tests use the same, HttpClient.

tests/BasicSearchTests.FunctionalTests.Core/AutocompleteFunctionalTests.cs
```diff
@@ -0,0 +1,50 @@
...
+            // Arrange
+            _client = new HttpClient { BaseAddress = new Uri(EnvironmentSettings.SearchServiceBaseUrl) };
+        }
+
+        [Fact]
+        public async Task CanGetEmptyResult()
+        {
+            // Act
+            var response = await _client.GetAsync(new AutocompleteBuilder { Query = Constants.NonExistentSearchString }.RequestUri);
+            var result = await response.Content.ReadAsAsync<AutocompleteResult>();
```
skofman1: >AutocompleteResult [](start = 60, length = 18)

what happens when we add a new property to the json? Will the deserialization succeed?

tests/BasicSearchTests.FunctionalTests.Core/TestSupport/V2SearchBuilder.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace BasicSearchTests.FunctionalTests.Core.TestSupport
+{
+    public class V2SearchBuilder
```
scottbommarito: Because these three builders are very similar, I would make a base class.

```
public class QueryBuilder
{
    protected readonly string Endpoint;

    public string Query { get; set; }
    public bool Prerelease { get; set; }

    protected virtual QueryString GetQueryString()
    {
        var queryString = System.Web.HttpUtility.ParseQueryString(string.Empty);
        queryString["q"] = Query;
        queryString["prerelease"] = Prerelease.ToString();
        return queryString;
    }
    
    public Uri RequestUri
    {
        get
        {
            return new Uri(Endpoint + GetQueryString(), UriKind.Relative);
        }
    }
}
```

Then set `Endpoint` in the subclass constructors and override `GetQueryString()` if necessary

tests/BasicSearchTests.FunctionalTests.Core/EnvironmentSettings.cs
```diff
@@ -0,0 +1,80 @@
...
+    /// This class reads the various test run settings which are set through env variable.
+    /// </summary>
+    public class EnvironmentSettings
+    {
+        private static string _searchServiceBaseurl;
+        private static string _indexBaseUrl;
+
+        /// <summary>
+        /// The environment against which the (search service) test has to be run. The value would be picked from env variable.
+        /// If nothing is specified, int search service is used as default.
```
scottbommarito: Also I'm not sure if we want to have our non-prod environment urls in our repos.

tests/BasicSearchTests.FunctionalTests.Core/AutocompleteFunctionalTests.cs
shishirx34: I verified by removing some properties from `AutocompleteResult` and it does in fact deserialize the properties of the object correctly. Any superfluous properties are ignored.

tests/BasicSearchTests.FunctionalTests.Core/EnvironmentSettings.cs
shishirx34: int is the default. Gallery uses the same.

tests/BasicSearchTests.FunctionalTests.Core/Models/LastCommitMetadata.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using System;
+
+namespace BasicSearchTests.FunctionalTests.Core.Models
+{
+    public class LastCommitMetadata
+    {
+        public DateTime CommitTimeStamp { get; set; }
+
+        public DateTime? Commit_Time_Stamp { get; set; }
```
shishirx34: /diag/search returns the same. Don't know why.

tests/BasicSearchTests.FunctionalTests.Core/AutocompleteFunctionalTests.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Net.Http;
+using System.Threading.Tasks;
+using BasicSearchTests.FunctionalTests.Core.Models;
+using BasicSearchTests.FunctionalTests.Core.TestSupport;
+using Xunit;
+using System.Net;
+
+namespace BasicSearchTests.FunctionalTests.Core
+{
+    public class AutocompleteFunctionalTests
```
shishirx34: I do not see this "convention" being used anywhere, the only functional tests I am aware of are in Gallery and they don't use it.

tests/BasicSearchTests.FunctionalTests.Core/StartupFunctionalTests.cs
shishirx34: This test doesn't rely on a job, it relies on the status of the service at the moment of deployment. That was the idea to add this test case to ensure that the indexes are not out of sync by whole lot which caused the search service to go down the other day.

tests/BasicSearchTests.FunctionalTests.Core/AutocompleteFunctionalTests.cs
scottbommarito: Oh I meant the unit tests. [For example...](https://github.com/NuGet/NuGetGallery/blob/master/tests/NuGetGallery.Facts/AutocompleteServicePackageVersionsQueryFacts.cs#L13)

If the Gallery functional tests don't have it then that's ok, but I think we should probably it consistent across all our XUnit tests.

tests/BasicSearchTests.FunctionalTests.Core/StartupFunctionalTests.cs
shishirx34: I think its an overkill for simple things. If we had more common functionality I'd agree. #wontfix

tests/BasicSearchTests.FunctionalTests.Core/TestSupport/V2SearchBuilder.cs
shishirx34: I think these are simplified enough already, but alright. Fixed.

tests/BasicSearchTests.FunctionalTests.Core/StartupFunctionalTests.cs
shishirx34: I don't think that there are any existing packages that has this defined.

tests/BasicSearchTests.FunctionalTests.Core/StartupFunctionalTests.cs
shishirx34: Nuget.Protocol was dragging too many unnecessary libraries with it, I have added the retry logic to the project itself and verified it works.

tests/BasicSearchTests.FunctionalTests.Core/StartupFunctionalTests.cs
skofman1: [https://github.com/NuGet/NuGet.Client/blob/cea464c27a439406c669643e4a6901b56a67a337/src/NuGet.Core/NuGet.Protocol.Core.v3/Constants.cs](https://github.com/NuGet/NuGet.Client/blob/cea464c27a439406c669643e4a6901b56a67a337/src/NuGet.Core/NuGet.Protocol.Core.v3/Constants.cs)

---
In reply to: [81028197](https://github.com/NuGet/NuGet.Services.Metadata/pull/94#discussion_r81028197) [](ancestors = 81028197)

shishirx34:
@skofman1 - Looks good?

skofman1:
:shipit:

PR #96, opened 2016-10-07
Added basic search analysis

harikmenon:
Added basic search term analysis to the service.


harikmenon:
@yishaigalatzer @joelverhagen @skofman1

PR #97, opened 2016-11-05
Parse autocomplete queries like regaular searches

skofman1:
Changed the way autocomplete queries are parsed, to avoid throwing 500 on illegal Lucene syntax.
Queries that contain special lucene characters would fail: https://lucene.apache.org/core/3_0_3/queryparsersyntax.html

Examples of such queries here: https://github.com/NuGet/NuGetGallery/issues/3330

@xavierdecoster @joelverhagen


maartenba:
Yes!

xavierdecoster:
:shipit: 👏

PR #98, opened 2016-11-09
Feed2Catalog: Use auditing blob last modified timestamp for deletions

skofman1:
The auditing file schema has changed, so we can't rely on it anymore for getting auditing record creation time.


skofman1:
https://github.com/NuGet/NuGetGallery/issues/3341

src/Catalog/Persistence/StorageListItem.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace NuGet.Services.Metadata.Catalog.Persistence
+{
+    public class StorageListItem
```
scottbommarito: I think the code would be tidier if you made constructors for this class instead of scattering them around in various places.

```
public StorageListItem(IListBlobItem item)
{
    Uri = item.Uri;
    LastModifiedUtc = (item as CloudBlockBlob)?.Properties.LastModified?.UtcDateTime;
}

public StorageListItem(FileInfo file)
{
    Uri = GetUri(file.FullName.Replace(Path, string.Empty));
    LastModifiedUtc = file.LastWriteTimeUtc;
}
```

skofman1: done

PR #103, opened 2016-11-18
Ignore excess nodes in NuSpec in XML transform

scottbommarito:
XSLT works by performing a transformation on all of the nodes in an XML document. These transformations are specified by templates which apply to a subset of the nodes as defined by `match`. When this transform happens, it will throw if any node is not covered by a template. This becomes an issue when someone uploads a `.nuspec` with excess nodes. This additional template applies to all nodes and says "apply the template that applies to you", preventing any excess nodes from causing a transformation issue. The additional nodes pass through our code, which processes the XML into a catalog JSON, which will throw out the meaningless nodes. The excess nodes within the metadata tag will still appear in our catalog, but they will be harmless and ignored by our infrastructure, which is the behavior we have had in the past.


scottbommarito:
@ryuyu @skofman1 @shishirx34

PR #106, opened 2016-12-06
Memory leak Fix

ryuyu:
It turns out that the cause of this was a bug in the AzureDirectorySynchronizer that in turn was caused by a series of interesting decisions on the FileModified API from Lucene.

FSDirectory returns FileModified as milliseconds since 00:00:00 Coordinated Universal Time (UTC), Thursday, 1 January 1970. 
 * https://lucenenet.apache.org/docs/3.0.3/d3/d69/_f_s_directory_8cs_source.html

RAMDirectory returns FileModified as milliseconds since 0:00:00 UTC on January 1, 0001.
 * https://lucenenet.apache.org/docs/3.0.3/d7/df5/_r_a_m_directory_8cs_source.html
 * https://msdn.microsoft.com/en-us/library/system.datetime.ticks(v=vs.110).aspx

AzureDirectory returns FileModified as whatever ToFileTimeAsUtc returns
 * https://github.com/azure-contrib/AzureDirectory/blob/master/AzureDirectory/AzureDirectory.cs#L147
 * https://msdn.microsoft.com/en-us/library/system.datetime.tofiletimeutc(v=vs.110).aspx

Since we were comparing to a timestamp that was relative to Unix epoch (00:00:00 Coordinated Universal Time (UTC), Thursday, 1 January 1970), we were not cleaning up any files that were being put in the RAMDirectory, which was the primary cause of the huge memory growth that we were seeing.

The proposed fix here is to make RAMDirectory, AzureDirectory, and our reference timestamp all use the same return. In this case, I have chosen to use the return from ToFileTimeUtc. Comparing this values gives us the proper delete behaviour in the AzureDirectorySynchronizer.

### DIFF
--- a/NuGet.Services.Metadata.sln
+++ b/NuGet.Services.Metadata.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.25123.0
+VisualStudioVersion = 14.0.25420.1
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tools", "Tools", "{C86C6DEE-84E1-4E4E-8868-6755D7A8E0E4}"
 EndProject
--- a/NuGet.Services.Metadata.sln
+++ b/NuGet.Services.Metadata.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.25420.1
+VisualStudioVersion = 14.0.25123.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tools", "Tools", "{C86C6DEE-84E1-4E4E-8868-6755D7A8E0E4}"
 EndProject
--- a/src/NuGet.Indexing/AzureDirectorySynchronizer.cs
+++ b/src/NuGet.Indexing/AzureDirectorySynchronizer.cs
@@ -128,6 +128,7 @@ namespace NuGet.Indexing
 
         private static long LuceneTimestampFromDateTime(DateTime date)
         {
+            // Use ToFileTimeUtc here to stay consistent with the returns from AzureDirectory.
             return date.ToFileTimeUtc();
         }
     }
--- a/src/NuGet.Indexing/AzureDirectorySynchronizer.cs
+++ b/src/NuGet.Indexing/AzureDirectorySynchronizer.cs
@@ -104,7 +104,7 @@ namespace NuGet.Indexing
 
             // we'll remove old files from both AzureDirectory's cache directory, as well as our destination directory
             // (only when older than 45 minutes - old files may still have active searches on them so we need a margin)
-            var referenceTimestamp = LuceneTimestampFromDateTime(DateTimeOffset.UtcNow.AddMinutes(-45));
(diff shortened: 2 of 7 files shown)


NuGet.Services.Metadata.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.25123.0
+VisualStudioVersion = 14.0.25420.1
```
scottbommarito: Not sure if we should change this.

src/NuGet.Indexing/RAMDirectoryWrapper.cs
```diff
@@ -0,0 +1,34 @@
+﻿using Lucene.Net.Store;
+using System;
+
+namespace NuGet.Indexing
+{
+    // This class wraps the RAMDirectory so we can correct the return of FileModified function
```
scottbommarito: should be an xml doc comment

```
/// <summary>
/// This class wraps the RAMDirectory so we can correct the return of FileModified function
/// </summary>
```

src/NuGet.Indexing/RAMDirectoryWrapper.cs
```diff
@@ -0,0 +1,34 @@
...
+        public RAMDirectoryWrapper() : base() { }
+
+        public RAMDirectoryWrapper(Directory seedDirectory) : base(seedDirectory) { }
+
+        /// <summary>
+        /// Returns the time (as a long) the named file was last modified in a Windows FileTime in UTC
+        /// </summary>
+        /// <param name="name">File Name</param>
+        /// <returns>A long that represents a UTC Windows FileTime</returns>
+        /// <remarks>The implmentation here is to keep in line with the implmentation is AzureDirectory for use in the AzureDirectorySychronizer.
```
scottbommarito: typos in `implementation` and `AzureDirectorySynchronizer`

scottbommarito: Also, to reference a class, use `<see cref="AzureDirectory"/>` and `<see cref="AzureDirectorySynchronizer"/>`

NuGet.Services.Metadata.sln
skofman1: revert this file

src/NuGet.Indexing/AzureDirectorySynchronizer.cs
```diff
@@ -104,7 +104,7 @@ private static void UnidirectionalSync(AzureDirectory sourceDirectory, Directory
 
             // we'll remove old files from both AzureDirectory's cache directory, as well as our destination directory
             // (only when older than 45 minutes - old files may still have active searches on them so we need a margin)
-            var referenceTimestamp = LuceneTimestampFromDateTime(DateTimeOffset.UtcNow.AddMinutes(-45));
+            var referenceTimestamp = LuceneTimestampFromDateTime(DateTime.UtcNow.AddMinutes(-45));
```
skofman1: Don't know what kind of UT framework exists for this code, but can we cover this fix with UTs?

ryuyu: There is not currently any unit testing around this code.
Adding some would be a nontrivial effort as we don't currently have any framework to load/unload a consistent set of files.
The other stuff used here (ie the Directory implementations) are not our implementations.
I guess we could add testing around those, but I struggle to see the value add there.

NuGet.Services.Metadata.sln
ryuyu: reverted

src/NuGet.Indexing/RAMDirectoryWrapper.cs
ryuyu: fixed

PR #107, opened 2016-12-09
Middleware for additional headers

ryuyu:
Adding a middleware to add a place to attach headers to responses.

This is to enable HSTS (https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security).

@skofman1 @shishirx34 @xavierdecoster


ryuyu:
Closed to reopen against right branch (https://github.com/NuGet/NuGet.Services.Metadata/pull/108)

PR #110, opened 2017-01-18
Improve build script

scottbommarito:
Split `build.ps1` into `build.ps1` and `test.ps1`.

Added `buildandtest.ps1` to invoke both at once.


build/common.ps1
```diff
@@ -313,7 +313,8 @@ Function New-Package {
 		[string]$ReleaseLabel = "zlocal",
 		[string]$BuildNumber,
 		[switch]$NoPackageAnalysis,
-		[string]$Version
+		[string]$Version,
```
ryuyu: Is this common.ps1 in sync with the others?

scottbommarito: It isn't, unfortunately. I made changes in the individual `common.ps1` files as they needed them. I think we might need to look into a long term solution for sharing this build code between our repositories (maybe a submodule) because right now it's unmaintainable.

@xavierdecoster any ideas?

xavierdecoster: We used to have a NuGet package that pulled these in, however, it was filled with MSBuild magic which made the whole build process rather hard to understand/modify.
A git submodule may be a bit easier indeed, as it also facilitates traversing history. I could live with that.

Alternatively, we could just host the PowerShell scripts on a GitHub repo, and restore it on disk dynamically before loading it. Similar to how [`dnvm`](https://github.com/aspnet/dnvm) got restored from GitHub. That also provides us with the VCS benefits, and is a simple blob download.

scottbommarito: The simple blob download seems extremely easy, with the reference `common.ps1` being the one in ServerCommon. I'm going to try implementing that.

PR #113, opened 2017-02-02
Chenriks/scan

chenriksson:
5a986aa - refactoring complex lambdas to prevent false positives from fortify null dereference checks
e669bd3 - refactoring to remove duplication and centralize xml apis so they can be secured for fortify scan


chenriksson:
Verified that unit tests passed, and debugged Feed2Catalog for package upload to verify that modified code didn't throw and generated correct catalog JSON.

src/Catalog/Helpers/Utils.cs
```diff
@@ -45,40 +52,131 @@ public static string GetResource(string resName)
...
+            // CodeAnalysis / XslCompiledTransform.Transform: set resolver property to null or instance
+            transform.Transform(reader, arguments, writer, documentResolver: null);
+        }
+
+        private static XslCompiledTransform SafeLoadXslTransform(string resourceName)
         {
-            XslCompiledTransform transform = new XslCompiledTransform();
-            transform.Load(XmlReader.Create(new StreamReader(GetResourceStream(name))));
+            // avoid dictionary cache since we're only working with 2 transforms
+            if (XslTransformNuSpec.Equals(resourceName, StringComparison.InvariantCultureIgnoreCase))
```
scottbommarito: To reduce redundancy, you can rewrite this function as (assuming my comment below is right)...

```
var isXslTransformNuSpec = XslTransformNuSpec.Equals(resourceName, StringComparison.InvariantCultureIgnoreCase);
var isXslTransformNormalizeNuSpec = XslTransformNormalizeNuSpecNamespace.Equals(resourceName, StringComparison.InvariantCultureIgnoreCase);

if (isXslTransformNuSpec || isXslTransformNormalizeNuSpec)
{
    var cache = isXslTransformNuSpec ? XslTransformNuSpecCache : XslTransformNormalizeNuSpecNamespaceCache;
    if (cache == null)
    {
        var resource = isXslTransformNuSpec ? XslTransformNuSpec : XslTransformNormalizeNuSpecNamespace;
        cache = SafeLoadXslTransformNoCache(resource);
    }
    return cache;
}

return SafeLoadXslTransformNoCache(resourceName);
```

src/Catalog/Helpers/Utils.cs
```diff
@@ -45,40 +52,131 @@ public static string GetResource(string resName)
...
+                arguments = new XsltArgumentList();
+            }
+
+            var transform = SafeLoadXslTransform(xsltResourceName);
+
+            // CodeAnalysis / XslCompiledTransform.Transform: set resolver property to null or instance
+            transform.Transform(reader, arguments, writer, documentResolver: null);
+        }
+
+        private static XslCompiledTransform SafeLoadXslTransform(string resourceName)
```
scottbommarito: Did you get the contents of this function from anywhere or was it a change you made for Fortify?

src/Catalog/Helpers/Utils.cs
```diff
@@ -24,6 +25,12 @@ namespace NuGet.Services.Metadata.Catalog
 {
     public class Utils
```
dtivel: This class should be `static`.

src/Catalog/Helpers/Utils.cs
dtivel: You can remove this method by using `Lazy<T>` to initialize the fields:

```
private static readonly Lazy<XslCompiledTransform> XslTransformNuSpecCache
    = new Lazy<XslCompiledTransform>(() => SafeLoadXslTransformNoCache(XslTransformNuSpec));
private static readonly Lazy<XslCompiledTransform> XslTransformNormalizeNuSpecNamespaceCache
    = new Lazy<XslCompiledTransform>(() => SafeLoadXslTransformNoCache(XslTransformNormalizeNuSpecNamespace));
```

...and then access the fields using `.Value`.

Also, rename `SafeLoadXslTransformNoCache(...)` to `SafeLoadXslTransform(...)`.

src/Catalog/Helpers/Utils.cs
```diff
@@ -123,25 +221,12 @@ public static ZipArchive GetPackage(Stream stream)
...
-                XslCompiledTransform xslt = new XslCompiledTransform();
-                xslt.Load(XmlReader.Create(new StreamReader(GetResourceStream("xslt.normalizeNuspecNamespace.xslt"))));
-                xslt.Transform(original.CreateReader(), writer);
-            }
-
-            return result;
-        }
-
+        // revisit to remove dead code (including view.xslt, view.html)? - skip caching for now
         public static string CreateHtmlView(Uri resource, string frame, string baseAddress)
```
dtivel: Where is this method called?  Why not cache the "xslt.view.xslt" transform as well?

tests/NuGet.Services.BasicSearchTests/TestSupport/StartedWebApp.cs
```diff
@@ -7,14 +7,14 @@
...
+using System.Xml;
 using System.Xml.Serialization;
 using Lucene.Net.Store;
 using Microsoft.Owin.Hosting;
 using Newtonsoft.Json;
 using NuGet.Services.BasicSearch;
-using NuGet.Services.BasicSearch.Configuration;
 using NuGet.Services.Configuration;
-using NuGet.Services.KeyVault;
+using Formatting = Newtonsoft.Json.Formatting;
```
chenriksson: Formatting exists in 2 namespaces

src/Catalog/Helpers/Utils.cs
chenriksson: Code was pre-existing, but I'm refactoring and removing duplication in order to centralize the security fixes to these Safe** methods.

src/Catalog/Helpers/Utils.cs
chenriksson: It's dead code, and wasn't originally cached. Original cache was concurrent dictionary on CatalogContext, only used for nuspec and normalizeNamespace transforms.

Will go ahead and move the dead code in this change.

src/Catalog/Helpers/Utils.cs
```diff
@@ -44,38 +51,94 @@ public static string GetResource(string resName)
...
+        
+        public static IGraph CreateNuspecGraph(XDocument nuspec, string baseAddress, bool normalizeXml = false)
         {
-            nuspec = NormalizeNuspecNamespace(nuspec);
-
-            XslCompiledTransform transform = CreateTransform("xslt.nuspec.xslt");
-
             XsltArgumentList arguments = new XsltArgumentList();
-            arguments.AddParam("base", "", baseAddress + "packages/");
+            arguments.AddParam("base", "", baseAddress);
```
xavierdecoster: Why this change? What's the impact?

chenriksson: See NuSpecStage.cs - just moving "packages/" from helper to caller. This was necessary to combine duplicate code (Utils.CreateNuSpecGraph, and CatalogItem.CreateNuSpecGraph).

src/Catalog/Helpers/Utils.cs
```diff
@@ -44,38 +51,94 @@ public static string GetResource(string resName)
...
+
+            foreach (XmlNode childNode in xmlNode.ChildNodes)
+            {
+                NormalizeXml(childNode);
+            }
+        }
+
+        internal static XmlDocument SafeCreateXmlDocument(XmlReader reader = null)
+        {
+            // CodeAnalysis / XmlDocument: set the resolver to null or instance
```
scottbommarito: Is there an error code associated with the CodeAnalysis here? Usually it's like AA000 or something. Could we add that to the messages here?

chenriksson: Good idea... but I don't see any code in the opened issues, or the output from my local scan.

scottbommarito: Ah...if you did some more digging in the logs of the builds you could probably find the specific codes but not worth the hassle.

PR #118, opened 2017-02-09
Export functional test results as XML and add Configure-FunctionalTests.ps1

scottbommarito:
Necessary for VSTS integration...basically a clone of the changes from https://github.com/NuGet/NuGetGallery/pull/3526


tests/Scripts/RunBasicSearchFunctionalCoreTests.bat
```diff
@@ -6,12 +6,18 @@ cd ..
 REM Configuration
 set config=Release
 set solutionPath="BasicSearchTests.FunctionalTests.sln"
+set exitCode=0
 
 REM Required Tools
 set msbuild="%WINDIR%\Microsoft.NET\Framework\v4.0.30319\msbuild"
 set xunit=".\packages\xunit.runner.console.2.0.0\tools\xunit.console.exe"
```
joelverhagen: Can we bump to 2.1.0 while you're in here? It supports `-Verbose` which makes debugging easier.

PR #119, opened 2017-02-09
Adding AI Exception tracking

ryuyu:
Adding a middleware that tracks our exception as AI exceptions.

@skofman1 @joelverhagen @shishirx34

### DIFF
--- a/src/NuGet.ApplicationInsights.Owin/ExceptionTrackingMiddleware.cs
+++ b/src/NuGet.ApplicationInsights.Owin/ExceptionTrackingMiddleware.cs
@@ -1,0 +1,48 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.ApplicationInsights.Extensibility;
+using Microsoft.Owin;
+using System;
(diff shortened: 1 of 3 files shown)


src/NuGet.Services.BasicSearch/Startup.cs
```diff
@@ -314,6 +317,7 @@ private void TrackIndexMetrics(NuGetSearcherManager searcherManager, SearchTelem
             catch (Exception e)
             {
                 await _responseWriter.WriteResponseAsync(context, e, _logger);
+                throw e;
```
shishirx34: There is ClientException catch above this block, when are these thrown? Do we want to log it?

joelverhagen: Don't `throw e;`. `throw;`

src/NuGet.ApplicationInsights.Owin/ExceptionTrackingMiddleware.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.ApplicationInsights;
```
joelverhagen: Sort usings

joelverhagen: System first

PR #121, opened 2017-02-15
Functional test hotfix to dev

scottbommarito:
Not sure why it shows these extra commits--did somebody squash when merging to master?


tests/Scripts/Configure-FunctionalTests.ps1
```diff
@@ -0,0 +1,43 @@
...
+    {
+        Write-Host "Logging into Azure as service principal."
+        # Log into Azure using a service principal with a certificate
+        $login = Add-AzureRmAccount -ApplicationId "$ApplicationId" -CertificateThumbprint "$AzureCertificateThumbprint" -ServicePrincipal -SubscriptionId "$SubscriptionId" -TenantId "$TenantId"
+        # Get the resource group name of the cloud service resource
+        $resourceGroupName = (Find-AzureRmResource -ResourceNameEquals "$CloudServiceName").ResourceGroupName
+        # Use the resource group name to construct the id of the slot resource
+        $slotResource = Get-AzureRmResource -Id "/subscriptions/$SubscriptionId/resourceGroups/$resourceGroupName/providers/Microsoft.ClassicCompute/domainNames/$CloudServiceName/slots/$Slot"
+        # Get the uri of the slot resource and make sure it is https
+        $SearchServiceUrl = ($slotResource.Properties.uri).Replace("http", "https")
```
dtivel: If at some point `uri` returns "https://..." this would result in "httpss://...", which sounds super secure, but really isn't.

tests/Scripts/Configure-FunctionalTests.ps1
```diff
@@ -0,0 +1,43 @@
...
+    }
+    Catch [System.Exception]
+    {
+        Write-Host "Failed to retrieve URL for testing!"
+        Write-Host $_.Exception.Message
```
dtivel: The callstack is useful information to report.  You get the message, callstack and exception type with `Exception.ToString()`.

tests/Scripts/Configure-FunctionalTests.ps1
```diff
@@ -0,0 +1,43 @@
...
+    $SearchServiceUrl = $ServiceRoot
+}
+else
+{
+    # Use Azure PowerShell cmdlets to find the url of the desired slot
+    Try
+    {
+        Write-Host "Logging into Azure as service principal."
+        # Log into Azure using a service principal with a certificate
+        $login = Add-AzureRmAccount -ApplicationId "$ApplicationId" -CertificateThumbprint "$AzureCertificateThumbprint" -ServicePrincipal -SubscriptionId "$SubscriptionId" -TenantId "$TenantId"
```
dtivel: `$login` isn't used.

scottbommarito: Good catch...I think at some point I used it but no longer needed to. The function still needs to be called though, but `$login` does not to be saved.

tests/Scripts/Configure-FunctionalTests.ps1
scottbommarito: That's unfortunate...should probably change to `($slotResource.Properties.uri).Replace("http:", "https:")`

scottbommarito:
@dtivel thanks for the comments. I will merge this (because I don't want to make changes directly to master) and open a pull request with the fixes you've suggested.

PR #123, opened 2017-02-23
Issue3531

cristinamanum:
Fix for issue 3531. The blobs from registration storage were not compressed. The stats for the container  v3-registration1-gz in nugetgallery are:
Total blobs: 1108173 = Zipped:979557 + NotZipped:128616

In addition to these changes we need to:1 decide what to do with the not compressed blobs and 2. The CDN profile is already  set to compress the blobs, in this case the question is what is the scenario that this data source supports?


src/Ng/CommandHelpers.cs
```diff
@@ -105,7 +105,7 @@ public static StorageFactory CreateCompressedStorageFactory(IDictionary<string,
                 { Arguments.StoragePath, Arguments.CompressedStoragePath }
             };
 
-            return CreateStorageFactoryImpl(arguments, names, verbose);
+            return CreateStorageFactoryImpl(arguments, names, verbose, compressed:true);
```
scottbommarito: `compressed: true`

src/Ng/CommandHelpers.cs
```diff
@@ -127,7 +127,10 @@ public static StorageFactory CreateSuffixedStorageFactory(string suffix, IDictio
             return CreateStorageFactoryImpl(arguments, names, verbose);
         }
 
-        private static StorageFactory CreateStorageFactoryImpl(IDictionary<string, string> arguments, IDictionary<string, string> argumentNameMap, bool verbose)
+        private static StorageFactory CreateStorageFactoryImpl(IDictionary<string, string> arguments, 
+                                                               IDictionary<string, string> argumentNameMap,
+                                                               bool verbose,
+                                                               bool compressed = false)
```
joelverhagen: nit: did do we need optional parameter here? can we make the caller explicitly specify this in all cases? Given there have been multiple bugs in this area, i would prefer explicit over implicit

scottbommarito: All of the other arguments but `compressed` are passed to this function through a dictionary. By adding this additional argument as a parameter, you've created a separate source of arguments for the method. I would prefer the consistency of adding a `Arguments.StorageIsCompressed` to the dictionary, but ultimately this isn't a big deal and this implementation is fine.

src/Ng/CommandHelpers.cs
cristinamanum: Added unit tests.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -110,13 +110,15 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
                     await blob.UploadFromStreamAsync(destinationStream, cancellationToken);
+                    Trace.WriteLine(String.Format("Saved compressed blob {0} to container {1}", blob.Uri.ToString(), _directory.Container.Name));
                 }
             }
             else
             {
                 using (Stream stream = content.GetContentStream())
                 {
                     await blob.UploadFromStreamAsync(stream, cancellationToken);
+                    Trace.WriteLine(String.Format("Saved not compressed blob {0} to container {1}", blob.Uri.ToString(), _directory.Container.Name));
```
cristinamanum: Updated.

src/Ng/CommandHelpers.cs
cristinamanum: Updated.

src/Ng/CommandHelpers.cs
cristinamanum: Updated.

src/Ng/CommandHelpers.cs
cristinamanum: There are different types of parameters. The ones from dictionary are mapped form external arguments. This is an internal argument. I let it at it is for now.

tests/NgTests/StorageFactoryTests.cs
```diff
@@ -0,0 +1,115 @@
...
+            StorageFactory factory = CommandHelpers.CreateStorageFactory(arguments, true);
+            AzureStorageFactory azureFactory = factory as AzureStorageFactory;
+            // Assert
+            Assert.True(azureFactory != null);
+            Assert.False(azureFactory.CompressContent, "The azure storage factory should not compress the content.");
+        }
+    }
+
+    public class Arguments
+    {
```
skofman1: why duplicate those consts?

cristinamanum: There is a private version in ng. I preferred to add what needed at the test level instead of making that public.

skofman1: What is the importantance of the argument names to be private? Another option is to set them as "internal", and set the test project as "internals visible".

cristinamanum: Made them public. Having them public it is harder to change/delete them.

tests/NgTests/StorageFactoryTests.cs
```diff
@@ -0,0 +1,95 @@
...
+                { Arguments.CompressedStorageContainer, storageContainer },
+                { Arguments.CompressedStoragePath, storagePath },
+                { Arguments.StorageType, storageType},
+                { Arguments.UseCompressedStorage, "true"}
+            };
+
+            StorageFactory factory =  CommandHelpers.CreateCompressedStorageFactory(arguments, true);
+            AzureStorageFactory azureFactory = factory as AzureStorageFactory;
+            // Assert
+            Assert.True(azureFactory!=null);
```
dtivel: For this and subsequent tests you can use `Assert.Null(...)` or `Assert.NotNull(...)` instead of `Assert.True(...)` and `Assert.False(...)`.

cristinamanum: As discussed offline, added messages where missed.

PR #128, opened 2017-02-28
Update Search Diagnostics Endpoint to return more data

shishirx34:
Updating search service endpoint to show more relevant data for monitoring. Fixes https://github.com/NuGet/NuGetGallery/issues/3551


src/NuGet.Indexing/AuxiliaryDataFileNames.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Indexing
+{
+    class AuxiliaryDataFileNames
```
joelverhagen: qualify access, e.g. `public`

src/NuGet.Indexing/FileLoader.cs
```diff
@@ -28,6 +28,14 @@ public JsonReader GetReader(string name)
             return new JsonTextReader(new StreamReader(fullName));
         }
 
+        public DateTime? GetLastUpdateTime(string name)
```
joelverhagen: Please use DateTimeOffset everywhere. Reduces timezone nightmares.

src/NuGet.Indexing/NuGetSearcherManager.cs
```diff
@@ -19,20 +19,31 @@ namespace NuGet.Indexing
...
     {
         public static readonly TimeSpan AuxiliaryDataRefreshRate = TimeSpan.FromHours(1);
-        
-        private readonly FrameworkLogger _logger;
 
-        private DateTime _lastTimeIndexReloaded = DateTime.MinValue;
-        private readonly TimeSpan _indexReloadRate;
+        public virtual string IndexName => _indexProvider.GetIndexContainerName();
+        public IDictionary<string, Uri> RegistrationBaseAddress { get; }
+        public virtual DateTime LastTimeIndexReloaded { get; private set; }
```
joelverhagen: why `virtual`?

src/NuGet.Indexing/NuGetSearcherManager.cs
```diff
@@ -19,20 +19,31 @@ namespace NuGet.Indexing
...
         private readonly ILoader _loader;
-
-        private DateTime _lastTimeAuxiliaryDataRefreshed = DateTime.MinValue;
+        private readonly TimeSpan _indexReloadRate;
         private readonly TimeSpan _auxiliaryDataRefreshRate;
         private readonly IDictionary<string, HashSet<string>> _owners = new Dictionary<string, HashSet<string>>(StringComparer.OrdinalIgnoreCase);
         private readonly IDictionary<string, HashSet<string>> _curatedFeeds = new Dictionary<string, HashSet<string>>();
         private readonly Downloads _downloads = new Downloads();
         private IReadOnlyDictionary<string, int> _rankings;
+        private readonly string[] AuxiliaryDataFileList = new string[] {
```
joelverhagen: nit: put the curly after `string[]` on a new line

src/NuGet.Indexing/NuGetIndexSearcher.cs
```diff
@@ -62,6 +62,12 @@ public class NuGetIndexSearcher : IndexSearcher
         public OpenBitSet LatestStableBitSet { get; }
         public OwnersResult Owners { get; }
         public DateTime LastReopen { get; }
+        public DateTime LastIndexReloadTime => Manager.LastTimeIndexReloaded;
+        public long LastIndexReloadDurationInMilliseconds => Manager.LastIndexReloadDurationInMilliseconds;
+        public DateTime LastAuxiliaryDataLoadTime => Manager.LastTimeAuxiliaryDataRefreshed;
+        public IDictionary<string, DateTime?> LastAuxiliaryDataUpdateTime => Manager.AuxiliaryDataLastUpdateTime;
```
joelverhagen: Why is this a dictionary? Don't we know all of the keys at compile time? I think a strongly typed object is more maintainable/discoverable.

src/NuGet.Indexing/FileLoader.cs
```diff
@@ -28,6 +28,14 @@ public JsonReader GetReader(string name)
...
+        {
+            string fullName = _folder == null ? name : _folder + name;
```
joelverhagen: `Path.Combine` instead of string concatenation.

src/NuGet.Indexing/ResponseFormatter.cs
```diff
@@ -346,7 +346,19 @@ public static void WriteStatsResult(JsonWriter jsonWriter, NuGetIndexSearcher se
             jsonWriter.WriteStartObject();
             WriteProperty(jsonWriter, "numDocs", searcher.IndexReader.NumDocs());
             WriteProperty(jsonWriter, "indexName", searcher.Manager.IndexName);
+            WriteProperty(jsonWriter, "machineName", searcher.MachineName);
             WriteProperty(jsonWriter, "lastReopen", searcher.LastReopen);
+            WriteProperty(jsonWriter, "lastIndexReloadTime", searcher.LastIndexReloadTime);
+            WriteProperty(jsonWriter, "lastIndexReloadDurationInMilliseconds", searcher.LastIndexReloadDurationInMilliseconds);
```
skofman1: lastIndexReloadDurationInMilliseconds is not useful to expose using a diag endpoint. It's a metric that needs to be viewed in comparison to past information, and hence should be exposed as a perf counter/ApplicationInsights metric. I would recommend removing it from this view.

src/NuGet.Indexing/ILoader.cs
```diff
@@ -16,5 +16,7 @@ public interface ILoader
         /// <param name="config">Configuration to use.</param>
         /// <returns>Returns true if the data has changed.</returns>
         bool Reload(IndexingConfiguration config);
+
+        DateTime? GetLastUpdateTime(string name);
```
joelverhagen: For my own understanding, why is the file's last write time the key information here? Can't the reload method record when it latest returned true? This is more of an open question than a request for change.

src/NuGet.Indexing/StorageLoader.cs
```diff
@@ -44,6 +41,35 @@ public JsonReader GetReader(string name)
...
+                {
+                    return null;
+                }
+
+                blob.FetchAttributes();
+
+                // LastModified time is always in UTC for Azure blobs
+                var lastModifiedUtcTime = blob.Properties.LastModified.HasValue
+                    ? blob.Properties.LastModified.Value
+                    : DateTimeOffset.MinValue;
```
skofman1: >   : DateTimeOffset.MinValue; [](start = 17, length = 29)

why fallback to DateTimeOffset.MinValue instead of returning null?

src/NuGet.Indexing/StorageLoader.cs
```diff
@@ -61,5 +87,12 @@ public bool Reload(IndexingConfiguration config)
                 !(oldDataContainerName == _dataContainerName &&
                   oldStorageAccountConnectionString == _storageAccountConnectionString);
         }
+
+        private CloudBlockBlob GetBlob(string blobName)
+        {
+            var client = _storageAccount.CreateCloudBlobClient();
```
skofman1: >   var client = _storageAccount.CreateCloudBlobClient(); [](start = 9, length = 56)

we can reuse the client and container in the different requests

skofman1:
>         public void WriteStatsResultTest(

can you please add tests that include edge cases such as exceptions thrown, nulls (because a file wasn't loaded yet), etc.

---
Refers to: tests/NuGet.IndexingTests/ResponseFormatterTests.cs:62 in 3941f06. [](commit_id = 3941f06848678b1b3dec75eb837b2264f3580f30, deletion_comment = False)

tests/NuGet.IndexingTests/ResponseFormatterTests.cs
```diff
@@ -65,16 +65,24 @@ public void WriteV2CountResultTest(int totalHits, string expected)
             Dictionary<string, string> commitUserData,
```
skofman1: also, please modify functional tests to verify the new features

src/NuGet.Indexing/ResponseFormatter.cs
```diff
@@ -346,7 +346,19 @@ public static void WriteStatsResult(JsonWriter jsonWriter, NuGetIndexSearcher se
             jsonWriter.WriteStartObject();
             WriteProperty(jsonWriter, "numDocs", searcher.IndexReader.NumDocs());
```
scottbommarito: I think this would look a lot cleaner if `SearcherManager` had a dictionary mapping property name to value, i.e.
```
{ numDocs, () => IndexReader.NumDocs()},
{ indexName, () => Manager.IndexName},
...
```

This code would be reduced significantly, and it would be very easy to add additional diagnostics.
```
foreach (var diagnostic in searcher.Diagnostics) {
    WriteProperty(jsonWriter, diagnostic.Name, diagnostic.GetValue());
}
```

src/NuGet.Indexing/ResponseFormatter.cs
shishirx34: @ryuyu - you want to take this?

src/NuGet.Indexing/StorageLoader.cs
shishirx34: No we can't. The `_storageAccount` changes on Reload, see above.

src/NuGet.Indexing/ResponseFormatter.cs
shishirx34: I went as per @ryuyu's requirements in https://github.com/NuGet/NuGetGallery/issues/3551. If my understanding is correct, these values will be read by the runner which will ultimately expose these as perf counters to geneva. Peter can elaborate.

src/NuGet.Indexing/ILoader.cs
shishirx34: I think this is to get the update time of the resource independent of the time it was reloaded. If for some reason the reload hasn't succeeded and we have a modified auxiliary file for sometime, this will tell us something is wrong or how long the data has remained stale. @ryuyu.

src/NuGet.Indexing/FileLoader.cs
shishirx34: we have a lot of inconsistencies in the usage of our time. I am trying to return all the times in UTC, however changing this to DateTimeOffset would mean changing other data(reopen time, reload time) in the diag endpoint to be consistent. I would rather keep this the same DateTime object, just ensuring that it is UTC, we need to update our commit-time-stamp values to be consistent as well to be UTC. I am not sure if this data is being used anywhere else and being parsed correctly or not, so I'd be hesitant to change it here.

src/NuGet.Indexing/ResponseFormatter.cs
shishirx34: I think this is an unrelated change to this PR, and simply doing the mapping is not enough. You need to check if the value is an object and recursively follow the object's properties to correctly marshal them into JSON. #won't fix

src/NuGet.Indexing/NuGetSearcherManager.cs
```diff
@@ -48,19 +59,21 @@ public class NuGetSearcherManager : SearcherManager<NuGetIndexSearcher>
...
 
             _logger = logger;
-            
+
             RegistrationBaseAddress = new Dictionary<string, Uri>();
 
             _indexProvider = indexProvider;
             _loader = loader;
 
+            LastTimeIndexReloaded = DateTime.MinValue;
```
shishirx34: sure.

src/NuGet.Indexing/ResponseFormatter.cs
scottbommarito: I accept that this is outside of the scope of the PR, but wouldn't C# handle any complexities in type conversions through `WriteProperty<T>`'s use of generics? I don't think there would be any more code necessary than what I put here.

Let me fix the second code example though...
```
foreach (var diagnosticName in searcher.Diagnostics.Keys) {
    WriteProperty(jsonWriter, diagnosticName, searcher.Diagnostics[diagnosticName]);
}
```

shishirx34: No `WriteProperty<T>` is our code not some magic that c# does. It will not handle it we need to add the logic to identify and serialize object. http://www.newtonsoft.com/json/help/html/M_Newtonsoft_Json_JsonWriter_WriteValue_29.htm

scottbommarito: There doesn't need to be any complex serialization logic. Just declare the dictionary in `SearcherManager` as `IDictionary<string, Func<string>>` and directly write the transformation there. All of our diagnostics are already strings.

```
foreach (var diagnosticName in searcher.Diagnostics.Keys) {
    WriteProperty(jsonWriter, diagnosticName, searcher.Diagnostics[diagnosticName]());
}
```

src/NuGet.Indexing/ResponseFormatter.cs
ryuyu: Currently, we don't have the service itself reporting any metrics. We can either have the service itself report these counters to geneva, or have this endpoint report it and have the runner report to geneva.

We current have the structure for the latter, and i don't think we should split it up and report some metrics directly from the service and others from the runner. Since the runner is already there, I would like to leverage that, but if we want to move to letting each instance report directly, that's a change we should discuss with the wider team (ie should all our services self report).

src/NuGet.Indexing/ResponseFormatter.cs
shishirx34: had an offline chat with Scott. resolved this issue as wont fix since it is unrelated to this PR.

scottbommarito: @shishirx34 and I spoke offline and resolved our confusions. I was specifically referring to the list of properties at the top of the file, not the part where we construct a series of objects.

src/NuGet.Indexing/ILoader.cs
ryuyu: From my understanding, there are few key timestamps here.
1. Last time the aux reload was attempted
2. last time the aux reload succeeded
3. Timestamp on the currently loaded aux files.

Since there is some delay on when the aux files are recorded and reloaded, there can be cases where the last reload changes, but the timestamp on the actual file has not changed. This is due to all our updates/reloads being on different cadences (ie index reload callback happens every X minutes, while aux data timeout is another every X minute, and aux data update [from jobs] is every X hours).

Since this is an endpoint view, we want to see when and if there is a problem here. Last reload time of this would tell us if the service is working, whereas the last timestamp on the file could potentially tell us if updating the files is failing. Ideally, we would have separate monitoring of the auxiliary data files, at which point the last updated timestamp information here could be deprecated.

src/NuGet.Indexing/AuxiliaryDataFileNames.cs
shishirx34: I do not think each auxiliary data file needs to be represented as an individual class. There aren't direct actionable items on each file as such. That would be an overkill of the design. 

That said, I have changed this file to be AuxiliaryFiles class which would hold the list of these files. If we were to add any new files in, we can just add it to the list in here and update the list of files update times accordingly from `NugetSeracherManager`.

src/NuGet.Indexing/FileLoader.cs
joelverhagen: Sure, could we have a tracking item for this future refactoring?

src/NuGet.Indexing/ILoader.cs
joelverhagen: Great, makes sense!

src/NuGet.Indexing/FileLoader.cs
shishirx34: https://github.com/NuGet/Engineering/issues/299

src/NuGet.Indexing/ResponseFormatter.cs
skofman1: We should have a clear seperation between reporting status (like last reload time, last successful reload time), and reporting metrics (like reload speed, CPU utilization). The first need to be monitored by a job, and exposed directly by the service. The second should be reported to Geneva or AI, and monitored as metrics (like we currently do for CPU). 
The fact that MDM jobs can also expose metrics, shouldn't be a path to get our metrics into Geneva, we should use a more direct method to get them there.

src/NuGet.Indexing/NuGetSearcherManager.cs
shishirx34: Its so that these values can be overridden from the `Mock<>` in tests.

tests/NuGet.IndexingTests/ResponseFormatterTests.cs
shishirx34: https://github.com/NuGet/NuGetGallery/issues/3652 - Need to check this in first to add functional tests.

src/NuGet.Indexing/StorageLoader.cs
skofman1: ?

shishirx34: I changed it to return null.

shishirx34: Please take a look at the latest commit. https://github.com/NuGet/NuGet.Services.Metadata/pull/128/commits/baa858d42373598638af4d0061054c46be3a515b#diff-abd3dac6b87987c81b2d39954087afdaR64

src/NuGet.Indexing/NuGetSearcherManager.cs
```diff
@@ -19,15 +19,19 @@ namespace NuGet.Indexing
...
-        
-        private readonly FrameworkLogger _logger;
 
-        private DateTime _lastTimeIndexReloaded = DateTime.MinValue;
-        private readonly TimeSpan _indexReloadRate;
+        public virtual AuxiliaryFiles AuxiliaryFiles { get; private set; }
+        public virtual string IndexName => _indexProvider.GetIndexContainerName();
+        public virtual long LastIndexReloadDurationInMilliseconds { get; private set; } = -1;
+        public virtual DateTime LastTimeIndexReloaded { get; private set; } = DateTime.MinValue;
+        public virtual DateTime LastTimeAuxiliaryDataRefreshed { get; private set; } = DateTime.MinValue;
```
shishirx34: Ah yes, missed this when I defaulted to null in other places. Will do.

src/NuGet.Indexing/AuxiliaryFiles.cs
```diff
@@ -0,0 +1,44 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace NuGet.Indexing
+{
+    public class AuxiliaryFiles
+    {
+        // List of all auxiliary files
+        public const string Owners = "owners.json";
```
shishirx34: Yes these can be internal. Changed it.

src/NuGet.Indexing/ResponseFormatter.cs
```diff
@@ -346,7 +346,19 @@ public static void WriteStatsResult(JsonWriter jsonWriter, NuGetIndexSearcher se
...
             WriteProperty(jsonWriter, "indexName", searcher.Manager.IndexName);
+            WriteProperty(jsonWriter, "machineName", searcher.Manager.MachineName);
             WriteProperty(jsonWriter, "lastReopen", searcher.LastReopen);
+            WriteProperty(jsonWriter, "lastIndexReloadTime", searcher.Manager.LastTimeIndexReloaded);
+            WriteProperty(jsonWriter, "lastIndexReloadDurationInMilliseconds", searcher.Manager.LastIndexReloadDurationInMilliseconds);
+            WriteProperty(jsonWriter, "lastAuxiliaryDataLoadTime", searcher.Manager.LastTimeAuxiliaryDataRefreshed);
```
shishirx34: Earlier I was passing the values through a wrapper class which had the naming consistency, but I realized that passing through was not necessary. Aligning these properties now.

src/NuGet.Indexing/ILoader.cs
shishirx34: This change is relevant to `StorageLoader`, here I think the `name` is appropriate. In `FileLoader` it could be `fileName`

src/NuGet.Indexing/StorageLoader.cs
scottbommarito: Low priority, but we could rewrite `Reload` to save the client and then reuse it until `Reload` is called again.

tests/NuGet.IndexingTests/ResponseFormatterTests.cs
```diff
@@ -65,16 +65,54 @@ public void WriteV2CountResultTest(int totalHits, string expected)
...
             string expected)
         {
-            var searcher = new MockSearcher(indexName, numDocs, commitUserData);
+            var currentTime = DateTime.UtcNow;
+            var searcher = new MockSearcher(indexName, numDocs, commitUserData, versions: null, reloadTime: currentTime, lastModifiedTimeForAuxFiles: null, machineName: "TestMachineX");
```
shishirx34: You are right, I will merge the two tests together, however, I think we should keep the member data separate as its testing two different sets of data.

scottbommarito: Oh yeah, I had forgotten you can stack `MemberData`s. I agree.

PR #134, opened 2017-03-08
Fix Intermittent 500's on Autocomplete

ryuyu:
We believe this issue was caused by a race condition in DownloadsByVersion.

Request thread would come in and attempt to refer to Total, which would start the computation of the sum. While this was happening, the auxiliary data would reload on another thread and start writing the dictionary.
Request thread would then fail because the underlying enumerable was mutated. "Collection was modified; enumeration operation may not execute."

Change DownloadsByVersion to keep a rolling total and not compute on reference to Total.

Also includes a minor fix to message logged to AI to actually include the exception.

@joelverhagen @skofman1 @shishirx34

### DIFF
--- a/src/NuGet.Indexing/DownloadsByVersion.cs
+++ b/src/NuGet.Indexing/DownloadsByVersion.cs
@@ -9,10 +9,10 @@ namespace NuGet.Indexing
 {
     public class DownloadsByVersion
     {
-        private IDictionary<string, int> _downloadsByVersion =
+        private readonly IDictionary<string, int> _downloadsByVersion =
             new Dictionary<string, int>(StringComparer.OrdinalIgnoreCase);
 
-        private int? total;
+        private int _total = 0;
 
         /// <summary>
         /// The total count of downloads across all versionss
@@ -24,14 +24,7 @@ namespace NuGet.Indexing
         {
             get
             {
-                if (total.HasValue)
-                {
-                    return total.Value;
-                }
-
-                total = _downloadsByVersion.Values.Sum();
-
-                return total.Value;
+                return _total;
             }
         }
 
@@ -48,8 +41,13 @@ namespace NuGet.Indexing
 
             set
             {
+                if (_downloadsByVersion.ContainsKey(version))
+                {
+                    _total = _total - _downloadsByVersion[version];
+                }
+
                 _downloadsByVersion[version] = value;
-                total = null;
+                _total = _total + value;
             }
         }
     }
--- a/src/NuGet.Indexing/DownloadsByVersion.cs
+++ b/src/NuGet.Indexing/DownloadsByVersion.cs
@@ -15,7 +15,7 @@ namespace NuGet.Indexing
         private int _total;
 
         /// <summary>
-        /// The total count of downloads across all versionss
+        /// The total count of downloads across all versions
         /// </summary>
         /// <remarks>
         /// This is thread safe as long as set is not being called from multiple threads
@@ -39,6 +39,7 @@ namespace NuGet.Indexing
                 return count;
             }
 
+            // Set is only ever called when the auxiliary data is reloaded, which is only ever done on one thread per instance.
             set
             {
                 int oldValue;
--- a/src/NuGet.Indexing/DownloadsByVersion.cs
+++ b/src/NuGet.Indexing/DownloadsByVersion.cs
@@ -12,7 +12,7 @@ namespace NuGet.Indexing
         private readonly IDictionary<string, int> _downloadsByVersion =
             new Dictionary<string, int>(StringComparer.OrdinalIgnoreCase);
 
-        private int _total = 0;
+        private int _total;
 
         /// <summary>
         /// The total count of downloads across all versionss
@@ -41,13 +41,11 @@ namespace NuGet.Indexing
 
             set
             {
-                if (_downloadsByVersion.ContainsKey(version))
-                {
-                    _total = _total - _downloadsByVersion[version];
-                }
-
+                int oldValue;
+                _downloadsByVersion.TryGetValue(version, out oldValue);
                 _downloadsByVersion[version] = value;
-                _total = _total + value;
+
+                _total = _total + value - oldValue;
             }
         }
     }
--- a/src/NuGet.Services.BasicSearch/ResponseHelpers.cs
+++ b/src/NuGet.Services.BasicSearch/ResponseHelpers.cs
@@ -58,7 +58,7 @@ namespace NuGet.Services.BasicSearch
 
         public async Task WriteResponseAsync(IOwinContext context, Exception e, FrameworkLogger logger)
         {
-            logger.LogError("Internal server error", e);
+            logger.LogError("Internal server error: {0}", e);
 
             await WriteResponseAsync(context, HttpStatusCode.InternalServerError, JObject.FromObject(new
             {
--- a/tests/NuGet.IndexingTests/DownloadsByVersionTests.cs
+++ b/tests/NuGet.IndexingTests/DownloadsByVersionTests.cs
@@ -1,0 +1,122 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using NuGet.Indexing;
+using Xunit;
+
+namespace NuGet.IndexingTests
+{
+    public class DownloadsByVersionTests
+    {
+        [Theory]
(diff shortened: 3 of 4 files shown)


src/NuGet.Indexing/DownloadsByVersion.cs
```diff
@@ -9,10 +9,10 @@ namespace NuGet.Indexing
 {
     public class DownloadsByVersion
     {
-        private IDictionary<string, int> _downloadsByVersion =
+        private readonly IDictionary<string, int> _downloadsByVersion =
             new Dictionary<string, int>(StringComparer.OrdinalIgnoreCase);
 
-        private int? total;
+        private int _total = 0;
```
joelverhagen: nit: initializing to zero in unnecessary

src/NuGet.Indexing/DownloadsByVersion.cs
```diff
@@ -48,8 +41,13 @@ public int Total
 
             set
             {
+                if (_downloadsByVersion.ContainsKey(version))
+                {
+                    _total = _total - _downloadsByVersion[version];
```
joelverhagen: There is a window of time where `_total` has less than it did before. Although this window of time is short, it could be confusing and can be avoided by incrementing `_total` by the different (calculated and stored in a local).

Also, use `TryGetValue` to avoid the additional lookup.

Also, this method is setter is NOT thread safe. This is fine because we have a single thread loading aux data, right? Please leave a loud comment stating this.

src/NuGet.Services.BasicSearch/ResponseHelpers.cs
```diff
@@ -58,7 +58,7 @@ public class ResponseWriter
 
         public async Task WriteResponseAsync(IOwinContext context, Exception e, FrameworkLogger logger)
         {
-            logger.LogError("Internal server error", e);
+            logger.LogError("Internal server error: {0}", e);
```
joelverhagen: 👍 👍 👍

src/NuGet.Indexing/DownloadsByVersion.cs
scottbommarito: Because you are operating on the live index, regardless of how you write this method, you will always have the download counts and total out of sync for a single operation. This has little impact, but could create some absolutely nasty issues in the future.

To prevent this, the entire `DownloadsByVersion` object should be constructed in memory and then swapped with the instance contained in the dictionaries in `Downloads`. We should adopt a similar flow for the other auxiliary data loads as well.

joelverhagen: Nastiness is inconsistency between individual version download counts and the stated total. Not too bad given client implementation but as Scott mentions it can be avoided.

cristinamanum: If the issue was caused by concurrency why not make this read/writes thread safe?

ryuyu: Auxiliary data is loaded by a single thread only.

ryuyu: It is only out of sync while the data is being reloaded. And looking further up the call tree into the usage of this object during reload, there is no assumption that the processed data is everything. In fact, the only assumption is that the incoming json file that we are "reloading" contains some download data and will only overwrite previous loads where they overlap. Currently, in practice, I don't THINK we have any cases where the new downloads isn't a superset of the previous one, but without confirmation, I'm hesitant to change it.

src/NuGet.Indexing/DownloadsByVersion.cs
```diff
@@ -9,10 +9,10 @@ namespace NuGet.Indexing
 {
     public class DownloadsByVersion
     {
-        private IDictionary<string, int> _downloadsByVersion =
+        private readonly IDictionary<string, int> _downloadsByVersion =
```
skofman1: please add UTs for this code.

src/NuGet.Indexing/DownloadsByVersion.cs
```diff
@@ -9,10 +9,10 @@ namespace NuGet.Indexing
...
             new Dictionary<string, int>(StringComparer.OrdinalIgnoreCase);
 
-        private int? total;
+        private int _total;
 
         /// <summary>
         /// The total count of downloads across all versionss
```
skofman1: >versionss [](start = 52, length = 9)

versions

src/NuGet.Indexing/DownloadsByVersion.cs
```diff
@@ -48,8 +41,11 @@ public int Total
 
             set
             {
+                int oldValue;
```
skofman1: are the updates done on a single thread? In any case, there should be a comment about that

src/NuGet.Indexing/DownloadsByVersion.cs
scottbommarito: Being out of sync while the data is being reloaded is still a major concern--we always want it to be in sync, and having intermittent sync issues is a nightmare to deal with as you experienced while debugging this issue.

You can still construct the new `DownloadsByVersion` using the old `DownloadsByVersion`. Clone the object, perform the modifications, and then swap. This would get rid of your concern and guarantee that the data is never out of sync.

src/NuGet.Indexing/DownloadsByVersion.cs
ryuyu: done

src/NuGet.Indexing/DownloadsByVersion.cs
ryuyu: done

src/NuGet.Indexing/DownloadsByVersion.cs
ryuyu: I don't want to churn this more than necessary to fix the exception. If this ends up being a real issue or we want to optimize in the future, we can revisit.

scottbommarito: I disagree. Shouldn't take very long to implement at all and will prevent us from ever experiencing future issues.

ryuyu: The time it takes to implement isn't the issue here. We don't know what other consequences an implementation will have, which is why we test. The change currently made is relatively limited in scope, and can be relatively reliably unit tested. Change the mechanism above it, while possibly trivial in implementation, could have further repercussions that we might not catch. Seeing as this is to fix an exception case, it's safer to make the minimal required change with as limited a scope as possible in an easily testable area, and then revisit making a larger change when there is time to thoroughly test it.

ryuyu: The original problem here was that calling total was not thread safe. The set, while not being thread safe, is only ever called from one thread (if this ever turns out to be NOT the case, we have another problem. Setting of this should never be done from anything other than the thread that updates the index/aux data).

joelverhagen: It sounds like you're saying we should separate the HTTP 500 bug fix from addressing the data consistency issue that Scott is referring to. This seems reasonable to me. I don't see  a problem in fixing one thing at a time, as long as the follow-up work is tracked and properly prioritized/scheduled.

joelverhagen: Please make sure there is a GitHub issue for the consistency issue @scottbommarito is talking about before merging, so things don't get lost, if you're not going to fix it as part of this PR.

tests/NuGet.IndexingTests/DownloadsByVersionTests.cs
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        [Fact]
+        public void ReturnZeroForUnknownTest()
+        {
+            var newDownloadsByVersion = new DownloadsByVersion();
+            Assert.Equal(0, newDownloadsByVersion["not.real.version"]);
+        }
+
+        public static IEnumerable<object[]> RollingTotalTestData
```
scottbommarito: I think this code would be a bit more readable if you moved the test data to be closer to the tests themselves rather than grouping them together.

E.g.
```
public static IEnumerable<object[]> RollingTotalTestData
{
    ...
}

public void RollingTotalCountTest(Dictionary<string, int> versionDownloads, int expectedTotal)
{
    ...
}

public static IEnumerable<object[]> UpdateTotalTestData
{
    ...
}

public void UpdateTotalTest(Dictionary<string, int> originalData, Dictionary<string, int> updatedVersionDownloads, int expectedTotal)
{
    ...
}
```

ryuyu:
Just realized this was made against master. Will close and reopen against dev.

PR #135, opened 2017-03-08
Issue292

cristinamanum:
feed2Catalog terminates the proccess with TaskCancelledException on httpClient timeout.  This changes allow the process continuation and log warning in such cases.


src/Ng/Jobs/LoopingNgJob.cs
```diff
@@ -38,7 +38,14 @@ protected LoopingNgJob(ILoggerFactory loggerFactory)
...
 
                 Logger.LogInformation("Running job.");
-                await RunInternal(cancellationToken);
+                try
+                {
+                    await RunInternal(cancellationToken);
+                }
+                catch(TransientException te)
+                {
+                    Logger.LogWarning("Transient exception: {0}. {1}", te.Message, te.InnerException.StackTrace??te.StackTrace);
```
skofman1: >InnerException [](start = 86, length = 14)

InnerException can be null.

src/Ng/Jobs/CatalogUtility.cs
```diff
@@ -171,7 +180,17 @@ private static DateTime ForceUtc(DateTime date)
                     // This query string will ensure the package is not cached
                     // (e.g. on the CDN) and returns the "latest and greatest" package metadata.
                     var packageUri = Utilities.GetNugetCacheBustingUri(packageItem.ContentUri, entry.Key.ToString("O"));
-                    var response = await client.GetAsync(packageUri, cancellationToken);
+                    HttpResponseMessage response = null;
+                    try
+                    {
+                        response = await client.GetAsync(packageUri, cancellationToken);
+                    }
+                    //if the client gets disposed because the timeout a task cancelled exception will happen
```
joelverhagen: nit: please put comment in the `catch` scope

src/Ng/Jobs/CatalogUtility.cs
```diff
@@ -171,7 +180,17 @@ private static DateTime ForceUtc(DateTime date)
                     // This query string will ensure the package is not cached
                     // (e.g. on the CDN) and returns the "latest and greatest" package metadata.
                     var packageUri = Utilities.GetNugetCacheBustingUri(packageItem.ContentUri, entry.Key.ToString("O"));
-                    var response = await client.GetAsync(packageUri, cancellationToken);
+                    HttpResponseMessage response = null;
+                    try
+                    {
+                        response = await client.GetAsync(packageUri, cancellationToken);
```
joelverhagen: Have we considered increasing the HTTP timeout? Will that help? Is it a good idea? 

I don't understand what cases would lead to this sort of timeout.

src/Ng/Jobs/LoopingNgJob.cs
```diff
@@ -38,7 +38,14 @@ protected LoopingNgJob(ILoggerFactory loggerFactory)
                 }
 
                 Logger.LogInformation("Running job.");
-                await RunInternal(cancellationToken);
+                try
+                {
+                    await RunInternal(cancellationToken);
+                }
+                catch(TransientException te)
```
skofman1: IMO the try catch should be done inside RunInternal, without changing LoopingNgJob

scottbommarito: Not possible--`RunInternal` is overridden by every looping job with its functionality. Additionally, the code changes apply to two jobs: Feed2Catalog and Package2Catalog.

src/Ng/Jobs/LoopingNgJob.cs
scottbommarito: I think it's somewhat confusing that the job will still log "`Job finished!`" when it fails due to a transient exception. I would include that log within the `try` and add a "`Job failed!`" log to the `catch`. On a related note, it may be worth including a "`Job failed!`" log at [Program.cs#L73](https://github.com/NuGet/NuGet.Services.Metadata/blob/master/src/Ng/Program.cs#L73) as well.

E.g.
```
try
{
    await RunInternal(cancellationToken);
    Logger.LogInformation("Job finished!");
}
catch (TransientException te)
{
    Logger.LogWarning("Transient exception: {0}. {1}", te.Message, te.InnerException.StackTrace??te.StackTrace);
    Logger.LogWarning("Job failed!");
}
```

src/Ng/Jobs/CatalogUtility.cs
cristinamanum: updated

src/Ng/Jobs/CatalogUtility.cs
cristinamanum: It is already a timeout of 5 minutes. We can increase it but what would be a good value?

src/Ng/Jobs/LoopingNgJob.cs
cristinamanum: If implemented in the LoopingNgJob all the jobs can only opt-in to use the exception. They will not need to try/catch/handle in their RunInternal.

src/Ng/Jobs/CatalogUtility.cs
```diff
@@ -79,9 +79,18 @@ public static HttpClient CreateHttpClient(bool verbose)
...
-            using (var stream = await client.GetStreamAsync(uri))
+            try
             {
-                feed = XElement.Load(stream);
+                using (var stream = await client.GetStreamAsync(uri))
+                {
+                    feed = XElement.Load(stream);
+                }
+            }
+            //if the client gets disposed because the timeout a task cancelled exception will happen
```
cristinamanum: Updated.

The above comment is a very good observation, with pros and cons that you mentioned.  In addition in the future we would probably like to integrate the exceptions with an alerting system as ICM. An  exception like this is not really actionable. Also having the transient exception concept we can implement retry patterns. In my opinion, would be good to avoid process termination if recovery is possible.

src/Ng/Jobs/CatalogUtility.cs
joelverhagen: Oh boy, 5 minutes is quite enough.

src/Ng/Jobs/LoopingNgJob.cs
cristinamanum: Good catch. Updated.

cristinamanum: Updated.

src/Ng/Jobs/CatalogUtility.cs
```diff
@@ -79,9 +79,18 @@ public static HttpClient CreateHttpClient(bool verbose)
...
+                using (var stream = await client.GetStreamAsync(uri))
+                {
+                    feed = XElement.Load(stream);
+                }
+            }
+            catch (TaskCanceledException tce)
+            {
+                // If the client gets disposed because the HTTP request timed out, a TaskCanceledException will be thrown.
+                // It is not necessary to restart the process in this case.
+                throw new TransientException("Http client was disposed.", tce);
```
dtivel: Adding resiliency for recoverable transient network conditions is the right thing to do.  However, do we know why HTTP GET requests are timing out after 5 minutes?  5 minutes is a long timeout.

Correct me if I'm wrong, but it looks like this spot is querying the OData feed for packages, while the [other place](https://github.com/NuGet/NuGet.Services.Metadata/pull/135/files#diff-368a08ab34918e5b2c6aee3991059d5eR186) where you catch a `TaskCanceledException` is querying CDN.  Rethrowing all `HttpClient` `TaskCanceledException`'s as a generic `TransientException` makes it difficult to understand the occurrence of faults in each location or distinguish a SQL timeout from an HTTP timeout; they're all transient and potentially recoverable, but you'd have to crawl exception call stacks or parse free-form exception messages to begin to perform root cause analysis.  If you can add more context at the time of failure to improve future monitoring and alerting, it will help down the road.

One option is to subclass `TransientException` with `HttpGetPackagesFailedException` and `HttpGetPackageMetadataFailedException`.  Then log the full exception (including its type).

src/Ng/Jobs/CatalogUtility.cs
```diff
@@ -171,7 +180,17 @@ private static DateTime ForceUtc(DateTime date)
...
-                    var response = await client.GetAsync(packageUri, cancellationToken);
+                    HttpResponseMessage response = null;
+                    try
+                    {
+                        response = await client.GetAsync(packageUri, cancellationToken);
+                    }
+                    catch (TaskCanceledException tce)
+                    {
+                        // If the client gets disposed because the HTTP request timed out, a TaskCanceledException will be thrown.
+                        // It is not necessary to restart the process in this case.
```
dtivel: Apply my early comment feedback here as well.

src/Ng/Jobs/LoopingNgJob.cs
```diff
@@ -38,8 +38,15 @@ protected LoopingNgJob(ILoggerFactory loggerFactory)
...
-                await RunInternal(cancellationToken);
-                Logger.LogInformation("Job finished!");
+                try
+                {
+                    await RunInternal(cancellationToken);
+                    Logger.LogInformation("Job finished!");
+                }
+                catch (TransientException te)
+                {
+                    Logger.LogWarning("Job execution ended due to a transient exception: {0}. {1}", te.Message, te.InnerException?.StackTrace??te.StackTrace);
```
scottbommarito: Something I thought about this weekend...

The way you've written this, the job will still wait the entire interval before trying again. If we know we encountered an intermittent failure, wouldn't we want to try again immediately rather than wait?

The interval of the V3 jobs is low (30 seconds) but I'm slightly concerned about waiting when we know the failure was intermittent.

src/Ng/Jobs/CatalogUtility.cs
```diff
@@ -79,9 +79,18 @@ public static HttpClient CreateHttpClient(bool verbose)
...
-                feed = XElement.Load(stream);
+                using (var stream = await client.GetStreamAsync(uri))
+                {
+                    feed = XElement.Load(stream);
+                }
+            }
+            catch (TaskCanceledException tce)
+            {
+                // If the client gets disposed because the HTTP request timed out, a TaskCanceledException will be thrown.
+                // It is not necessary to restart the process in this case.
```
cristinamanum: Updated. The observation is correct, the client is not disposed.

src/Ng/Jobs/CatalogUtility.cs
cristinamanum: The observation is correct. The client is not disposed.Updated the comments.

cristinamanum: I like the subclass idea. In this case it was httpclient request timeout. Implemented a new exception type and added more information to make easier the debug scenario.

src/Ng/Jobs/CatalogUtility.cs
cristinamanum: Updated.

src/Ng/Jobs/LoopingNgJob.cs
cristinamanum: The idea is interesting. The failure could be due to low resources on the machine, waiting could be beneficial in case like this. I would let the job with the regular time interval for now.

scottbommarito: Well, previously, when the jobs encountered these issues, the process would crash and they would restart immediately. I think it's fine that we wait, however, because with no wait there's the potential for an infinite restart loop.

PR #136, opened 2017-03-09
Sign Ng and NuGet.ApplicationInsights.Owin

joelverhagen:
Addresses https://github.com/NuGet/Engineering/issues/278.


joelverhagen:
Verified on dev by deploying the V3 jobs, publishing a package, and making sure it flows through properly.

PR #137, opened 2017-03-09
Additional Functional test Coverage

ryuyu:
Add Parsing for whole search result object and add test for nonzero download count on json.net.

@skofman1 @joelverhagen @shishirx34

### DIFF
--- a/tests/BasicSearchTests.FunctionalTests.Core/V3SearchFunctionalTests.cs
+++ b/tests/BasicSearchTests.FunctionalTests.Core/V3SearchFunctionalTests.cs
@@ -39,7 +39,7 @@ namespace BasicSearchTests.FunctionalTests.Core
         }
 
         [Fact]
-        public async Task ResultForJsonNetHasDownloads()
+        public async Task EmptyQueryResultHasDownloads()
         {
             // Act
             var response = await Client.GetAsync(new V3SearchBuilder().RequestUri);
--- a/tests/BasicSearchTests.FunctionalTests.Core/V3SearchFunctionalTests.cs
+++ b/tests/BasicSearchTests.FunctionalTests.Core/V3SearchFunctionalTests.cs
@@ -34,7 +34,7 @@ namespace BasicSearchTests.FunctionalTests.Core
 
             // Assert
             Assert.Equal(HttpStatusCode.OK, response.StatusCode);
-            Assert.True(result.TotalHits.HasValue && result.TotalHits.Value > 0, "No results found, should find atleast some results for empty string query.");
+            Assert.True(result.TotalHits.HasValue && result.TotalHits.Value > 0, "No results found, should find at least some results for empty string query.");
             Assert.NotNull(result.Data);
         }
 
@@ -42,12 +42,12 @@ namespace BasicSearchTests.FunctionalTests.Core
         public async Task ResultForJsonNetHasDownloads()
         {
             // Act
-            var response = await Client.GetAsync(new V3SearchBuilder() { Query = "JSON.Net" }.RequestUri);
+            var response = await Client.GetAsync(new V3SearchBuilder().RequestUri);
             var result = await response.Content.ReadAsAsync<V3SearchResult>();
 
             // Assert
             Assert.Equal(HttpStatusCode.OK, response.StatusCode);
-            Assert.True(result.TotalHits.HasValue && result.TotalHits.Value > 0, "No results found, should find atleast some results for empty string query.");
+            Assert.True(result.TotalHits.HasValue && result.TotalHits.Value > 0, "No results found, should find at least some results for empty string query.");
             Assert.NotNull(result.Data);
 
             var topResult = result.Data[0];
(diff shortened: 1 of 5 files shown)


tests/BasicSearchTests.FunctionalTests.Core/V3SearchFunctionalTests.cs
```diff
@@ -37,5 +37,23 @@ public class V3SearchFunctionalTests : BaseFunctionalTests
...
+        [Fact]
+        public async Task ResultForJsonNetHasDownloads()
+        {
+            // Act
+            var response = await Client.GetAsync(new V3SearchBuilder() { Query = "JSON.Net" }.RequestUri);
+            var result = await response.Content.ReadAsAsync<V3SearchResult>();
+
+            // Assert
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            Assert.True(result.TotalHits.HasValue && result.TotalHits.Value > 0, "No results found, should find atleast some results for empty string query.");
```
shishirx34: for JSON.Net query may be?

tests/BasicSearchTests.FunctionalTests.Core/V3SearchFunctionalTests.cs
```diff
@@ -37,5 +37,23 @@ public class V3SearchFunctionalTests : BaseFunctionalTests
             Assert.True(result.TotalHits.HasValue && result.TotalHits.Value > 0, "No results found, should find atleast some results for empty string query.");
             Assert.NotNull(result.Data);
         }
+
+        [Fact]
+        public async Task ResultForJsonNetHasDownloads()
+        {
+            // Act
+            var response = await Client.GetAsync(new V3SearchBuilder() { Query = "JSON.Net" }.RequestUri);
```
shishirx34: This won't necessarily check the download count for JSON.net, rather the close match of the package. In INT and DEV environment we do not have any JSON.Net package. Also, in local testing this test would fail because we won't have any data necessarily for this query. 

Honestly, I see very low value in this test. If we do want to check some data exists on a query, I would rather do an empty string query and test the download count, which again is possibly useless when running this locally(which arguably is not the idea of functional tests).

tests/BasicSearchTests.FunctionalTests.Core/V3SearchFunctionalTests.cs
scottbommarito: 1 - "atleast" -> "at least"
2 - This string (`"No results found, should find atleast some results for empty string query."`) is now used twice. We should either make it a constant or move this chain of asserts (`Assert.Equal(...); Assert.True(...); Assert.NotNull(...);`) into a function and then call it in both of these functions.

tests/BasicSearchTests.FunctionalTests.Core/V3SearchFunctionalTests.cs
scottbommarito: I agree that this should be a blank query that verifies the packages in the response don't have 0 download counts.

Not sure if all of our gallery functional tests will always succeed when we run them locally anyway...it's fine if this fails but ideally if we could write it in a way that it wouldn't that would be great.

ryuyu: Empty search should return top x results, so I'll make that change.

tests/BasicSearchTests.FunctionalTests.Core/V3SearchFunctionalTests.cs
ryuyu: String is descriptive for both cases, it doesn't matter if they are identical or not.
The assertions should NOT be moved because as they are, they are very explicit about what is being verified here. There is no guarantee that these being the same will always be the case, and looking at each test should be a self contained package of what is being tested and what conditions are being checked.

scottbommarito: I don't see how having a helper function (`AssertSearchResultHasData`) harms the structure of the test case--it's fairly explicit about what it's asserting and if we decide we want to change what each test is asserting specifically we can just split the function. Ultimately however, I'm fine with how it is now.

tests/BasicSearchTests.FunctionalTests.Core/V3SearchFunctionalTests.cs
```diff
@@ -34,8 +34,26 @@ public class V3SearchFunctionalTests : BaseFunctionalTests
 
             // Assert
             Assert.Equal(HttpStatusCode.OK, response.StatusCode);
-            Assert.True(result.TotalHits.HasValue && result.TotalHits.Value > 0, "No results found, should find atleast some results for empty string query.");
+            Assert.True(result.TotalHits.HasValue && result.TotalHits.Value > 0, "No results found, should find at least some results for empty string query.");
             Assert.NotNull(result.Data);
         }
+
+        [Fact]
+        public async Task ResultForJsonNetHasDownloads()
```
scottbommarito: To match the rest of the file this should probably be `ShouldReturnResultsWithDownloads`

ryuyu: ✔️

PR #139, opened 2017-03-16
Delete unused "pipeline" code

joelverhagen:
This code was added in late 2014 and never seemed to be used in the product.

Removing to avoid unnecessary work in future changes or refactorings.


tests/CatalogTests/CatalogTests.csproj
```diff
@@ -162,6 +162,9 @@
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </Content>
   </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
```
xavierdecoster: What's this?

joelverhagen: Apparently this is a tag for test projects (which this project is):
http://stackoverflow.com/a/22917902/52749

It was added automatically by VS.

PR #142, opened 2017-03-21
Adding support for filtering Semver2 in BasicSearch

ryuyu:
This PR does the following:
1. Add support for a new bit filter based on a new field in the index, semverLevel
2. Adds new query parameter for filtering based on said filter
3. Removes ranking and find endpoints.

Tests are not included here and will be included in a followup.

@joelverhagen @xavierdecoster @skofman1


src/NuGet.Services.BasicSearch/ServiceEndpoints.cs
```diff
@@ -116,6 +100,17 @@ private static bool GetIncludePrerelease(IOwinContext context)
             return includePrerelease;
         }
 
+        private static bool GetIncludeSemver2(IOwinContext context)
+        {
+            System.Version semVerLevel;
+            if (System.Version.TryParse(context.Request.Query["semVerLevel"], out semVerLevel))
```
scottbommarito: The values for the parameter names ("`semVerLevel`", "`luceneQuery`", etc) should be in a constant.

Also you could probably replace these methods with a generic helper function (`static T GetParameter<T>(IOwinContext context, string name)`) using [Convert.ChangeType](https://msdn.microsoft.com/en-us/library/dtb69x08(v=vs.110).aspx), but I could understand if that would be too much refactoring for this PR.

src/NuGet.Services.BasicSearch/Startup.cs
```diff
@@ -297,9 +294,6 @@ private void TrackIndexMetrics(NuGetSearcherManager searcherManager, SearchTelem
                         case "/search/query":
                             await ServiceEndpoints.V2SearchAsync(context, _searcherManager, _responseWriter);
                             break;
-                        case "/rankings":
```
scottbommarito: I feel like removing these endpoints is a little bit out of the scope of what a "SemVer2 filtering" PR should do, but I'm fine with it being in here.

src/NuGet.Indexing/Handlers/LatestListedHandler.cs
```diff
@@ -105,5 +111,11 @@ private static bool GetListed(Document document)
             string listed = document.Get("Listed");
             return (listed == null) ? false : listed.Equals("true", StringComparison.InvariantCultureIgnoreCase);
         }
+
+        private static bool GetSemver2(Document document)
+        {
+            string semverLevel = document.Get("SemverLevel");
```
joelverhagen: nit: should be SemVerLevel

src/NuGet.Indexing/Handlers/LatestListedHandler.cs
```diff
@@ -105,5 +111,11 @@ private static bool GetListed(Document document)
...
+            return (semverLevel == null) ? false : semverLevel.Equals("2.0.0", StringComparison.InvariantCultureIgnoreCase);
```
joelverhagen: ignore case is not necessary

src/NuGet.Indexing/NuGetSearcherManager.cs
```diff
@@ -240,15 +240,23 @@ protected override NuGetIndexSearcher CreateSearcher(IndexReader reader)
...
-                var h00 = new LatestListedHandler(includeUnlisted: false, includePrerelease: false);
-                var h01 = new LatestListedHandler(includeUnlisted: false, includePrerelease: true);
-                var h10 = new LatestListedHandler(includeUnlisted: true, includePrerelease: false);
-                var h11 = new LatestListedHandler(includeUnlisted: true, includePrerelease: true);
-
-                indexReaderProcessor.AddHandler(h00);
-                indexReaderProcessor.AddHandler(h01);
-                indexReaderProcessor.AddHandler(h10);
-                indexReaderProcessor.AddHandler(h11);
+                var h000 = new LatestListedHandler(includeUnlisted: false, includePrerelease: false, includeSemver2: false);
```
joelverhagen: This is a permutation generator. Can be simplified to bitmask of integers from 0 to 2^n - 1. n = 3 today. 😄

src/NuGet.Indexing/ServiceImpl.cs
```diff
@@ -17,6 +17,7 @@ public static class ServiceImpl
             string scheme,
             string q,
             bool includePrerelease,
+            bool includeSemver2,
```
joelverhagen: `semVer` instead of `semver` -- other places in this PR too

src/NuGet.Indexing/ServiceImpl.cs
```diff
@@ -90,21 +91,6 @@ public static void AutoComplete(JsonWriter jsonWriter, NuGetSearcherManager sear
             }
         }
 
-        public static void Find(JsonWriter jsonWriter, NuGetSearcherManager searcherManager, string id, string scheme)
```
joelverhagen: 👏

src/NuGet.Indexing/ServiceInfoImpl.cs
```diff
@@ -19,18 +19,5 @@ public static void Stats(JsonWriter jsonWriter, NuGetSearcherManager searcherMan
                 searcherManager.Release(searcher);
             }
         }
-
-        public static void Rankings(JsonWriter jsonWriter, NuGetSearcherManager searcherManager)
```
joelverhagen: 👏

src/NuGet.Services.BasicSearch/ServiceEndpoints.cs
```diff
@@ -116,6 +100,17 @@ private static bool GetIncludePrerelease(IOwinContext context)
...
+            {
+                return semVerLevel.CompareTo(new System.Version("2.0.0")) >= 0;
```
joelverhagen: Parse as `NuGetVersion`

src/NuGet.Indexing/Handlers/LatestListedHandler.cs
```diff
@@ -105,5 +111,11 @@ private static bool GetListed(Document document)
             string listed = document.Get("Listed");
             return (listed == null) ? false : listed.Equals("true", StringComparison.InvariantCultureIgnoreCase);
         }
+
+        private static bool GetSemver2(Document document)
```
joelverhagen: This will work with today's documents right? Meaning `document.Get("SemverLevel")` just returned null instead of throwing when this field does not exist.

joelverhagen:
We need unit and integration test coverage here.

src/NuGet.Indexing/Handlers/LatestListedHandler.cs
ryuyu: Yes

src/NuGet.Indexing/Handlers/LatestListedHandler.cs
cristinamanum: Maybe a IsSemVer2 instead of GetSemVer2? I would expect as  GetSemVer2 to return the version.

scottbommarito: Actually I think the best naming would be `GetIsSemVer2`.

`GetListed` should also be changed to `GetIsListed`.

src/NuGet.Indexing/GalleryServiceImpl.cs
```diff
@@ -10,7 +10,7 @@ namespace NuGet.Indexing
 {
     public class GalleryServiceImpl
     {
-        public static void Search(JsonWriter jsonWriter, NuGetSearcherManager searcherManager, string q, bool countOnly, bool includePrerelease, string sortBy, int skip, int take, string feed, bool ignoreFilter, bool luceneQuery)
+        public static void Search(JsonWriter jsonWriter, NuGetSearcherManager searcherManager, string q, bool countOnly, bool includePrerelease, bool includeSemver2, string sortBy, int skip, int take, string feed, bool ignoreFilter, bool luceneQuery)
```
xavierdecoster: Agree with @scottbommarito on the rename to `includeSemVer2`.
nit: align parameters vertically for readability and avoid horizontal scrolling (other places in this PR too).

src/NuGet.Indexing/NuGetIndexSearcher.cs
```diff
@@ -14,21 +14,23 @@ namespace NuGet.Indexing
...
-            Filter[][] latest,
+            Filter[][][] latest,
             IReadOnlyDictionary<string, int[]> docIdMapping,
             Downloads downloads,
             VersionResult[] versions,
             RankingResult rankings,
             QueryBoostingContext context,
             OpenBitSet latestBitSet,
             OpenBitSet latestStableBitSet,
+            OpenBitSet latestSemver2BitSet,
```
xavierdecoster: Replace `Semver` by `SemVer` throughout the code base.

src/NuGet.Indexing/NuGetSearcherManager.cs
```diff
@@ -240,15 +240,23 @@ protected override NuGetIndexSearcher CreateSearcher(IndexReader reader)
 
                 // There are four flavors of Latest/Listed filter to reflect all the possible combinations.
```
skofman1: eight :)

src/NuGet.Indexing/NuGetSearcherManager.cs
```diff
@@ -240,15 +240,23 @@ protected override NuGetIndexSearcher CreateSearcher(IndexReader reader)
...
+                var h000 = new LatestListedHandler(includeUnlisted: false, includePrerelease: false, includeSemver2: false);
+                var h010 = new LatestListedHandler(includeUnlisted: false, includePrerelease: true, includeSemver2: false);
+                var h100 = new LatestListedHandler(includeUnlisted: true, includePrerelease: false, includeSemver2: false);
+                var h110 = new LatestListedHandler(includeUnlisted: true, includePrerelease: true, includeSemver2: false);
+                var h001 = new LatestListedHandler(includeUnlisted: false, includePrerelease: false, includeSemver2: true);
+                var h011 = new LatestListedHandler(includeUnlisted: false, includePrerelease: true, includeSemver2: true);
+                var h101 = new LatestListedHandler(includeUnlisted: true, includePrerelease: false, includeSemver2: true);
+                var h111 = new LatestListedHandler(includeUnlisted: true, includePrerelease: true, includeSemver2: true);
+
+                indexReaderProcessor.AddHandler(h000);
```
skofman1: This code must be backwards compatible, otherwise we will be in a world of pain during deployment/rollback. Meaning, in case the index doesn't contain the new SemVer 2 properties, the code should fallback to default and just work. 
Is it?

src/NuGet.Indexing/Handlers/LatestListedHandler.cs
```diff
@@ -74,27 +76,31 @@ public void End(IndexReader indexReader)
...
 
-            Update(isListed, readerName, perSegmentDocumentNumber, id, version);
+            Update(isListed, isSemver2, readerName, perSegmentDocumentNumber, id, version);
         }
 
-        private void Update(bool isListed, string readerName, int n, string id, NuGetVersion version)
+        private void Update(bool isListed, bool isSemver2, string readerName, int n, string id, NuGetVersion version)
         {
-            if (isListed || _includeUnlisted)
+            if (_includeSemver2 || (!_includeSemver2 && !isSemver2))
```
ryuyu: done

src/NuGet.Indexing/GalleryServiceImpl.cs
ryuyu: Done

src/NuGet.Indexing/NuGetSearcherManager.cs
ryuyu: adjust this a bit to using a bitmask.

src/NuGet.Indexing/Handlers/LatestListedHandler.cs
ryuyu: IsSemVer2 is probably more correct. GetListed specifically refers to retrieving the field named Listed from the document. The GetSemVer2 isn't actually getting a boolean field but rather getting a test field and running a test to return a boolean value. As such, IsSemVer2 makes more sense, and i've changed it to match.

src/NuGet.Indexing/Handlers/LatestListedHandler.cs
ryuyu: done

src/NuGet.Indexing/ServiceImpl.cs
ryuyu: done

src/NuGet.Services.BasicSearch/ServiceEndpoints.cs
ryuyu: done

src/NuGet.Services.BasicSearch/ServiceEndpoints.cs
```diff
@@ -15,6 +15,7 @@ public class ServiceEndpoints
             var skip = GetSkip(context);
             var take = GetTake(context);
             var includePrerelease = GetIncludePrerelease(context);
+            var includeSemver2 = GetIncludeSemver2(context);
```
ryuyu: My understanding of the result of the discussion was that semVerLevel >=2.0.0 would be treated as all inclusive, which in the current case is equivalent to semVerLevel==2.0.0.

Was this not the case?

src/NuGet.Indexing/NuGetSearcherManager.cs
ryuyu: Yes.

tests/NuGet.IndexingTests/LatestListedHandlerTests.cs
```diff
@@ -0,0 +1,271 @@
...
+            int expectedNumDocsInResult)
+        {
+            var handler = new LatestListedHandler(includeUnlisted, includePrerelease, includeSemVer2);
+            var fakeReader = MockObjectFactory.CreateMockIndexReader(5, numberOfSubReaders: numSubReadersToFake).Object;
+
+            handler.Begin(fakeReader);
+
+            // Pick a random subreader for each document
+            var docNumber = 0;
+            var rnd = new Random();
```
joelverhagen: Please minimize randomness of tests. If you need `Random` specify a constant seed.

tests/NuGet.IndexingTests/LatestListedHandlerTests.cs
```diff
@@ -0,0 +1,271 @@
...
+            bool includePrerelease,
+            bool includeSemVer2,
+            string readerName,
+            string id,
+            Document[] documents,
+            NuGetVersion[] versions,
+            int numSubReadersToFake,
+            int expectedNumDocsInResult)
+        {
+            var handler = new LatestListedHandler(includeUnlisted, includePrerelease, includeSemVer2);
```
joelverhagen: Tests (especially longer ones) should have Arrange, Act, and Assert sections (denoted by comments).

src/NuGet.Services.BasicSearch/NuGet.Services.BasicSearch.csproj
```diff
@@ -234,6 +234,9 @@
     <Reference Include="NuGet.Services.Logging, Version=2.1.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
       <HintPath>..\..\packages\NuGet.Services.Logging.2.1.0\lib\net452\NuGet.Services.Logging.dll</HintPath>
     </Reference>
+    <Reference Include="NuGet.Versioning, Version=3.5.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
```
xavierdecoster: Why not adopt v4.0.0 at once?

src/NuGet.Services.BasicSearch/ServiceEndpoints.cs
skofman1: The issue is not the meaning of the parameter, but the fact that it is bool. IMO the code at this level should be more general, and keep the SemVer value as a NuGetVersion.

src/NuGet.Services.BasicSearch/ServiceEndpoints.cs
```diff
@@ -101,15 +101,15 @@ private static bool GetIncludePrerelease(IOwinContext context)
...
         {
             NuGetVersion semVerLevel;
             if (NuGetVersion.TryParse(context.Request.Query["semVerLevel"], out semVerLevel))
             {
-                return semVerLevel.CompareTo(new NuGetVersion("2.0.0")) >= 0;
+                return semVerLevel;
             }
 
-            return false;
+            return new NuGetVersion("1.0.0");
```
joelverhagen: cache this `NuGetVersion`

PR #146, opened 2017-03-28
Catalog2Registration: multiple catalog commits for a single package should be processed by commit time

skofman1:
Fix for: https://github.com/NuGet/NuGetGallery/issues/3716


src/Catalog/Registration/RegistrationCollector.cs
```diff
@@ -57,7 +57,7 @@ protected override Task<IEnumerable<CatalogItemBatch>> CreateBatches(IEnumerable
                 .GroupBy(item => GetKey(item.Value))
                 .Select(group => new CatalogItemBatch(
                     group.Min(item => item.CommitTimeStamp),
-                    group));
```
scottbommarito: Add the reason why we are ordering the batch in the gigantic comment above.

E.g.
```
// Order the batch so that catalog items are processed in the order that they occurred.
// Otherwise, several simultaneous changes can result in them being processed in the wrong order.
```

src/Catalog/Registration/RegistrationCollector.cs
```diff
@@ -57,7 +57,7 @@ protected override Task<IEnumerable<CatalogItemBatch>> CreateBatches(IEnumerable
...
+                    group.OrderBy(x => x.CommitTimeStamp)));
```
joelverhagen: I'm trying to understand why we do the chronological sort here instead of in `SortingCollector`. The name of `SortingCollector` suggests to me that it would be a better place.

tests/NgTests/Data/TestCatalogEntries.resx
```diff
@@ -172,6 +172,247 @@
...
+    <value>{
+  "@id": "http://tempuri.org/page0.json",
+  "@type": "CatalogPage",
+  "commitId": "8a9e7694-73d4-4775-9b7a-20aa59b9773e",
+  "commitTimeStamp": "2017-02-08T17:18:55.3335317Z",
+  "count": 3,
+  "items": [
+    {
+      "@id": "http://tempuri.org/data/2017.02.08.17.16.18/mypackage.3.0.0.json",
+      "@type": "nuget:PackageDetails",
```
joelverhagen: nit: can we put this in non-chronological order? Really we should test all of the order variants (ABC - asc, CBA - desc, BAC - unordered) to make sure catalog items that are coincidentally in order do not hide a bug.

tests/NgTests/Data/TestCatalogEntries.resx
```diff
@@ -172,6 +172,247 @@
...
+		},
+		"catalog:commitTimeStamp": {
+			"@type": "xsd:dateTime"
+		}
+	}
+}</value>
+  </data>
+  <data name="TestCatalogStorageMyPackageListed" xml:space="preserve">
+    <value>{
+"@id": "http://tempuri.org/data/2017.02.08.17.16.18/mypackage.3.0.0.json",
```
joelverhagen: nit this should be indented too

src/Catalog/Registration/RegistrationCollector.cs
joelverhagen: yes and yes

tests/NgTests/RegistrationCollectorTests.cs
```diff
@@ -18,42 +18,51 @@ namespace NgTests
...
-
-            var mockServer = new MockServerHttpClientHandler();
+        private MemoryStorage _catalogToRegistrationStorage;
+        private TestStorageFactory _catalogToRegistrationStorageFactory;
+        private MockServerHttpClientHandler _mockServer;
+        private RegistrationCollector _target;
 
-            mockServer.SetAction("/", request => Task.FromResult(new HttpResponseMessage(HttpStatusCode.OK)));
-            await mockServer.AddStorage(catalogStorage);
+        void SharedInit()
```
joelverhagen: qualify access level e.g. `private`

src/Catalog/Registration/RegistrationCollector.cs
skofman1: Sort was moved to CatalogItemBatch

joelverhagen:
Per offline conversation, could we have an item tracking the "hack" you pointed out? This should be cleaned up eventually (maybe cleaning up that inheritance tree at the same time).

joelverhagen:
Per offline conversation, the `CatalogItemBatch` is not a hack. Filed an issue for the confusing catalog collectors: https://github.com/NuGet/NuGetGallery/issues/3720

tests/NgTests/Data/TestCatalogEntries.resx
skofman1: done

PR #145, opened 2017-03-24
Update feed2catalog and catalog2registration to properly handle SemVer 2.0.0

joelverhagen:
1. Write full version to catalog entry JSON
1. Use normalized version for catalog URLs
1. Use normalized version for downloading packages from the gallery
1. Use normalized version for registration URLs and min/max ranges
1. Write full version to registration entry JSON
1. Write SemVer 2.0.0 packages to a new registration container

Addresses https://github.com/NuGet/NuGetGallery/issues/3560 and https://github.com/NuGet/NuGetGallery/issues/3680.

PR #149, opened 2017-03-29
Update lightning to handle the new SemVer 2.0.0 container

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/3682.

1. Update lightning to handle the new SemVer 2.0.0 container
1. Improve how lightning reads the catalog. Previously, it ate over 2 GB memory. Now it eats 1.4 GB.

This PR depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/145.

I am currently running it against all of prod's catalog as additional verification.


src/Catalog/CatalogIndexReader.cs
```diff
@@ -86,37 +43,58 @@ public JObject GetContext()
...
-                    var entry = new CatalogIndexEntry(new Uri(item["@id"].ToString()),
-                            item["@type"].ToString(),
-                            item["commitId"].ToString(),
-                            DateTime.Parse(item["commitTimeStamp"].ToString()),
-                            item["nuget:id"].ToString(),
-                            NuGetVersion.Parse(item["nuget:version"].ToString()));
+                    // This string is unique.
+                    var id = item["@id"].ToString();
+
+                    // This string is discarded quickly, since DateTime.Parse does not hold a reference.
```
scottbommarito: Note that `DateTime`s are structs which are pass by value, so even if it did hold a reference, it would be cloned every time anyway.

src/Catalog/StringInterner.cs
```diff
@@ -0,0 +1,45 @@
...
+                if (_instances.TryGetValue(value, out output))
+                {
+                    return output;
+                }
+
+                _lock.EnterWriteLock();
+                try
+                {
+                    _instances.Add(value, value);
+                    output = value;
```
scottbommarito: nit: I think you could put `return value;` here and remove `return output;` at the bottom

src/Catalog/CatalogIndexReader.cs
joelverhagen: It wouldn't clone the string :)

This comment is confusing though. Previously I had `item["commitTimeStamp"].ToObject<string>()`.

src/Catalog/StringInterner.cs
joelverhagen: Yeah, nice one.

src/Ng/Jobs/LightningJob.cs
```diff
@@ -146,20 +172,14 @@ protected override void Init(IDictionary<string, string> arguments, Cancellation
...
-            _storageContainer = arguments.GetOrThrow<string>(Arguments.StorageContainer);
-            _storageBaseAddress = arguments.GetOrThrow<string>(Arguments.StorageBaseAddress);
-            _compress = arguments.GetOrDefault(Arguments.Compress, false);
         }
 
         private void InitPrepare(IDictionary<string, string> arguments)
         {
             _outputFolder = arguments.GetOrThrow<string>(Arguments.OutputFolder);
-            _catalogIndex = arguments.GetOrThrow<string>(Arguments.CatalogIndex);
+            _catalogIndex = arguments.GetOrThrow<string>(Arguments.Source);
```
scottbommarito: IMO all the jobs should use `CatalogIndex` instead of `Source`, because it's confusing what `Source` refers to unless you're already familiar with the jobs, but I like aligning this with the other jobs.

src/Ng/Jobs/LightningJob.cs
```diff
@@ -131,10 +143,24 @@ protected override void Init(IDictionary<string, string> arguments, Cancellation
         {
             PrintLightning();
 
+            // Hard code against Azure storage.
+            arguments[Arguments.StorageType] = Arguments.AzureStorageType;
```
scottbommarito: What if I want to write registration blobs to my local file system?

src/Ng/RegistrationStorageFactories.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using NuGet.Services.Metadata.Catalog.Persistence;
+
+namespace Ng
+{
+    public class RegistrationStorageFactories
```
scottbommarito: XML comments in this file?

I know you have them in `Catalog2Registration`'s `Init`, but I think they belong here now that you've shared the storage factory creation logic.

```
/// <summary>
/// Contains the storage factories used to create registration blobs.
/// </summary>
public class RegistrationStorageFactories
```

```
/// <summary>
/// Supports legacy clients which cannot use the current registration blobs.
/// Maintains the cursor of the job.
/// Contains an aggregate storage factory that writes two sets of registration blobs.
/// 1 - not gzipped without SemVer 2.0.0 packages
/// 2 - gzipped without SemVer 2.0.0 packages
/// </summary>
public StorageFactory LegacyStorageFactory { get; }

/// <summary>
/// Writes registration blobs that are gzipped and contain SemVer 2.0.0 packages
/// </summary>
public StorageFactory SemVer2StorageFactory { get; }
```

src/Ng/lightning-template.txt
```diff
@@ -2,15 +2,15 @@
...
 
-title [cursor]
+title [cursorFile]
 
 @echo starting
 
-start /w ng.exe lightning -command strike -indexFile [index] -cursorFile [cursor] -contentBaseAddress [contentbaseaddress] -storageAccount [storageaccount] -storageContainer [storagecontainer] -storageBaseAddress [storagebaseaddress] -compress [compress] -verbose true
+start /w ng.exe lightning -command strike -indexFile "[indexFile]" -cursorFile "[cursorFile]" -contentBaseAddress "[contentBaseAddress]" -verbose "[verbose]" -storageBaseAddress "[storageBaseAddress]" -storageAccountName "[storageAccountName]" -storageKeyValue "[storageKeyValue]" -storageContainer "[storageContainer]" -useCompressedStorage "[useCompressedStorage]" -compressedStorageBaseAddress "[compressedStorageBaseAddress]" -compressedStorageAccountName "[compressedStorageAccountName]" -compressedStorageKeyValue "[compressedStorageKeyValue]" -compressedStorageContainer "[compressedStorageContainer]" -useSemVer2Storage "[useSemVer2Storage]" -semVer2StorageBaseAddress "[semVer2StorageBaseAddress]" -semVer2StorageAccountName "[semVer2StorageAccountName]" -semVer2StorageKeyValue "[semVer2StorageKeyValue]" -semVer2StorageContainer "[semVer2StorageContainer]"
```
scottbommarito: It seems a little dangerous to me that whenever we change the parameters of the job, the person needs to know to update this script as well. It's probably outside the scope of this PR, but it'd be nice if the job just copied over all of the commands that aren't `prepare` specific.

tests/CatalogTests/CatalogIndexReaderTests.cs
```diff
@@ -0,0 +1,81 @@
...
+                new CollectorHttpClient(new InMemoryHttpHandler(responses)));
+
+            // Act
+            var entries = await reader.GetEntries();
+
+            // Assert
+            var entryList = entries.ToList();
+            Assert.Equal(3, entryList.Count);
+
+            Assert.Equal("http://tempuri.org/data/2015.10.12.10.08.55/listedpackage.1.0.1.json", entryList[0].Uri.ToString());
```
scottbommarito: Add verification of all the fields to this test. You're missing `Types`, `CommitId`, and `Version`.

tests/CatalogTests/CatalogIndexReaderTests.cs
```diff
@@ -0,0 +1,81 @@
...
+            Assert.Equal("http://tempuri.org/data/2015.10.12.10.08.54/listedpackage.1.0.0.json", entryList[1].Uri.ToString());
+            Assert.Equal("http://tempuri.org/data/2015.10.12.10.08.54/unlistedpackage.1.0.0.json", entryList[2].Uri.ToString());
+
+            Assert.Equal("2015-10-12T10:08:55.3335317", entryList[0].CommitTimeStamp.ToString("O"));
+            Assert.Equal("2015-10-12T10:08:54.1506742", entryList[1].CommitTimeStamp.ToString("O"));
+            Assert.Equal("2015-10-12T10:08:54.1506742", entryList[2].CommitTimeStamp.ToString("O"));
+
+            Assert.Same(entryList[0].Id, entryList[1].Id);
```
scottbommarito: Check the value directly as well? If `entryList[0].Id` and `entryList[1].Id` are both null or some equally wrong value this won't catch it.
```
Assert.Equal("listedpackage", entryList[0].Id);
Assert.Equal("listedpackage", entryList[1].Id);
Assert.Equal("unlistedpackage", entryList[2].Id);
```

tests/NgTests/StorageFactoryTests.cs
```diff
@@ -118,5 +120,144 @@ public class StorageFactoryTests
...
+        public void CreateRegistrationStorageFactories_WithAllThreeFactories()
+        {
+            // Arrange
+            var arguments = new Dictionary<string, string>()
+            {
+                { Arguments.StorageType, "azure" },
+
+                { Arguments.StorageBaseAddress, "http://localhost/reg" },
+                { Arguments.StorageAccountName, "testAccount" },
+                { Arguments.StorageKeyValue, "AiWrbTMNTApZxLeWB7a0yN52gJj+ZlGE0ipRi9PaTcn9AU4epwvsngE5rLSMk9TwpazxUtzeyBnFeWFAhfkGpw==" },
```
scottbommarito: Where did you get this key from lol?

src/Ng/Jobs/LightningJob.cs
joelverhagen: Not supported.

src/Ng/Jobs/LightningJob.cs
joelverhagen: This is a discussion of semantics.

I agree that `CatalogIndex` is a little clearer but the benefit here is small. The cost is changing all of `catalog2*` jobs. I don't think that's worth it.

tests/NgTests/StorageFactoryTests.cs
joelverhagen: Copied from @cristinamanum's tests.

src/Ng/Jobs/LightningJob.cs
joelverhagen: Additionally, this is a rather high risk change. It's easy to get configuration wrong.

src/Ng/lightning-template.txt
joelverhagen: Nice idea, tracked here:
https://github.com/NuGet/NuGetGallery/issues/3730

tests/CatalogTests/CatalogIndexReaderTests.cs
joelverhagen: sure, good idea. Not sure how I came up with these two properties already tested 😄

tests/CatalogTests/CatalogIndexReaderTests.cs
joelverhagen: Fixed.

src/Ng/RegistrationStorageFactories.cs
joelverhagen: Fixed. Was a little more terse.

joelverhagen:
Anyone else wanna take a 👀 at this?

src/Ng/CommandHelpers.cs
```diff
@@ -75,6 +75,39 @@ private static ISecretInjector GetSecretInjector(IDictionary<string, string> arg
             return new SecretInjector(secretReader);
         }
 
+        public static RegistrationStorageFactories CreateRegistrationStorageFactories(IDictionary<string, string> arguments, bool verbose)
+        {
+            StorageFactory legacyStorageFactory;
```
skofman1: legacyStorageFactoy is not the best name here... perhaps SemVer1StorageFactory?

xavierdecoster: @skofman1 see https://github.com/NuGet/NuGet.Services.Metadata/pull/145#discussion_r108620265

PR #150, opened 2017-03-30
Careful cursor

joelverhagen:
Only write the cursor timestamp if it changes from the previous.

Suppose you have a page with:
```
ID   Version   Commit Timestamp
--   -------   ----------------
B    1.0.0     10:00 PM
A    1.0.0     10:00 PM
A    1.0.1     10:05 PM
```

Also suppose you have the cursor set to 9:55 PM. This means all of the items in the page need to be processed. You will have two batches processed by `RegistrationCollector`:

```
Batch 1: B 1.0.0
 - Commit Timestamp: 10:00 PM

Batch 2: A 1.0.0, A 1.0.1
 - Commit Timestamp: 10:00 PM
```

Suppose batch 1 is processed successfully. The [`CommitCollector` will update the cursor](https://github.com/NuGet/NuGet.Services.Metadata/blob/cb1e7afa4d345a8a6d46ff3edd874bd8a7579e8f/src/Catalog/CommitCollector.cs#L60) to 10:00 PM (the commit timestamp of batch 1) then start processing batch 2.

Now suppose the processing of batch 2 fails such that neither A 1.0.0 nor A 1.0.1 are completed.

The job will crash due to the failure then start up again. Since the cursor is now at 10:00 PM, the set of items in the page considered will be:

```
ID   Version   Commit Timestamp
--   -------   ----------------
A    1.0.1     10:05 PM
```

This means A 1.0.0 is skipped. Remember that the cursor means "everything strictly after this timestamp needs to be processed".


joelverhagen:
Addresses https://github.com/NuGet/NuGetGallery/issues/3733.

src/Catalog/CommitCollector.cs
```diff
@@ -56,10 +57,25 @@ public CommitCollector(Uri index, Func<HttpMessageHandler> handlerFunc = null)
                         batch.CommitTimeStamp == lastBatch.CommitTimeStamp, 
                         cancellationToken);
 
-                    front.Value = batch.CommitTimeStamp;
-                    await front.Save(cancellationToken);
+                    // Only commit if this is the last batch that is going to be processed or if the commit timestamp
```
skofman1: I'm concered about putting this fix in the collectors base class.
It's only relevant for RegistrationCollector. Since the code base is not well UTed, and we dont have monitoring in place yet, this change introduces a risk. 
Lets discuss F2F.

src/Catalog/CommitCollector.cs
```diff
@@ -56,10 +57,25 @@ public CommitCollector(Uri index, Func<HttpMessageHandler> handlerFunc = null)
...
+                    // has changed from the previous batch. This is important because if two batches have the same
+                    // commit timestamp but processing the second fails, we should not progress the cursor forward.
+                    if (ReferenceEquals(batch, lastBatch))
+                    {
+                        front.Value = batch.CommitTimeStamp;
+                        await front.Save(cancellationToken);
+                        Trace.TraceInformation("CommitCatalog.Fetch front.Value saved due to last batch: {0}", front);
+                    }
+                    else if (previousCommitTimeStamp.HasValue && previousCommitTimeStamp != batch.CommitTimeStamp)
```
dtivel: Suppose the following:

ID  | Version  | Commit Timestamp
--  | -------  | ----------------
A   | 1.0.0    | 10:00 PM
B   | 1.0.0    | 10:05 PM
C   | 1.0.0    | 10:10 PM

There will be 3 batches.

With this logic the cursor won't be updated upon successful completion of batch 1 (A 1.0.0).

Shouldn't it be the following?

    else if (!previousCommitTimeStamp.HasValue || previousCommitTimeStamp != batch.CommitTimeStamp)

src/Catalog/CommitCollector.cs
joelverhagen: Sure, seem like this could be a good opportunity to test cursors on all jobs running in prod 👍

tests/NgTests/RegistrationCollectorTests.cs
```diff
@@ -40,6 +41,68 @@ void SharedInit()
...
+        private void FailFirstRequest(string relativeUri)
+        {
+            var originalAction = _mockServer.Actions[relativeUri];
+            var hasFailed = false;
+            Func<HttpRequestMessage, Task<HttpResponseMessage>> failFirst = request =>
+            {
+                if (!hasFailed)
+                {
+                    hasFailed = true;
+                    throw new HttpRequestException("The tubes are clogged!");
```
dtivel: Use a less humorous and more factual message.  Example:  "Simulated failure."

src/Catalog/CommitCollector.cs
joelverhagen: This won't work either. This will make you always commit after the first batch. Suppose:

ID  | Version  | Commit Timestamp
--  | -------  | ----------------
A   | 1.0.0    | 10:00 PM
B   | 1.0.0    | 10:00 PM
C   | 1.0.0    | 10:05 PM

If you fail on the second batch, you will miss it upon re-entry since the cursor has moved to 10:00 PM.

I think the way the code is written (committing after the batch as complete but having no reference to the next batch) makes it hard to get it right. 

Really, we want to write the cursor _between_ batches. So if your timestamp goes from 10:00 PM to 10:05 PM, you should write the cursor of 10:00 PM before the first 10:05 PM batch is even touched.

dtivel: Is it safe to reprocess a successfully processed batch?

Since timestamp alone is inadequate precision for the cursor, what should the cursor be based on?

joelverhagen: > Is it safe to reprocess a successfully processed batch?

Yes. Jobs always need to be idempotent. Same logic as a distributed queue message (a la Azure Storage Queue, etc).

> Since timestamp alone is inadequate precision for the cursor, what should the cursor be based on?

Why do you think it is not adequate?

dtivel: > Yes. Jobs always need to be idempotent. Same logic as a distributed queue message (a la Azure Storage Queue, etc).

I'm more concerned with how jobs now than what they need to be.

> Why do you think it is not adequate?

In your last example a timestamp is insufficient to distinguish between completion of A or completion of A and B.

joelverhagen: > I'm more concerned with how jobs are now than what they need to be.

They are all designed to be idempotent and have demonstrated themselves to since their implementation. That being said, there are bugs, such as this one. And no doubt others that lead to all of the V3 inconsistencies. 

> In your last example a timestamp is insufficient to distinguish between completion of A or completion of A and B.

Correct. If there is an error in B, we should reprocess A. This happens all the time today in all of our jobs. When we are processing a single batch and a transient blob storage error occurs, we reprocess the whole commit.

Modifying the cursor to have more granularity (such as tracking commit timestamp as well as IDs within that commit timestamp that have been completed) is possible but won't help much in the 99% case where jobs just plod along happily without any transient errors.

PR #152, opened 2017-04-03
Use UTC time when determining the listed property in registration

joelverhagen:
Addresses https://github.com/NuGet/NuGetGallery/issues/3740.


tests/NgTests/Data/TestCatalogEntries.resx
```diff
@@ -879,11 +879,11 @@
   "lastEdited": "2015-01-01T00:00:00Z",
   "licenseNames": "",
   "licenseReportUrl": "",
-  "listed": true,
+  "listed": false,
   "packageHash": "0LUKl3HqBsrLxjTaA+J6vaa+XB5zWb5hTJb1Ht74oIzPsBJmZalApZAIdueMgzbsUxG74jjfnDZWoxUrs2A3dg==",
   "packageHashAlgorithm": "SHA512",
   "packageSize": 2540,
-  "published": "1900-01-01T01:00:00+01:00",
+  "published": "1900-01-01T01:00:00+00:00",
```
scottbommarito: Don't we want to include a test case with a `DateTime` with a negative offset to verify this issue?

src/Catalog/Registration/RegistrationMakerCatalogItem.cs
```diff
@@ -106,7 +106,7 @@ DateTime GetPublishedDate()
                 }
             }
 
-            _listed = (_publishedDate.Year == 1900) ? false : true;
+            _listed = (_publishedDate.ToUniversalTime().Year == 1900) ? false : true;
```
scottbommarito: Would a safer check be `_publishedDate.ToUniversalTime().Year <= 1900`?

We should add some logging if the published value is not what we expect--`(,1900)` or `(1900,2010]`.

tests/NgTests/Data/TestCatalogEntries.resx
joelverhagen: The negative offset only ever occurs in memory, but I will add this.

src/Catalog/Registration/RegistrationMakerCatalogItem.cs
joelverhagen: > Would a safer check be `_publishedDate.ToUniversalTime().Year <= 1900`?

This does not match the checks in other places in the code. The logic always has to do with matching against year 1900.

joelverhagen: Fixed.

tests/NgTests/Data/TestCatalogEntries.resx
joelverhagen: Fixed.

PR #153, opened 2017-04-05
Pass normalized version when creating flat container

ryuyu:
This is to strip metadata from versions before we write to the index.json.
Versions were already being normalized before we were writing the nuspec/nupkg, so no changes are necessary here.

@joelverhagen @skofman1 @xavierdecoster

### DIFF
--- a/tests/NgTests/DnxCatalogCollectorTests.cs
+++ b/tests/NgTests/DnxCatalogCollectorTests.cs
@@ -15,6 +15,7 @@ using NuGet.Services.Metadata.Catalog;
 using NuGet.Services.Metadata.Catalog.Dnx;
 using NuGet.Versioning;
 using Xunit;
+using Newtonsoft.Json.Linq;
 
 namespace NgTests
 {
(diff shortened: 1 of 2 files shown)


tests/NgTests/DnxCatalogCollectorTests.cs
```diff
@@ -175,19 +176,18 @@ private void FailFirstRequest(string relativeUri)
 
         [Theory]
         [InlineData("1.2.0")]
-        [InlineData("1.2")]
         [InlineData("0.1.2")]
-        [InlineData("1.2.3.0")]
         [InlineData("1.2.3.4")]
         [InlineData("1.2.3-beta1")]
+        [InlineData("1.2.3-beta.1")]
         [Description("Test the dnxmarker save and delete scenarios.")]
```
xavierdecoster: nit: Description still contains typo `dnxma*r*ker`.

tests/NgTests/DnxCatalogCollectorTests.cs
```diff
@@ -196,19 +196,72 @@ private void FailFirstRequest(string relativeUri)
...
+            await maker.DeletePackage(id, version, CancellationToken.None);
+            //Assert
+            Assert.Equal(catalogToDnxStorage.Content.Count, 0);
+        }
+
+        [Theory]
+        [InlineData("1.2")]
+        [InlineData("1.2.3.0")]
+        [InlineData("1.02.3")]
+        [Description("Test the dnxmarker save and delete scenarios.")]
```
xavierdecoster: typo

PR #154, opened 2017-04-05
When SemVer 2.0.0 registration is disabled, put SemVer 2.0.0 packages in legacy registration

joelverhagen:
This is only necessary because of https://github.com/NuGet/NuGetGallery/issues/3737. If one of these packages is edited (metadata edit or unlist), we should not ignore the catalog entry's effect on the existing registration hives (original and gzipped).


tests/NgTests/RegistrationCollectorTests.cs
```diff
@@ -306,11 +306,44 @@ public static IEnumerable<object[]> PageContent
             await _target.Run(front, back, CancellationToken.None);
 
             // Assert
-            // Verify the contents of the SemVer 1.0.0 storage
+            var legacyCursor = _legacyStorage.Content.FirstOrDefault(pair => pair.Key.PathAndQuery.EndsWith("cursor.json"));
+            Assert.NotNull(legacyCursor.Key);
+            var legacyIndex = _legacyStorage.Content.FirstOrDefault(pair => pair.Key.PathAndQuery.EndsWith("/testpackage.semver2/index.json"));
```
scottbommarito: check for existence of the leaf node as well just in case?

joelverhagen: Yup, good idea. This is already implied by the cursor `NotNull` assertion and the count equaling 1 assertion. But it's best to be explicit here.

joelverhagen: Fixed.

PR #161, opened 2017-04-13
Parallelize creation of legacy and SemVer 2.0.0 registration hives

joelverhagen:
Currently the gzipped and non-gzipped registration hives are written to storage in parallel. This change makes it so the SemVer 2.0.0 hive is also written in parallel.

The RDF `Graph` type is thread safe when used in a read-only way. To be defensive, I have wrapped the `Graph` used in `RegistrationCollector` so that it cannot be mutated by introducing a `ReadOnlyGraph`.

To limit impact on other components, this read-only graph is only used when the `SortingGraphCollector` loads the catalog page. Today, this is only used by `RegistrationCollector`.

More details here:
http://www.dotnetrdf.org/api/html/T_VDS_RDF_Graph.htm


src/Catalog/Helpers/Utils.cs
```diff
@@ -244,12 +245,22 @@ public static IGraph CreateGraph(string json)
...
+            return CreateGraph(compacted, readOnly: false);
+        }
+
+        public static IGraph CreateGraph(JToken compacted, bool readOnly)
+        {
             JToken flattened = JsonLdProcessor.Flatten(compacted, new JsonLdOptions());
 
             IRdfReader rdfReader = new JsonLdReader();
-            IGraph graph = new Graph();
+            Graph graph = new Graph();
```
scottbommarito: nitpick: I think this should stay `IGraph`, as it's the interface rather than the implementation, and this method returns an `IGraph`

src/Catalog/ReadOnlyGraph.cs
```diff
@@ -0,0 +1,69 @@
...
+namespace NuGet.Services.Metadata.Catalog
+{
+    /// <summary>
+    /// An immutable graph supporting multi-threaded read-only access. According to dotNetRDF documentation, the
+    /// <see cref="Graph"/> class is thread-safe when used in a read-only manner.
+    /// <see cref="http://www.dotnetrdf.org/api/html/T_VDS_RDF_Graph.htm"/>
+    /// </summary>
+    public class ReadOnlyGraph : Graph
+    {
+        private readonly bool _isReadOnly;
```
scottbommarito: Isn't this always true?

tests/CatalogTests/ReadOnlyGraphTests.cs
```diff
@@ -0,0 +1,211 @@
...
+            var mutableGraph = GetIdMutableGraph();
+            var readOnlyGraph = new ReadOnlyGraph(mutableGraph);
+
+            // Act
+            mutableGraph.Assert(
+                mutableGraph.CreateUriNode(PackageUri),
+                mutableGraph.CreateUriNode(Schema.Predicates.Version),
+                mutableGraph.CreateLiteralNode(PackageVersion));
+
+            // Act
```
scottbommarito: Another `// Act`?

src/Catalog/ReadOnlyGraph.cs
joelverhagen: Nope, not before or during the `Merge` call in the constructor.

tests/CatalogTests/ReadOnlyGraphTests.cs
joelverhagen: Should be `// Assert`

src/Catalog/Helpers/Utils.cs
joelverhagen: Sure, I changed this to be the closest type both implemented by `Graph` and `ReadOnlyGraph` but it doesn't matter.

tests/CatalogTests/ReadOnlyGraphTests.cs
joelverhagen: Fixed.

src/Catalog/Helpers/Utils.cs
joelverhagen: Fixed.

src/Catalog/ReadOnlyGraph.cs
scottbommarito: Oh I see--the `Graph` constructor must call these methods internally, so if we override them then it fails initialization. Interesting...

PR #162, opened 2017-04-20
Adding SemVer2 awareness to search result registration values.

ryuyu:
Updates search to be able to return different registration containers in the result based on if we are responding to a SemVer2 compatible search.
Updates default registration url to point to registration1

@joelverhagen @skofman1 @xavierdecoster

### DIFF
--- a/src/NuGet.Indexing/NuGetSearcherManager.cs
+++ b/src/NuGet.Indexing/NuGetSearcherManager.cs
@@ -26,7 +26,7 @@ namespace NuGet.Indexing
         public virtual DateTime? LastIndexReloadTime { get; private set; } = null;
         public virtual DateTime? LastAuxiliaryDataLoadTime { get; private set; } = null;
         public virtual string MachineName => Environment.MachineName;
-        public RegistrationAddresses RegistrationBaseAddresses { get; }
+        public RegistrationBaseAddresses RegistrationBaseAddresses { get; }
 
         private readonly FrameworkLogger _logger;
         private readonly IIndexDirectoryProvider _indexProvider;
@@ -53,7 +53,7 @@ namespace NuGet.Indexing
 
             _logger = logger;
 
-            RegistrationBaseAddresses = new RegistrationAddresses();
+            RegistrationBaseAddresses = new RegistrationBaseAddresses();
 
             _indexProvider = indexProvider;
             _loader = loader;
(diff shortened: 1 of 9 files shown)


src/NuGet.Indexing/NuGetSearcherManager.cs
```diff
@@ -122,8 +122,12 @@ public class NuGetSearcherManager : SearcherManager<NuGetIndexSearcher>
             var searcherManager = new NuGetSearcherManager(logger, indexProvider, loader, config.AuxiliaryDataRefreshRateSec, config.IndexReloadRateSec);
 
             var registrationBaseAddress = config.RegistrationBaseAddress;
-            searcherManager.RegistrationBaseAddress["http"] = MakeRegistrationBaseAddress("http", registrationBaseAddress);
-            searcherManager.RegistrationBaseAddress["https"] = MakeRegistrationBaseAddress("https", registrationBaseAddress);
+            var semVer2RegistrationBaseAddress = config.SemVer2RegistrationBaseAddress;
+
+            searcherManager.RegistrationBaseAddresses.HttpRegistrationAddress = MakeRegistrationBaseAddress("http", registrationBaseAddress);
```
scottbommarito: nit: I think this logic would make more sense inside the constructor for `RegistrationAddresses`.

src/NuGet.Indexing/RegistrationAddresses.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace NuGet.Indexing
+{
+    public class RegistrationAddresses
```
scottbommarito: All of these should be `BaseAddresses`, not just `Addresses`. It's fairly confusing for the terminology to jump back and forth.

src/NuGet.Indexing/ResponseFormatter.cs
```diff
@@ -26,7 +26,25 @@ public static class ResponseFormatter
...
+            switch (scheme)
+            {
+                case "http":
+                    baseAddress = (useSemVer2Registration ? registrationAddresses.SemVer2HttpRegistrationAddress : registrationAddresses.HttpRegistrationAddress);
+                    break;
+                case "https":
+                    baseAddress = (useSemVer2Registration ? registrationAddresses.SemVer2HttpsRegistrationAddress : registrationAddresses.HttpsRegistrationAddress);
+                    break;
+                case "test":
+                    // this case is explicitly for test purposes. When ResponseFormatterTests calls into this, it calls with "test" scheme, which we will then return the default value.
```
scottbommarito: It may be clearer if we have this explicitly set `baseAddress` to `Constants.BaseUrl`.

PR #163, opened 2017-04-21
Normalize version when applying deletes

joelverhagen:
Fix https://github.com/NuGet/Engineering/issues/399


src/Catalog/Registration/RegistrationKey.cs
```diff
@@ -34,7 +34,7 @@ public override bool Equals(object obj)
                 return false;
             }
 
-            return (Id == rhs.Id); 
+            return StringComparer.OrdinalIgnoreCase.Equals(Id, rhs.Id);
```
skofman1: we need tests for this class that make sure that if 2 objects are equal with Equals, GetHashCode also equals

src/Catalog/Registration/RegistrationEntryKey.cs
```diff
@@ -21,7 +26,7 @@ public override string ToString()
 
         public override int GetHashCode()
         {
-            return ToString().GetHashCode();
+            return $"{RegistrationKey}/{_normalizedVersion}".GetHashCode();
```
skofman1: same here

src/Catalog/Registration/RegistrationEntryKey.cs
```diff
@@ -1,18 +1,23 @@
...
 {
     public class RegistrationEntryKey
     {
+        private readonly string _normalizedVersion;
+
         public RegistrationEntryKey(RegistrationKey registrationKey, string version)
         {
             RegistrationKey = registrationKey;
             Version = version;
+            _normalizedVersion = NuGetVersionUtility.NormalizeVersion(version).ToLowerInvariant();
```
skofman1: I wonder if ToLowerInvariant should go into a more common location, so all the code converts a version  to string in the same way

joelverhagen: In general, we've had to make these sort of changes in a localized way. For example, JSON blobs should always contain the original casing but URLs should always contain the normalized + lower-cased version.

src/Catalog/Registration/RegistrationEntryKey.cs
joelverhagen: Fixed.

src/Catalog/Registration/RegistrationKey.cs
joelverhagen: Fixed.

PR #164, opened 2017-04-25
V3bcdr

cristinamanum:
Add a secondary catalog2dnx.


src/Ng/Scripts/Catalog2DnxV3Secondary.cmd
```diff
@@ -0,0 +1,14 @@
+@echo OFF
+	
+cd Ng
+
+:Top
+	echo "Starting job - #{Jobs.ngcatalog2dnx.Title.Secondary}"
+
+	title #{Jobs.ngcatalog2dnx.Title.Secondary}
```
shishirx34: Do you need to add these values to the Octopus variable set too or all of these already in there? I would make a note somewhere so that these values aren't missed if so during deployments.

cristinamanum: I added them in octopus.

src/Ng/Scripts/Catalog2DnxV3Secondary.cmd
```diff
@@ -0,0 +1,14 @@
+@echo OFF
+	
+cd Ng
+
+:Top
+	echo "Starting job - #{Jobs.ngcatalog2dnx.Title.Secondary}"
```
cristinamanum: Thank you Scott - yes, i saw that.

PR #166, opened 2017-04-26
Added functional tests for search's required properties

loic-sharma:
Added functional tests that ensure that the required properties `Id`
and `Version` are present on both v2 and v3 search. These tests will
validate search results from both an empty search query as well as
Newtonsoft.Json

Related to [Gallery's #3826](https://github.com/NuGet/NuGetGallery/issues/3826)


tests/BasicSearchTests.FunctionalTests.Core/V2SearchFunctionalTests.cs
```diff
@@ -37,5 +38,30 @@ public class V2SearchFunctionalTests : BaseFunctionalTests
...
+        [Fact]
+        public async Task EmptyQueryResultHaveRequiredProperties()
+        {
+            // Act
+            var response = await Client.GetAsync(new V2SearchBuilder().RequestUri);
+            var content = await response.Content.ReadAsStringAsync();
+
+            // Newtonsoft's Json deserialization will throw exceptions if required
+            // properties are missing.
+            JsonConvert.DeserializeObject<V2SearchResult>(content);
```
skofman1: How will this show up in the logs? 
How about Assert.Throws

skofman1:
Please add comments to the tests explaining the purpose

tests/BasicSearchTests.FunctionalTests.Core/V2SearchFunctionalTests.cs
loic-sharma: This test should not throw an exception on the happy path. If the required fields are not present, an exception will be thrown and will cause the unit test to fail.

loic-sharma:
Improved comments and made behaviour of tests a little more explicit.

tests/BasicSearchTests.FunctionalTests.Core/Models/V2SearchResultEntry.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        [JsonProperty(Required = Required.Always)]
+        public string Version;
+
+        [JsonProperty(Required = Required.Always)]
+        public string NormalizedVersion;
+
+        public string Title;
+
+        public string Description;
```
joelverhagen: Should be `{ get; set; }` property

tests/BasicSearchTests.FunctionalTests.Core/V2SearchFunctionalTests.cs
```diff
@@ -37,5 +38,60 @@ public class V2SearchFunctionalTests : BaseFunctionalTests
             Assert.True(result.TotalHits.HasValue && result.TotalHits.Value > 0, "No results found, should find atleast some results for empty string query.");
             Assert.NotNull(result.Data);
         }
+
+        [Fact]
+        public void ShouldThrowExceptionForInvalidSearchResults()
```
joelverhagen: These are kind of unit tests on the `V2SearchResult` class, not functional tests. Do we have something that actually makes an HTTP request and asserts HTTP 500?

tests/BasicSearchTests.FunctionalTests.Core/Models/V2SearchResultEntry.cs
loic-sharma: Sure. I was following the pattern used for the `V3SearchResultEntry`, I will update that guy as well.

joelverhagen: thanks. In general public properties should be preferred over public fields (this is idiomatic C#)

tests/BasicSearchTests.FunctionalTests.Core/V2SearchFunctionalTests.cs
loic-sharma: That is correct. The real functional tests are `EmptyQueryResultsHaveRequiredProperties` and `ExistingPackageResultsHaveRequiredProperties`. Both of these functional tests assume that invalid results, should they happen, will cause a `JsonSerializationException` exception to be thrown. This `ShouldThrowExceptionForInvalidSearchResults` test is simply a sanity check that validates that assumption.

loic-sharma:
Added functional tests to verify that results for `BaseTestPackage` are exactly as expected.

tests/BasicSearchTests.FunctionalTests.Core/Constants.cs
```diff
@@ -6,7 +6,8 @@ namespace BasicSearchTests.FunctionalTests.Core
     public static class Constants
     {
         // Predefined Texts
-        public const string ExistentSearchString = "Newtonsoft.Json";
+        public const string TestPackageId = "BaseTestPackage";
```
shishirx34: Can you ensure that this package is available in search in Dev, INT and Prod all environments. The idea behind using Newtonsoft.Json was that it was certain to be there in these environments.

shishirx34: Also make sure all the properties are the same as required below in these envs.

tests/BasicSearchTests.FunctionalTests.Core/V3SearchFunctionalTests.cs
```diff
@@ -99,11 +101,71 @@ public void ShouldThrowExceptionForInvalidSearchResults()
...
+            Assert.True(package.Tags.Count() == 2);
+            Assert.True(package.Tags[0] == "Tag1");
+            Assert.True(package.Tags[1] == "Tag2");
+            Assert.True(package.Authors.Count() == 1);
+            Assert.True(package.Authors[0] == "clayco");
+            Assert.True(package.TotalDownloads != default(long));
+            Assert.True(package.Versions.Count() == 1);
+            Assert.True(package.Versions[0].Version == "1.0.0");
+            Assert.True(package.Versions[0].Downloads != default(long));
+            Assert.False(string.IsNullOrEmpty(package.Versions[0].AtId));
```
shishirx34: You could probably refactor these validations since reused above in V2 tests too. Up to you.

tests/BasicSearchTests.FunctionalTests.Core/V2SearchFunctionalTests.cs
```diff
@@ -37,5 +40,113 @@ public class V2SearchFunctionalTests : BaseFunctionalTests
...
+            Assert.True(package.Published != default(DateTime));
+            Assert.True(package.LastUpdated != default(DateTime));
+            Assert.True(package.DownloadCount != default(long));
+            Assert.True(package.FlattenedDependencies == "");
+            Assert.True(package.Dependencies.Count() == 0);
+            Assert.True(package.SupportedFrameworks.Count() == 0);
+            Assert.True(package.Hash == "5KJqge5+IYZkmba5C/pRVwjqwwaF1YM28xs6AiWMoxfxE/dzFVXJ5QGR7Rx2JmKWPLwz0R3eO+jWjd4lRX1WxA==");
+            Assert.True(package.HashAlgorithm == "SHA512");
+            Assert.True(package.PackageFileSize == 3943);
+            Assert.True(package.RequireslicenseAcceptance == false);
```
joelverhagen: 👍

tests/BasicSearchTests.FunctionalTests.Core/Constants.cs
loic-sharma: Will do! We decided to not test Newtonsoft.Json as the author could break our tests by simply changing the package's metadata. We avoid this potential issue by using a package that we control.

tests/BasicSearchTests.FunctionalTests.Core/V3SearchFunctionalTests.cs
loic-sharma: Talked to Shishir offline - I will add constants for all the common values.

tests/BasicSearchTests.FunctionalTests.Core/Constants.cs
loic-sharma: I tested this across all environments and ensured the package existed and its metadata matched what the functional test expects.

PR #169, opened 2017-04-28
Made verbatimVersion metadata field required

loic-sharma:
Made the package metadata field `verbatimVersion` required to ensure that V2 search queries return a `Version` in results.

Previously, an error would be thrown only if both the `verbatimVersion` and `version` fields were missing from the metadata. This was because the v3 search endpoint only uses the `verbatimVersion` as a fallback for when the `version` field isn't provided. However, v2 search ignores the `version` field and only uses the `verbatimVersion`. Thus, the `verbatimVersion` field should be required.

Related to [#3847](https://github.com/NuGet/NuGetGallery/issues/3847)


loic-sharma:
Scott ran this on primary and secondary of dev, int, and prod.

PR #170, opened 2017-04-28
Fixed QueryBoostingContext bug

loic-sharma:
The QueryBoostingContext's Load would return null if the given config
file was empty, causing exceptions to be thrown on searches. The
QueryBoostingContext will now return the default context on
deserialization failure and will log the failure.

Related to [#3851](https://github.com/NuGet/NuGetGallery/issues/3851)


src/NuGet.Indexing/Querys/QueryBoostingContext.cs
```diff
@@ -45,21 +45,24 @@ public static QueryBoostingContext Load(string fileName, ILoader loader, Framewo
             {
                 using (var reader = loader.GetReader(fileName))
                 {
-                    var serializer = new JsonSerializer();
+                    var value = new JsonSerializer().Deserialize<QueryBoostingContext>(reader);
```
joelverhagen: When would the file be empty?

joelverhagen:
Unit tests.

src/NuGet.Indexing/Querys/QueryBoostingContext.cs
scottbommarito: This is for running the search service locally without the aux data files.

loic-sharma:
Tweaked behaviour so that now we will throw if the configs are incorrect (instead of swallowing exceptions and then logging an error). Added unit tests.

loic-sharma:
Sorry about that, git was being silly. Finally added in the unit tests! :)

src/NuGet.Indexing/Querys/QueryBoostingContext.cs
```diff
@@ -49,20 +49,15 @@ public static QueryBoostingContext Load(string fileName, ILoader loader, Framewo
 
                     var value = serializer.Deserialize<QueryBoostingContext>(reader);
```
skofman1: is it still possible to run search service without any query boosting, or will it fail?

loic-sharma: The search service will now fail without query boosting. Notice that for the unit tests I added this line:

```
 { "searchSettings.v1.json", JsonConvert.SerializeObject(QueryBoostingContext.Default) },
```

This adds the query boosting context to allow the unit tests to pass. Before, the tests passed even without the query boosting context.

PR #174, opened 2017-05-01
V3 monitoring job: Fix V2 Hijacking

scottbommarito:
Last PR did not stop hijacking correctly. Apparently, adding `$filter=true` is only sufficient for `Packages`. Preventing hijacking of other endpoints must be done by adding `$orderby=Version` (or another ordering that the search service does not support).

(Also added a quick fix to `AggregateCursor`.)


src/Catalog/Helpers/UriUtils.cs
```diff
@@ -39,11 +45,25 @@ public static Uri GetNonhijackableUri(Uri originalUri)
...
+                /// Find the start of the next parameter (<see cref="BridgingCharacter"/>) or the end of the query.
+                var orderByEndIndex = originalUriString.IndexOf(BridgingCharacter, orderByStartIndex);
+                if (orderByEndIndex == -1)
+                {
+                    orderByEndIndex = originalUriString.Length;
+                }
+
+                // Replace the entire orderby with a nonhijackable orderby.
+                var orderByExpression = originalUriString.Substring(orderByStartIndex, orderByEndIndex - orderByStartIndex);
+                nonhijackableUriString = originalUriString.Replace(orderByExpression, OrderBy + NonhijackableOrderBy);
```
chenriksson: nit: Could this be simplified with a regex?

PR #175, opened 2017-05-02
V3 monitoring job: add registration validators

scottbommarito:
This PR adds a series of validators to the V3 monitoring job. They all check a single field in registration, or that it exists.

These include:
- `RegistrationExistsValidator`
    - Checks that registration index and leaf exist if `FindPackagesById` and `Package(...)` exist in the V2 feed.
- `RegistrationIdValidator`
    - Checks that the `Id` field in registration matches the V2 feed.
- `RegistrationVersionValidator`
    - Checks that the `Version` field in registration matches the V2 feed.
- `RegistrationListedValidator`
    - Checks that the `Listed` field in registration matches the V2 feed.
- `RegistrationPackageContentValidator`
    - Checks that the `PackageContent` field in registration and the `DownloadUrl` field in the V2 feed redirect to the same address.
- `RegistrationRequireLicenseAcceptanceValidator`
    - Checks that the `RequireLicenseAcceptance` field in registration matches the V2 feed.


src/NuGet.Services.Metadata.Catalog.Monitoring/Model/PackageRegistrationIndexMetadata.cs
```diff
@@ -0,0 +1,60 @@
...
+using NuGet.Versioning;
+
+namespace NuGet.Services.Metadata.Catalog.Monitoring
+{
+    /// <summary>
+    /// The metadata for a particular package in its registration index.
+    /// </summary>
+    public class PackageRegistrationIndexMetadata : PackageRegistrationLeafMetadata
+    {
+        public PackageRegistrationIndexMetadata()
```
skofman1: why do we need  an empty ctor?

src/NuGet.Services.Metadata.Catalog.Monitoring/Model/PackageRegistrationMetadataResourceV2Feed.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    public class PackageRegistrationMetadataResourceV2Feed : IPackageRegistrationMetadataResource
+    {
+        public PackageRegistrationMetadataResourceV2Feed(
+            V2FeedParser feedParser)
+        {
+            _feedParser = feedParser;
+        }
+
+        private V2FeedParser _feedParser;
```
skofman1: fields always first in the class!

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/ComparisonUtils.cs
```diff
@@ -0,0 +1,246 @@
...
+
+        /// <summary>
+        /// Compares <paramref name="v2"/> and <paramref name="v3"/>.
+        /// </summary>
+        /// <returns>
+        /// If both objects are perfectly equal, returns null.
+        /// If one object is null and the other is not, returns an empty <see cref="IEnumerable{PropertyInfo}"/>.
+        /// Otherwise, returns an <see cref="IEnumerable{PropertyInfo}"/> containing all properties that are not equal.
+        /// </returns>
+        public static IEnumerable<PropertyInfo> CheckEquality<T>(T v2, T v3) where T : class
```
skofman1: As discussed, comparing all properties is noisy and unhelpful, aspecially when we have issues open on the comparison. Lets not log any failed comparison with known issues until we resolve those issues.

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/ComparisonUtils.cs
```diff
@@ -0,0 +1,246 @@
...
+        {
+            public new bool Equals(object x, object y)
+            {
+                var dateX = (DateTimeOffset)x;
+                var dateY = (DateTimeOffset)y;
+
+                dateX = dateX.ToUniversalTime();
+                dateY = dateY.ToUniversalTime();
+
+                if (IsNullOr1900(dateX))
```
skofman1: add comments why 1900 is important

src/NuGet.Services.Metadata.Catalog.Monitoring/Model/PackageRegistrationLeafMetadata.cs
```diff
@@ -0,0 +1,31 @@
...
+        public PackageRegistrationLeafMetadata()
+        {
+        }
+
+        public PackageRegistrationLeafMetadata(V2FeedPackageInfo package)
+        {
+            Listed = package.IsListed;
+            PackageContent = package.DownloadUrl;
+            Published = package.Published;
+        }
```
chenriksson: Why not keep naming consistent across classes, i.e.  DownloadUrl ; DownloadUrlEqualityComparer?

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/ComparisonUtils.cs
```diff
@@ -0,0 +1,246 @@
...
+                // Consider null and empty equal.
+                // For fields without a value, the V2 implementation returns empty string and the V3 implementation returns null.
+                if (string.IsNullOrEmpty(stringX))
+                {
+                    return string.IsNullOrEmpty(stringY);
+                }
+
+                /// Try to parse the strings as <see cref="Uri"/>s and then compare them on that.
+                /// The V2 and V3 implementations sometimes returns <see cref="Uri"/>s in different forms (i.e. trailing slash).
+                if (Uri.TryCreate(stringX, UriKind.Absolute, out var uriX) && Uri.TryCreate(stringY, UriKind.Absolute, out var uriY))
```
chenriksson: Is there another way to distinguish Url properties, instead of attempting Uri.TryCreate on all string props? Property naming (ends with "Url"), or whitelist?

src/NuGet.Services.Metadata.Catalog.Monitoring/NuGet.Services.Metadata.Catalog.Monitoring.csproj
```diff
@@ -84,8 +84,18 @@
...
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="Model\IPackageRegistrationMetadataResource.cs" />
+    <Compile Include="Model\PackageRegistrationIndexMetadata.cs" />
+    <Compile Include="Model\PackageRegistrationLeafMetadata.cs" />
+    <Compile Include="Model\PackageRegistrationMetadataResourceV2Feed.cs" />
+    <Compile Include="Model\PackageRegistrationMetadataResourceV3.cs" />
+    <Compile Include="Providers\PackageRegistrationMetadataResourceV2FeedProvider.cs" />
+    <Compile Include="Providers\PackageRegistrationMetadataResourceV3Provider.cs" />
     <Compile Include="Utility\CommonLogger.cs" />
```
scottbommarito: Following client code naming convention

e.g.
- V2: `PackageMetadataResourceV2Feed`
- V3: `PackageMetadataResourceV3`

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/Json/NuGetVersionConverter.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using NuGet.Versioning;
+
+namespace NuGet.Services.Metadata.Catalog.Monitoring
+{
+    public class NuGetVersionConverter : JsonConverter
+    {
```
scottbommarito: As far as I can tell, it doesn't support null. I think I could probably leverage it and reduce the complexity of this code here.

src/NuGet.Services.Metadata.Catalog.Monitoring/Model/PackageRegistrationIndexMetadata.cs
scottbommarito: JSON serialization.

src/NuGet.Services.Metadata.Catalog.Monitoring/Model/PackageRegistrationLeafMetadata.cs
scottbommarito: `PackageContent` is the name of the field in the registration blobs. This constructor converts a `V2FeedPackageInfo` into a form that can be directly compared to a `PackageRegistrationLeafMetadata`.

Perhaps I should
a) add a comment here
b) remove this and compare the values directly (given that I'm already planning to remove `ComparisonUtils` which was the only reason a single class was required)

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/Json/DateTimeOffsetNullableConverter.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace NuGet.Services.Metadata.Catalog.Monitoring
+{
+    public class DateTimeOffsetNullableConverter : JsonConverter
+    {
```
scottbommarito: looked into it--turns out it's unnecessary to have this, as `Newtonsoft.Json` is smart enough to convert it without this custom converter.

src/NuGet.Services.Metadata.Catalog.Monitoring/Model/PackageRegistrationLeafMetadata.cs
scottbommarito: Added

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/ComparisonUtils.cs
scottbommarito: Removed this class entirely.

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/ComparisonUtils.cs
scottbommarito: Removed this class entirely

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/ComparisonUtils.cs
scottbommarito: Done--split into `Validator`s that exclusively check the properties we care about--which is currently  `Listed`, `PackageContent`, `Id`, `Version`, and `RequireLIcenseAcceptance`. also added a `Validator` that checks if the registration pages exist.

src/NuGet.Services.Metadata.Catalog.Monitoring/Model/PackageRegistrationMetadataResourceV2Feed.cs
scottbommarito: fixed in all classes

scottbommarito:
Added tests!

src/Catalog/Helpers/UriUtils.cs
```diff
@@ -83,5 +85,18 @@ public static Uri GetNonhijackableUri(Uri originalUri)
...
             return nonhijackableUri;
         }
+
+        /// <summary>
+        /// Makes a web request to <paramref name="uri"/> and returns <see cref="HttpRequestMessage.RequestUri"/> of the <see cref="HttpResponseMessage.RequestMessage"/>.
+        /// In other words, if <paramref name="uri"/> redirects to another address, this method returns the address that was redirected to.
+        /// </summary>
+        public static async Task<Uri> GetRedirectedRequestMessageUri(Uri uri)
+        {
+            using (var client = new HttpClient())
```
loic-sharma: Constantly creating `HttpClient` can cause issues. See [this](https://github.com/mspnp/performance-optimization/blob/master/ImproperInstantiation/docs/ImproperInstantiation.md) and [this](https://aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/). Can you cache this instead?

src/Catalog/Helpers/UriUtils.cs
```diff
@@ -1,7 +1,9 @@
 ﻿using System;
```
agr: Add the header

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/Exceptions/MetadataInconsistencyException.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Services.Metadata.Catalog.Monitoring
+{
+    public class MetadataInconsistencyException : ValidationException
+    {
+        public MetadataInconsistencyException(string additionalMessage)
+            : base("The metadata between V2 and V3 is inconsistent!" +
+                  (additionalMessage != null ? $" {additionalMessage}" : ""))
```
agr: `additionalMessage ?? ""` ?

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/Registration/RegistrationExistsValidator.cs
```diff
@@ -0,0 +1,39 @@
...
+        public override async Task CompareLeaf(PackageRegistrationLeafMetadata v2, PackageRegistrationLeafMetadata v3)
+        {
+            var v2Exists = v2 != null;
+            var v3Exists = v3 != null;
+            var isEqual = v2Exists == v3Exists;
+
+            if (!isEqual)
+            {
+                throw new MetadataInconsistencyException<PackageRegistrationLeafMetadata>(v2, v3,
+                    $"One exists and the other does not!");
```
agr: Wouldn't it be better to specify which one exists and which one is not?

tests/NgTests/Validators/RegistrationValidatorTests.cs
```diff
@@ -0,0 +1,123 @@
...
+                return registrationValidatorTestDataTypes.Select(
+                    t => t.GetConstructor(new Type[] { }).Invoke(null) as RegistrationValidatorTestData);
+            }
+        }
+
+        private static IEnumerable<Tuple<T, T>> GetEqualPairs<T>(IEnumerable<T> values)
+        {
+            foreach (var obj in values)
+            {
+                yield return Tuple.Create(obj, obj);
```
ryuyu: This should create 2 objects that have the same values, and not a tuple of two references to the same object. That way, we actually test your comparisons.

src/NuGet.Services.Metadata.Catalog.Monitoring/Model/PackageRegistrationIndexMetadata.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            Authors = string.Join(", ", package.Authors);
+            Description = package.Description;
+            IconUrl = package.IconUrl;
+            Id = package.Id;
+            MinClientVersion = package.MinClientVersion;
+            ProjectUrl = package.ProjectUrl;
+            RequireLicenseAcceptance = package.RequireLicenseAcceptance;
+            Summary = package.Summary;
+            Tags = package.Tags.Split(' ').Where(s => !string.IsNullOrEmpty(s));
```
xavierdecoster: Can be replaced by `Tags = package.Tags.Split(' ', StringSplitOptions.RemoveEmptyEntries)`?

src/NuGet.Services.Metadata.Catalog.Monitoring/NuGet.Services.Metadata.Catalog.Monitoring.csproj
xavierdecoster: For consistency, should we perhaps track an issue to rename the other classes to also follow client code naming convention?

scottbommarito: Not sure how necessary that is--I'm doing it here because the specific classes that I'm following the convention for are extensions of the `NuGet.Core` APIs.

src/NuGet.Services.Metadata.Catalog.Monitoring/Model/PackageRegistrationIndexMetadata.cs
loic-sharma: `isNullOrEmpty` may be better as `StringSplitOptions.RemoveEmptyEntries` keeps whitespace entries. If you stick to `isNullOrEmpty`, I would recommend you add a `ToArray` or `ToList` at the end so that empy tags aren't filtered out every time the `Tags` enumerable is iterated.

src/NuGet.Services.Metadata.Catalog.Monitoring/Model/PackageRegistrationMetadataResourceV2Feed.cs
```diff
@@ -0,0 +1,46 @@
...
+            return feedPackage != null ? new PackageRegistrationLeafMetadata(feedPackage) : null;
+        }
+
+        private async Task<V2FeedPackageInfo> GetPackageFromIndex(PackageIdentity package, ILogger log, CancellationToken token)
+        {
+            var feedPackages = await _feedParser.FindPackagesByIdAsync(package.Id, log, token);
+            return feedPackages.FirstOrDefault(p => p.Version == package.Version);
+        }
+
+        private async Task<V2FeedPackageInfo> GetPackageFromLeaf(PackageIdentity package, ILogger log, CancellationToken token)
```
loic-sharma: Any particular reason for this helper method? You could just call `_feedParser.GetPackage` directly in `GetLeaf`

src/NuGet.Services.Metadata.Catalog.Monitoring/Model/PackageRegistrationMetadataResourceV3.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        public async Task<PackageRegistrationLeafMetadata> GetLeaf(PackageIdentity package, ILogger log, CancellationToken token)
+        {
+            var feedPackage = await GetPackageFromLeaf(package, log, token);
+            return feedPackage != null ?
+                JsonConvert.DeserializeObject<PackageRegistrationLeafMetadata>(feedPackage.ToString()) :
+                null;
+        }
+
+        private async Task<JObject> GetPackageFromIndex(PackageIdentity package, ILogger log, CancellationToken token)
```
loic-sharma: Any particular reason for this helper method? You could just call `_registration.GetPackageMetadata` directly in `GetIndex`

src/NuGet.Services.Metadata.Catalog.Monitoring/Providers/PackageRegistrationMetadataResourceV2FeedProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+            {
+                var serviceDocumentResource = await source.GetResourceAsync<ODataServiceDocumentResourceV2>(token);
+
+                var httpSourceResource = await source.GetResourceAsync<HttpSourceResource>(token);
+                var feedParser = new V2FeedParser(httpSourceResource.HttpSource, serviceDocumentResource.BaseAddress, source.PackageSource.Source);
+
+                resource = new PackageRegistrationMetadataResourceV2Feed(feedParser);
+            }
+
+            return new Tuple<bool, INuGetResource>(resource != null, resource);
```
loic-sharma: Can we use C# 7 in this project? If so, you could use the new Tuple syntax! `return (result != null, resource);`

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/Registration/RegistrationValidator.cs
```diff
@@ -0,0 +1,77 @@
...
+                exceptions.Add(new ValidationException("Registration leaf metadata does not match the Packages(Id='...',Version='...') metadata!", e));
+            }
+
+            if (exceptions.Any())
+            {
+                throw new AggregateException(exceptions);
+            }
+        }
+
+        public abstract Task CompareIndex(PackageRegistrationIndexMetadata v2, PackageRegistrationIndexMetadata v3);
```
loic-sharma: Why do `CompareIndex` and `CompareLeaf` not have `async` on their method signatures? Is that intentional?

src/NuGet.Services.Metadata.Catalog.Monitoring/Model/PackageRegistrationIndexMetadata.cs
scottbommarito: Added `ToArray`

src/NuGet.Services.Metadata.Catalog.Monitoring/Model/PackageRegistrationMetadataResourceV2Feed.cs
scottbommarito: I wanted to make `GetIndex` and `GetLeaf` look as  similar as possible. Tried to reduce it down further but it seemed like over-engineering. I like it like this.

src/NuGet.Services.Metadata.Catalog.Monitoring/Model/PackageRegistrationMetadataResourceV3.cs
scottbommarito: I wanted to make `GetIndex` and `GetLeaf` look as  similar as possible. Tried to reduce it down further but it seemed like over-engineering. I like it like this.

src/NuGet.Services.Metadata.Catalog.Monitoring/Providers/PackageRegistrationMetadataResourceV2FeedProvider.cs
scottbommarito: I'm trying to mirror client code, which currently does not use the C# 7 syntax.

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/Registration/RegistrationValidator.cs
scottbommarito: Yes, `async` is only an implementation detail of the method--specifically, whether or not the implementation of the method returns `Task<T>` or `T`. It has no meaning and is not valid in an `abstract` declaration.

scottbommarito: E.g.

```
public abstract Task<int> GetValue();
...
public async Task<int> GetValue()
{
    return 0;
}
...
public Task<int> GetValue()
{
    return Task.FromResult(0);
}
```

tests/NgTests/Validators/RegistrationValidatorTests.cs
scottbommarito: nice catch! fixed

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/Registration/RegistrationExistsValidator.cs
scottbommarito: done!

src/Catalog/Helpers/UriUtils.cs
scottbommarito: done!

src/Catalog/Helpers/UriUtils.cs
scottbommarito: done!

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/Registration/RegistrationIdValidator.cs
```diff
@@ -0,0 +1,51 @@
...
+    public class RegistrationIdValidator : RegistrationExistsValidator
+    {
+        public RegistrationIdValidator(
+            IDictionary<FeedType, SourceRepository> feedToSource, 
+            ILogger<RegistrationIdValidator> logger) 
+            : base(feedToSource, logger)
+        {
+        }
+
+        public override async Task CompareIndex(ValidationContext data, PackageRegistrationIndexMetadata v2, PackageRegistrationIndexMetadata v3)
```
agr: Probably would be worth to make a helper method to share the `Compare*` implementation between this and other classes below.
Something like
```
public static void AreEqual<T>(PackageRegistrationIndexMetadata v2, PackageRegistrationIndexMetadata v3, Func<PackageRegistrationIndexMetadata, T> extractValue, string nameOfValue)
{
    if (v2 == null || v3 == null) 
    {
        return; 
    } 
    var isEqual = extractValue(v2) == extractValue(v3);
    if (!isEqual) 
    { 
        throw new MetadataInconsistencyException<PackageRegistrationIndexMetadata>(v2, v3, 
            $"{nameOfValue} does not match!"); 
    } 
}
```

scottbommarito: Restructured the validations in a different way that resolves this.

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/Registration/RegistrationExistsValidator.cs
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        public override async Task CompareLeaf(ValidationContext data, PackageRegistrationLeafMetadata v2, PackageRegistrationLeafMetadata v3)
+        {
+            var v2Exists = v2 != null;
+            var v3Exists = v3 != null;
+
+            if (v2Exists != v3Exists)
+            {
+                var existsString = "exists";
```
agr: minor note: `const string existsString` here and below?

scottbommarito: fixed

PR #176, opened 2017-05-03
Make use of different names of versions consistent

ryuyu:
Normalize use of VerbatimVersion,NormalizedVersion, and FullVersion names

@joelverhagen @xavierdecoster

### DIFF
--- a/src/NuGet.Indexing/Extraction/DocumentCreator.cs
+++ b/src/NuGet.Indexing/Extraction/DocumentCreator.cs
@@ -106,7 +106,7 @@ namespace NuGet.Indexing
                 NuGetVersion parsedVerbatimVersion;
                 if (NuGetVersion.TryParse(verbatimVersion, out parsedVerbatimVersion))
                 {
-                    AddField(document, LuceneConstants.NormalizedVersionPropertyName, parsedVerbatimVersion.ToNormalizedString(), Field.Index.NOT_ANALYZED);
+                    AddField(document, LuceneConstants.NormalizedVersionPropertyName, parsedVerbatimVersion.ToNormalizedString(), Field.Index.ANALYZED);
                     AddField(document, LuceneConstants.FullVersionPropertyName, parsedVerbatimVersion.ToFullString(), Field.Index.NOT_ANALYZED);
                 }
                 else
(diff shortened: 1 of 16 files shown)


tests/NuGet.IndexingTests/TestSupport/Constants.cs
```diff
@@ -24,7 +24,7 @@ public class Constants
         public static readonly string LucenePropertyProjectUrl = "ProjectUrl";
         public static readonly string LucenePropertySummary = "Summary";
         public static readonly string LucenePropertyTitle = "Title";
-        public static readonly string LucenePropertyVersion = "Version";
+        public static readonly string LucenePropertyVersion = "FullVersion";
```
xavierdecoster: Did you mean `LucenePropertyFullVersion`?

joelverhagen:
Looks really good. We do need:

- unit tests in Search service and the lucene index creation code
- functional test on search service

src/NuGet.Indexing/Extraction/DocumentCreator.cs
```diff
@@ -99,41 +99,24 @@ private static void AddId(Document document, IDictionary<string, string> package
...
-                    NuGetVersion nuGetVersion;
-                    if (NuGetVersion.TryParse(verbatimVersion, out nuGetVersion))
-                    {
-                        version = nuGetVersion.ToNormalizedString();
-                    }
-                    else
-                    {
-                        errors.Add($"Unable to parse '{MetadataConstants.VersionPropertyName}' as NuGetVersion.");
-                    }
+                    AddField(document, LuceneConstants.NormalizedVersionPropertyName, parsedVerbatimVersion.ToNormalizedString(), Field.Index.NOT_ANALYZED);
```
loic-sharma: Previously, the `NormalizedVersionPropertyName` was marked as `Field.Index.ANALYZED`, now it's `Field.Index.NOT_ANALYZED`. Is that intended? What is the difference between the two?

joelverhagen: Nice catch. This needs to be `ANALYZED` if we want the `version:1.0.0` to work (we do!)

joelverhagen: Also, we now need a functional test to verify searching by version work.

ryuyu:
Test have been corrected. It appears we have tests that cover this already. Note that this is not a behavioural change, so the tests we have should validate the same.

PR #177, opened 2017-05-04
Enable blob snapshot

cristinamanum:
Add support for taking blob snapshots when a blob is updated.


scottbommarito:
Should we make this change here now that we have https://github.com/NuGet/ServerCommon/blob/develop/src/NuGet.Services.Storage/AzureStorage.cs? At the bare minimum, the change will need to be made in both places.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -125,8 +126,43 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
+            {
+                return blob;
+            }
+            Stopwatch sw = new Stopwatch();
+            Trace.WriteLine($"SnapshotStart:{blob.Uri.ToString()}.");
+            sw.Start();
+            try
+            {
+                //read the current snapshot list ordered decreasing by the SnapshotTime
+                var allSanpshots = blob.Container.ListBlobs(blob.Name, true, blobListingDetails: BlobListingDetails.Snapshots)
```
shishirx34: allSnapshots - typo

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -125,8 +126,43 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
+                                   .Where((item) => { return ((CloudBlockBlob)item).IsSnapshot; })
+                                   .OrderBy((item) => { return ((CloudBlockBlob)item).SnapshotTime; });
+
+                if (allSanpshots.Count() >= MaxCountOfBlobSnapshots)
+                {
+                    CloudBlockBlob snapshotToBeDeleted = (CloudBlockBlob)allSanpshots.First();
```
shishirx34: minor thing but ideally you'd want to delete all snapshots exceeding `MaxCountofBlobSnapshots` rather than just one. May be not get into this situation if this runs as a single job all the time. But better to put it in here.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -125,8 +126,43 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
+                }
+                var snapshot =  blob.CreateSnapshot();
+                sw.Stop();
+                Trace.WriteLine($"SnapshotCreated:milliseconds={sw.ElapsedMilliseconds}:{blob.Uri.ToString()}:{snapshot.SnapshotQualifiedUri}");
+                return snapshot;
+            }
+            catch(StorageException storageException)
+            {
+                sw.Stop();
+                Trace.WriteLine($"EXCEPTION:CreateSnapshot: Failed to take the snapshot for blob {blob.Uri.ToString()}. Exception{storageException.ToString()}");
```
shishirx34: might as well add the time it took in here(`sw.ElapsedMilliseconds`)?

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -183,8 +219,7 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
             string name = GetName(resourceUri);
 
             CloudBlockBlob blob = _directory.GetBlockBlobReference(name);
-
-            await blob.DeleteAsync(cancellationToken);
+            await blob.DeleteAsync(deleteSnapshotsOption:DeleteSnapshotsOption.IncludeSnapshots, accessCondition:null, options:null, operationContext:null,  cancellationToken:cancellationToken);
```
shishirx34: On the delete operation of a blob, all snapshots will get deleted, is that the desired behavior? Essentially we don't have a recovery on delete.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -125,8 +126,43 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
+                                   .Where((item) => { return ((CloudBlockBlob)item).IsSnapshot; })
+                                   .OrderBy((item) => { return ((CloudBlockBlob)item).SnapshotTime; });
+
+                if (allSanpshots.Count() >= MaxCountOfBlobSnapshots)
+                {
+                    CloudBlockBlob snapshotToBeDeleted = (CloudBlockBlob)allSanpshots.First();
+                    snapshotToBeDeleted.DeleteIfExists();
+                    Trace.WriteLine($"SnapshotDelete:{snapshotToBeDeleted.SnapshotQualifiedUri} was deleted.");
+                }
+                var snapshot =  blob.CreateSnapshot();
```
joelverhagen: Create first, then delete. If snapshot creation fails we don't want less than 30.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -125,8 +126,43 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
+                return blob;
+            }
+            Stopwatch sw = new Stopwatch();
+            Trace.WriteLine($"SnapshotStart:{blob.Uri.ToString()}.");
+            sw.Start();
+            try
+            {
+                //read the current snapshot list ordered decreasing by the SnapshotTime
+                var allSanpshots = blob.Container.ListBlobs(blob.Name, true, blobListingDetails: BlobListingDetails.Snapshots)
+                                   .Where((item) => { return ((CloudBlockBlob)item).IsSnapshot; })
```
joelverhagen: Use `Cast<CloudBlock>()` LINQ to avoid the repeated casts.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -16,6 +16,7 @@ namespace NuGet.Services.Metadata.Catalog.Persistence
     public class AzureStorage : Storage
     {
         private readonly CloudBlobDirectory _directory;
+        private readonly int MaxCountOfBlobSnapshots = 30;
```
joelverhagen: We need to make this snapshot count configurable. Flat container may need a different number of snapshots from catalog.

skofman1: Also, this code should be reused (@agr moved it to a shared repo), and not for all usages we will require snapshots.

src/Catalog/Persistence/AzureStorage.cs
skofman1: true? please use named parameters for readability

src/Catalog/Persistence/AzureStorage.cs
skofman1: nit: use var

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -125,8 +126,43 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
+
+                if (allSanpshots.Count() >= MaxCountOfBlobSnapshots)
+                {
+                    CloudBlockBlob snapshotToBeDeleted = (CloudBlockBlob)allSanpshots.First();
+                    snapshotToBeDeleted.DeleteIfExists();
+                    Trace.WriteLine($"SnapshotDelete:{snapshotToBeDeleted.SnapshotQualifiedUri} was deleted.");
+                }
+                var snapshot =  blob.CreateSnapshot();
+                sw.Stop();
+                Trace.WriteLine($"SnapshotCreated:milliseconds={sw.ElapsedMilliseconds}:{blob.Uri.ToString()}:{snapshot.SnapshotQualifiedUri}");
```
skofman1: Can we reduce the amount of traces? Might get noisy in the log

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Updated.

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Updated.

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Updated.

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Yes - a blob cannot be deleted if it has snapshots.

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Updated.

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Updated, Also Updated to have default 1 snapshot per the offline discussion

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Updated. I will follow up with @agr on the shared repo.

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Updated

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: I would like to capture the operation time.  It can be re-evaluated if it is too noisy.

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Updated to have only one snapshot per blob.

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: In order to eliminate the perf impact, the design changed to have only one snapshot per blob.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -125,8 +124,43 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
+            if (blob == null )
+            {
+                //no action
+                return false;
+            }
+            Stopwatch sw = new Stopwatch();
+            sw.Start();
+            try
+            {
+                var allSnapshots = blob.Container.ListBlobs(prefix: blob.Name, useFlatBlobListing: true, blobListingDetails: BlobListingDetails.Snapshots);
```
joelverhagen: nit: please wrap long lines (120 is good b/c GitHub source viewing is most natural)

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -125,8 +124,43 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
+                //the above call will return at least one blob the original
+                if (allSnapshots.Count() == 1)
+                {
+                    var snapshot = blob.CreateSnapshot();
```
joelverhagen: Is there an async version of this method?

cristinamanum: Yes. Updated to go use the async version

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Updated

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -125,8 +124,46 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
             }
+            await TryTakeBlobSnapshotAsync(blob);
         }
 
+        /// <summary>
+        /// Take one snapshot only if there is not any snapshot for the specific blob
+        /// This will prevent the blob to be deleted by a not intended delete action
+        /// </summary>
+        /// <param name="blob"></param>
+        /// <returns></returns>
```
skofman1: nit: document the return value of remove the tag

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -125,8 +124,46 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
+            try
+            {
+                var allSnapshots = blob.Container.
+                                   ListBlobs(prefix: blob.Name,
+                                             useFlatBlobListing: true,
+                                             blobListingDetails: BlobListingDetails.Snapshots);
+                //the above call will return at least one blob the original
+                if (allSnapshots.Count() == 1)
+                {
+                    var snapshot = await blob.CreateSnapshotAsync();
```
skofman1: nit: you can move sw.Stop() and tracing the time elapsed to a finally {} block

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -125,8 +124,46 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
+        /// <returns></returns>
+        private async Task<bool> TryTakeBlobSnapshotAsync(CloudBlockBlob blob)
+        {
+            if (blob == null )
+            {
+                //no action
+                return false;
+            }
+            Stopwatch sw = new Stopwatch();
+            sw.Start();
```
agr: `Stopwatch` has a static [`StartNew()`](https://msdn.microsoft.com/en-us/library/system.diagnostics.stopwatch.startnew(v=vs.110).aspx) method that does what these two lines do. Would save you one line of code :)

PR #180, opened 2017-05-11
Remove unnecessary reflection

loic-sharma:
Tweaked the `FeedHelper`'s `GetPackagesInOrder` method to no longer use reflection. We should avoid reflection unless absolutely necessary as issues may go undetected at compile-time.


skofman1:
Is there an issue related to this change? How was the code validated?

loic-sharma:
There's no issue, no. Should I create one?

This doesn't change any existing functionality and there are unit tests around this method already. I ensured that those tests still passed.

PR #182, opened 2017-05-16
Remove non XUnit tests from CatalogTests

skofman1:
Fixes: https://github.com/NuGet/NuGetGallery/issues/3819


tests/CatalogTests/PackageCatalogItemTests.cs
```diff
@@ -1,128 +0,0 @@
...
-using Xunit;
-
-namespace CatalogTests.Helpers
-{
-    public class PackageCatalogItemTests
-    {
-        [Theory]
-        [InlineData("Newtonsoft.Json.9.0.2-beta1")]
-        [InlineData("TestPackage.SemVer2.1.0.0-alpha.1")]
-        public void CreateContent_ProducesExpectedJson(string packageName)
```
joelverhagen: Aren't these real tests?

skofman1: Good catch!

PR #183, opened 2017-05-17
Fixed Application Insights Sampling and Duplication Issues

loic-sharma:
Don't sample Traces and Exceptions sent to Application Insights. Fixed request log duplication issue.

Fix for [Gallery #3817](https://github.com/NuGet/NuGetGallery/issues/3817)


skofman1:
how did u test this?

loic-sharma:
Two manual tests were necessary:

1. For the request logs getting duplicated as traces, I verified that traces did not contain request information after my change.
2. I cheated a little for the sampling configuration change. I tweaked the search service's configuration to sample aggressively and then generated load that caused traces to be sampled. I then added the `<ExcludedTypes>Trace;Exception</ExcludedTypes>` config from this pull request and ensured that traces were not sampled even with the aggressive sampling turned on.

PR #185, opened 2017-05-18
V3 backup job.

cristinamanum:
Contains two new jobs:
1. Backup of the V3 pipeline for the catalog, lucene and registration 
2. Job for taking one blob snapshot. This job intended to be used offline, for taking initial snapshot of the blobs that need to be protected from delete.


src/Ng/Ng.csproj
```diff
@@ -415,6 +418,34 @@
     </BootstrapperPackage>
   </ItemGroup>
   <ItemGroup>
+    <Content Include="binaries\azcopy\AzCopy.exe">
```
joelverhagen: Are we allowed to check in AzCopy like this? From the website it looks like only an MSI. Now we're checking in a copy to our repo with no license.

src/Ng/Jobs/BackupV3Storage.cs
```diff
@@ -0,0 +1,236 @@
...
+                                   $"/SourceKey:{sourceAccountKey} /DestKey:{_destStorageKeyValue} " +
+                                   $"/S /Z:{azCopyTempFolder} /V:{azCopyLogPath}";
+
+                string argumentsLog = $"/Source:{azCopyTempFolder} /Dest:https://{_destStorageAccountName}.blob.core.windows.net/logs" +
+                                      $" /DestKey:{_destStorageKeyValue} /destType:blob /Pattern:{logFile}";
+
+                try
+                {
+                    ProcessStartInfo copyToAzureProc = new ProcessStartInfo();
+                    copyToAzureProc.FileName = @"cmd.exe";
```
joelverhagen: Why not call azcopy directly?

src/Ng/Jobs/BackupV3Storage.cs
```diff
@@ -0,0 +1,236 @@
...
+                    return false;
+                }
+                try
+                {
+                    ProcessStartInfo copyToAzureProcLog = new ProcessStartInfo();
+                    copyToAzureProcLog.FileName = @"cmd.exe";
+                    copyToAzureProcLog.Arguments = $"/c {AzCopyPath} {argumentsLog}";
+                    copyToAzureProcLog.CreateNoWindow = false;
+                    copyToAzureProcLog.UseShellExecute = false;
+                    Process pLog = Process.Start(copyToAzureProcLog);
```
joelverhagen: nit var

joelverhagen: `Process` is disposable. Please use `using`.

src/Ng/Jobs/BackupV3Storage.cs
```diff
@@ -0,0 +1,236 @@
...
+                Console.Write(result);
+#endif
+                    sw.Stop();
+                    int exitCode = p.ExitCode;
+                    Logger.LogInformation($"EndContainerCopy:{container}:{exitCode}:{sw.ElapsedMilliseconds}");
+                    p.Close();
+                }
+                catch (Exception ex)
+                {
+                    Logger.LogError($"BackupV3Storage:Exception on backup save.{ex.ToString()}.");
```
joelverhagen: You are encoding the type name into the log message, this should not be necessary if you use `ILogger<T>`.

src/Ng/Jobs/BackupV3Storage.cs
```diff
@@ -0,0 +1,236 @@
...
+                    sw.Start();
+                    Process p = Process.Start(copyToAzureProc);
+                    p.WaitForExit();
+#if Debug
+                string result = p.StandardOutput.ReadToEnd();
+                Console.Write(result);
+#endif
+                    sw.Stop();
+                    int exitCode = p.ExitCode;
+                    Logger.LogInformation($"EndContainerCopy:{container}:{exitCode}:{sw.ElapsedMilliseconds}");
```
joelverhagen: Please parameterize log messages:
```csharp
Logger.LogInformation("EndContainerCopy:{container}:{exitCode}:{duration}", container, exitCode, sw.ElapsedMilliseconds);
```

src/Ng/Jobs/BackupV3Storage.cs
```diff
@@ -0,0 +1,236 @@
...
+                                      $" /DestKey:{_destStorageKeyValue} /destType:blob /Pattern:{logFile}";
+
+                try
+                {
+                    ProcessStartInfo copyToAzureProc = new ProcessStartInfo();
+                    copyToAzureProc.FileName = @"cmd.exe";
+                    copyToAzureProc.Arguments = $"/c {AzCopyPath} {arguments}";
+                    copyToAzureProc.CreateNoWindow = false;
+                    copyToAzureProc.UseShellExecute = false;
+#if Debug
```
joelverhagen: Is it necessary to have different behavior on debug?

src/Ng/Jobs/BackupV3Storage.cs
```diff
@@ -0,0 +1,236 @@
...
+                   + $"[-{Arguments.VaultName} <keyvault-name> "
+                   + $"[-{Arguments.VaultName} <keyvault-name> "
+                   + $"-{Arguments.ClientId} <keyvault-client-id> "
+                   + $"-{Arguments.CertificateThumbprint} <keyvault-certificate-thumbprint> "
+                   + $"[-{Arguments.ValidateCertificate} true|false]]";
+        }
+
+        protected override void Init(IDictionary<string, string> arguments, CancellationToken cancellationToken)
+        {
+            _luceneStorageAccountName = arguments.GetOrThrow<string>(Arguments.LuceneStorageAccountName);
```
joelverhagen: Looks like you need an account, key, and container for each container. Could you make the arguments more generic to allow any number of containers?

joelverhagen: Order does matter but that could be derived from argument order

src/Ng/Jobs/BackupV3Storage.cs
```diff
@@ -0,0 +1,236 @@
...
+        }
+
+        private void RefreshLogData(string logFolder, string logFile)
+        {
+            if(Directory.Exists(logFolder))
+            {
+                Directory.Delete(logFolder, true);
+            }
+            Directory.CreateDirectory(logFolder);
+            var stream = File.Create(logFile);
```
joelverhagen: `using`

src/Ng/Jobs/BackupV3Storage.cs
```diff
@@ -0,0 +1,236 @@
...
+            }
+            else
+            {
+                return false;
+            }
+        }
+
+        private void RefreshLogData(string logFolder, string logFile)
+        {
+            if(Directory.Exists(logFolder))
```
joelverhagen: Why do you have to delete the log folder?

src/Ng/Jobs/BackupV3Storage.cs
```diff
@@ -0,0 +1,236 @@
...
+
+            try
+            {
+                CloudBlobContainer container = blobClient.GetContainerReference(destContainer);
+                await container.DeleteIfExistsAsync();
+                return true;
+            }
+            catch (Exception ex)
+            {
+                Logger.LogError($"BackupV3Storage:Exception on delete container {destContainer}.{ex.ToString()}.");
```
joelverhagen: Log error has a special overload for an exception, right?

Also, does this end up in the `exceptions` table in AI?

src/Ng/Jobs/BackupV3Storage.cs
```diff
@@ -0,0 +1,236 @@
...
+        private CloudBlobClient GetCloudBlobClient(string storageAccountName, string storageAccountKey)
+        {
+            string storageConnectionString = $"DefaultEndpointsProtocol=https;AccountName={storageAccountName};AccountKey={storageAccountKey}";
+            CloudStorageAccount storageAccount = CloudStorageAccount.Parse(storageConnectionString);
+            CloudBlobClient blobClient = storageAccount.CreateCloudBlobClient();
+
+            return blobClient;
+        }
+
+        private string Format(DateTime date)
```
joelverhagen: Nit please use `DateTimeOffset` for all new code

src/Ng/Jobs/BlobSnapshotJob.cs
```diff
@@ -0,0 +1,90 @@
...
+            CloudStorageAccount storageAccount = CloudStorageAccount.Parse(_connectionString);
+            CloudBlobClient blobClient = storageAccount.CreateCloudBlobClient();
+            TakeOneSnapshot(_container, blobClient);
+            return Task.FromResult(true);
+        }
+
+        private void TakeOneSnapshot(string containerName, CloudBlobClient client)
+        {
+            CloudBlobContainer container = client.GetContainerReference(containerName);
+            var blobList = container.ListBlobs(prefix:null, useFlatBlobListing:true, blobListingDetails:BlobListingDetails.None);
```
joelverhagen: nit spaces between named parameter and value

src/Ng/Jobs/BlobSnapshotJob.cs
```diff
@@ -0,0 +1,90 @@
...
+            int snapshotCount = 0;
+            Stopwatch sw = new Stopwatch();
+            sw.Start();
+            Parallel.ForEach(blobList, (item) =>
```
joelverhagen: What determines the maximum degree of parallelism here?

src/Ng/Jobs/BlobSnapshotJob.cs
joelverhagen: This should use `ListBlobsSegmentedAsync` so the whole blob list is never in memory.

src/Ng/Jobs/BlobSnapshotJob.cs
```diff
@@ -0,0 +1,90 @@
...
+                catch(Exception ex)
+                {
+                    this.Logger.LogError("The snapshot failed for blob {0}. Exception {1}", item.Uri, ex.ToString());
+                }
+            });
+            sw.Stop();
+
+            if (_verbose)
+            {
+                this.Logger.LogInformation("Created {0} snapshots in {1} milliseconds", snapshotCount, sw.ElapsedMilliseconds);
```
joelverhagen: You can use names instead of numbers for `Logger.Log*` methods.

src/Ng/Jobs/BlobSnapshotJob.cs
```diff
@@ -0,0 +1,90 @@
...
+                {
+                    CloudBlockBlob blob = item as CloudBlockBlob;
+                    if (blob != null)
+                    {
+                        //list the snapshots and take a snapshot only if the blob does not have any snapshot
+                        var expandedList = container.ListBlobs(prefix: blob.Name, useFlatBlobListing: true, blobListingDetails: BlobListingDetails.Snapshots).ToList();
+                        if (expandedList.Count() == 1)
+                        {
+                            Interlocked.Increment(ref snapshotCount);
+                            blob.Snapshot();
```
joelverhagen: Can we use `SnapshotAsync`? Less waiting on IO theoretically, right?

src/Ng/Jobs/BlobSnapshotJob.cs
```diff
@@ -0,0 +1,90 @@
...
+                   + $"-{Arguments.ConnectionString} <connectionString> "
+                   + $"-{Arguments.Container} <container> "
+                   + $"[-{Arguments.Verbose} true|false]";
+        }
+
+        protected override void Init(IDictionary<string, string> arguments, CancellationToken cancellationToken)
+        {
+            _connectionString = arguments.GetOrThrow<string>(Arguments.ConnectionString);
+            _container = arguments.GetOrThrow<string>(Arguments.Container);
+            _verbose = arguments.GetOrThrow<bool>(Arguments.Verbose);
```
joelverhagen: Can this verbose do a little more? It only hides a single message right?

src/Ng/Jobs/BackupV3Storage.cs
scottbommarito: something like this should work

```
-numStorage 3 -storageAccountName1 ... -storageAccountKey1 ... -storageAccountContainer1 ... -storageAccountName2 ...
```

```
var numStorage = arguments.GetOrThrow<string>(Arguments.NumStorage);
for (int i = 0; i < numStorage; i++)
{
    var accountName = arguments.GetOrThrow<string>(Arguments.StorageAccountName + i);
    ...
}
```

skofman1:
1. This job can be made much simpler and more general: a job that receives a source storage account and a destination storage account, and copies a container (or a list of containers) from one to the other. This way it can be reused in the future for anything else we want to backup.

2. We usually add new jobs to NuGet.Jobs repository. A reason to add a job to NuGet.Services.Metadata is if it depends heavily on classes in this repository, which this job doesn't. Also, Xavier added jobs to NuGet.Jobs that can be ran using a windows scheduler. This is an option for this job as well, since we would prefer to run it at known times.

cristinamanum:
These changes are moved to Nuget.Jobs. The feedback is implemented in the new pull request.

src/Ng/Jobs/BackupV3Storage.cs
chenriksson: It uses the same log file per container each time, so only the file needs to be cleaned up?

src/Ng/Jobs/BackupV3Storage.cs
```diff
@@ -0,0 +1,236 @@
...
+
+            deleteTasks.Add(TryDeleteContainerAsync(_luceneStorageContainer, currentDate, _destStorageAccountName, _destStorageKeyValue));
+            deleteTasks.Add(TryDeleteContainerAsync(_catalogStorageContainer, currentDate, _destStorageAccountName, _destStorageKeyValue));
+            deleteTasks.Add(TryDeleteContainerAsync(_registrationStorageContainer, currentDate, _destStorageAccountName, _destStorageKeyValue));
+            deleteTasks.Add(TryDeleteContainerAsync(_registrationgzStorageContainer, currentDate, _destStorageAccountName, _destStorageKeyValue));
+
+            await TryCopyContainerAsync(_luceneStorageContainer, currentDate, _luceneStorageAccountName, _luceneStorageKeyValue);
+            await TryCopyContainerAsync(_catalogStorageContainer, currentDate, _catalogStorageAccountName, _catalogStorageKeyValue);
+            await TryCopyContainerAsync(_registrationStorageContainer, currentDate, _registrationStorageAccountName, _registrationStorageKeyValue);
+            await TryCopyContainerAsync(_registrationgzStorageContainer, currentDate, _registrationStorageAccountName, _registrationStorageKeyValue);
```
chenriksson: Does AzCopy copy directly from storage to storage? If so, could the copies be done in parallel?

src/Ng/Jobs/BackupV3Storage.cs
```diff
@@ -0,0 +1,236 @@
...
+
+        protected override async Task RunInternal(CancellationToken cancellationToken)
+        {
+            List<Task> deleteTasks = new List<Task>();
+            DateTime currentDate = DateTime.UtcNow;
+
+            deleteTasks.Add(TryDeleteContainerAsync(_luceneStorageContainer, currentDate, _destStorageAccountName, _destStorageKeyValue));
+            deleteTasks.Add(TryDeleteContainerAsync(_catalogStorageContainer, currentDate, _destStorageAccountName, _destStorageKeyValue));
+            deleteTasks.Add(TryDeleteContainerAsync(_registrationStorageContainer, currentDate, _destStorageAccountName, _destStorageKeyValue));
+            deleteTasks.Add(TryDeleteContainerAsync(_registrationgzStorageContainer, currentDate, _destStorageAccountName, _destStorageKeyValue));
```
chenriksson: There's a little inconsistency w/ some APIs taking destination storage info, and others assuming it. For instance, TryDelete and TryCreate have destination storage args, whereas TryCopy doesn't. Maybe just name all to clarify that they work against dest storage, and only take in the dest container name?

src/Ng/Jobs/BackupV3Storage.cs
```diff
@@ -0,0 +1,236 @@
...
+            catch(Exception ex)
+            {
+                Logger.LogError($"BackupV3Storage:Exception on create container {containerName}.{ex.ToString()}.");
+                return false;
+            }
+        }
+
+        private async Task<bool> TryDeleteContainerAsync(string containerName, DateTime date, string storageAccountName, string storageAccountKey)
+        {
+            string destContainer = $"{Format(date.AddDays(-1* _backupDays))}-{containerName}";
```
chenriksson: Should we try to delete any older than _backupDays? For instance, if we decrease _backupDays during a deployment, we'd need to manually clean up the days that aren't exactly 14 days back?

src/Ng/Jobs/BackupV3Storage.cs
```diff
@@ -0,0 +1,236 @@
...
+            stream.Close();
+        }
+
+        private async Task<bool> TryCreateDestinationContainerAsync(string containerName, string storageAccountName, string storageAccountKey)
+        {
+            CloudBlobClient blobClient = GetCloudBlobClient(storageAccountName, storageAccountKey);
+
+            try
+            {
+                CloudBlobContainer container = blobClient.GetContainerReference(containerName);
```
chenriksson: GetContainerReference doesn't make storage requests, so shouldn't need to be in try block. Could always include this as part of the GetCloudBlobClient helper.

src/Ng/Jobs/BlobSnapshotJob.cs
```diff
@@ -0,0 +1,90 @@
...
+    public class BlobSnapshotJob : NgJob
+    {
+        private string _connectionString;
+        private string _container;
+        private bool _verbose;
+
+        public BlobSnapshotJob(ILoggerFactory loggerFactory) : base(loggerFactory)
+        {
+
+        }
```
chenriksson: nit: remove empty line

src/Ng/Jobs/BlobSnapshotJob.cs
```diff
@@ -0,0 +1,90 @@
...
+            sw.Start();
+            Parallel.ForEach(blobList, (item) =>
+            {
+                try
+                {
+                    CloudBlockBlob blob = item as CloudBlockBlob;
+                    if (blob != null)
+                    {
+                        //list the snapshots and take a snapshot only if the blob does not have any snapshot
+                        var expandedList = container.ListBlobs(prefix: blob.Name, useFlatBlobListing: true, blobListingDetails: BlobListingDetails.Snapshots).ToList();
```
chenriksson: Can you list snapshots once upfront instead of once per blob?

cristinamanum: Not sure that i understand. The snapshots are at the blob level. 

PR: https://github.com/NuGet/NuGet.Jobs/pull/151

src/Ng/Jobs/BlobSnapshotJob.cs
cristinamanum: Updated.
PR: https://github.com/NuGet/NuGet.Jobs/pull/151

src/Ng/Jobs/BackupV3Storage.cs
cristinamanum: Added support to retain only fix number of containers . 
PR: https://github.com/NuGet/NuGet.Jobs/pull/151

src/Ng/Jobs/BackupV3Storage.cs
cristinamanum: Added to the helper.
PR: https://github.com/NuGet/NuGet.Jobs/pull/151

src/Ng/Jobs/BackupV3Storage.cs
```diff
@@ -0,0 +1,236 @@
...
+                    copyToAzureProcLog.Arguments = $"/c {AzCopyPath} {argumentsLog}";
+                    copyToAzureProcLog.CreateNoWindow = false;
+                    copyToAzureProcLog.UseShellExecute = false;
+                    Process pLog = Process.Start(copyToAzureProcLog);
+                    pLog.WaitForExit();
+                    pLog.Close();
+                }
+                catch (Exception ex)
+                {
+                    Logger.LogError($"BackupV3Storage:Exception on log save.{ex.ToString()}.");
```
cristinamanum: The AzCopy logs are saved to the Azure storage. They should contain all the info needed for detailed debugging.
PR: https://github.com/NuGet/NuGet.Jobs/pull/151

src/Ng/Jobs/BackupV3Storage.cs
```diff
@@ -0,0 +1,236 @@
...
+                copyToAzureProc.RedirectStandardOutput = true;
+#endif
+
+                    Logger.LogInformation($"StartContainerCopy:{container}");
+                    sw.Start();
+                    Process p = Process.Start(copyToAzureProc);
+                    p.WaitForExit();
+#if Debug
+                string result = p.StandardOutput.ReadToEnd();
+                Console.Write(result);
```
cristinamanum: Added
PR: https://github.com/NuGet/NuGet.Jobs/pull/151

src/Ng/Jobs/BackupV3Storage.cs
cristinamanum: We need a staged copy. We decided on a specific order. 
PR: https://github.com/NuGet/NuGet.Jobs/pull/151

src/Ng/Jobs/BackupV3Storage.cs
cristinamanum: PR. https://github.com/NuGet/NuGet.Jobs/pull/151

src/Ng/Jobs/BackupV3Storage.cs
```diff
@@ -0,0 +1,236 @@
...
+            _destStorageAccountName = arguments.GetOrThrow<string>(Arguments.DestStorageAccountName);
+            _destStorageKeyValue = arguments.GetOrThrow<string>(Arguments.DestStorageKeyValue);
+
+            _backupDays = arguments.GetOrDefault<int>(Arguments.BackupDays, DefaultBackupDays);
+        }
+
+        protected override async Task RunInternal(CancellationToken cancellationToken)
+        {
+            List<Task> deleteTasks = new List<Task>();
+            DateTime currentDate = DateTime.UtcNow;
```
cristinamanum: Updates to use var 
PR: https://github.com/NuGet/NuGet.Jobs/pull/151

cristinamanum:
The change was moved to NuGet.Jobs:
PR: https://github.com/NuGet/NuGet.Jobs/pull/151

Close this PR

PR #186, opened 2017-05-19
Update test.ps1 to support CatalogTests.dll

scottbommarito:
`CatalogTests` was changed from an `exe` to a `dll`.


skofman1:
Thanks@!@!

PR #187, opened 2017-05-19
Forcing HTTPS for the Search service

agr:
[#3975](https://github.com/NuGet/NuGetGallery/issues/3975)

* Added the configuration for forcing the HTTPS;
* Added the `ForceSslMiddleware` from `NuGet.Services.Owin`;
* Added functional tests.


src/NuGet.Services.BasicSearch/Startup.cs
```diff
@@ -73,6 +73,12 @@ public class Startup
 
             _logger.LogInformation(LogMessages.AppStartup);
 
+            // redirect all HTTP requests to HTTPS
+            if (config.RequireSsl)
```
joelverhagen: Looks great, nice work!

skofman1:
:shipit:

PR #188, opened 2017-05-19
Patch version correctly and include referenced projects in pack

scottbommarito:
Fixing the issue that `NuGet.Services.Metadata.Catalog.Monitoring` did not have a dependency on `NuGet.Services.Metadata.Catalog` despite having a project reference.


build.ps1
```diff
@@ -68,6 +68,20 @@ Invoke-BuildStep 'Restoring solution packages' { `
     } `
     -skip:$SkipRestore `
     -ev +BuildErrors
+    
+Invoke-BuildStep 'Set version metadata in AssemblyInfo.cs' {
+        $assemblyInfos = `
+            "src\NuGet.Indexing\Properties\AssemblyInfo.g.cs", `
```
chenriksson: Should this use PSScriptRoot like Gallery's build, in case it's not run from the script directory?

scottbommarito: fixed

PR #189, opened 2017-05-19
V3 monitoring job: Add option to only log [Warning] or [Error] logs from NuGet.Common.ILogger

scottbommarito:
Currently, there is an excessive amount of [Information] logs coming`NuGet.Common.ILogger`. There is so much that it drowns out the actual useful output--which packages have been validated. This PR provides an option to only log [Warning] or [Error] logs from `NuGet.Common.ILogger` which will greatly improve the readability of the logging.


skofman1:
The best practice when you want to introduce a min log level is to setup your internal logger with this setting.

scottbommarito:
@skofman1 Currently the monitoring job output and the `NuGet.Common` output use the same `ILoggerFactory`, so I will replace this PR with one that splits the implementation into two factories.

PR #190, opened 2017-05-20
V3 monitoring job: use separate ILoggerFactory for NuGet.Common logging

scottbommarito:
Because `NuGet.Common` output is so verbose, we need to be able to use a different min log level for it. By using a separate `ILoggerFactory` for its logging, we can control the min log level separately from the output of the V3 monitoring job itself.


src/Ng/Jobs/EndpointMonitoringJob.cs
```diff
@@ -22,7 +22,7 @@ public class EndpointMonitoringJob : LoopingNgJob
 
         public EndpointMonitoringJob(ILoggerFactory loggerFactory) : base(loggerFactory)
         {
-            _collectorFactory = new ValidationCollectorFactory(loggerFactory);
+            _collectorFactory = new ValidationCollectorFactory(loggerFactory, null);
```
joelverhagen: >null [](start = 78, length = 4)

name null params

joelverhagen:
There is a better way to do this. Passing around two factories of the same type should be a sniff test that something is wrong. Ideally a single factory should handle scenarios where you need different sorts of things being constructed.
[https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging#log-filtering](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging#log-filtering)

What is an example of "so verbose"?

Where will the two log levels be controlled? I don't see that code.

joelverhagen:
:clock1:

scottbommarito:
@joelverhagen 

It is currently "so verbose" that 90-95% of the log messages are "GET https://...", which means the client APIs made a web request. I think at some point it will be worth investigating why it makes so many logs but for now I would just like to filter those messages out and only show warnings and errors.

In a previous version of this fix I had the user pass the min log level of client APIs they want to emit, but @skofman1 disagreed, asserting that the job should be completely agnostic of the logging. Using a single `ILoggerFactory`, with log filtering, would break this principal, as the job would accept a single `ILoggerFactory` and then apply a filter (either hardcoded or specified by the user).

Personally, I don't have a problem with that, but I would like to hear @skofman1's opinion on it.

joelverhagen:
I would argue that a second `ILoggerFactory` is just another way of dealing with logging... and is not completely agnostic. Either you pass around two logger factories or you pass around a logger factory and a  client API log level. Either way you are controlling log levels for one source differently than another.

I'm just saying that Microsoft.Extensions.Logging has a built-in way of doing this with a single `ILoggerFactory`. If you apply a category to log messages generated by `CommonLogger`, you can filter out specific levels of logs just for that category using `FilterLoggerSettings` (a type in `Microsoft.Extensions.Logging`). When initializing your `ILoggerFactory` you will configure these different levels of logs that you want to see.

scottbommarito:
@joelverhagen will look into `FilterLogSettings`--this PR is no longer necessary.

PR #193, opened 2017-05-23
Tweaked Search Service's Successful Request Telemetry Logic

loic-sharma:
The Search Service marked `400` and `404` response codes as unsuccessful requests on Application Insights, which causes lots of unwanted noise. This adds the `TelemetryResponseCodeProcessor` to the Search Service's Application Insights configuration to mark these responses' telemetry as successful.


src/NuGet.Services.BasicSearch/ApplicationInsights.config
```diff
@@ -5,6 +5,12 @@
     <Add Type="Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.AdaptiveSamplingTelemetryProcessor, Microsoft.AI.ServerTelemetryChannel">
       <MaxTelemetryItemsPerSecond>5</MaxTelemetryItemsPerSecond>
     </Add>
+    <Add Type="NuGet.Services.Logging.TelemetryResponseCodeProcessor, NuGet.Services.Logging">
```
skofman1: This file resets every time you update AI nuget package version. Can we move the changes to code?

loic-sharma: Ah, I didn't know that. That sounds good to me.

PR #194, opened 2017-05-27
Fixed RoleInstance log value on Search Service

loic-sharma:
The search service uses ServerCommon's Application Insights helper to set the application key, which leverages the [`TelemetryContextInitializer`](https://github.com/NuGet/ServerCommon/blob/master/src/NuGet.Services.Logging/TelemetryContextInitializer.cs) to set default values for Device Id, Operating System, Cloud Role Instance and Role Name. Unfortunately, the `TelemetryContextInitializer`'s default value for Role Instance is incorrect. This change removes ServerCommon's Application Insights helper, thus allowing the `AzureRoleEnvironmentTelemetryInitializer` to set the proper default values.

Related to [Gallery's #3853](https://github.com/NuGet/NuGetGallery/issues/3853)


loic-sharma:
Blocked by #195

loic-sharma:
Peter asked that I add in the machine's name as a custom dimension to logs to maintain feature parity with what we have today.

ryuyu:
:shipit:

PR #197, opened 2017-06-02
V3 monitoring job: make RegistrationIdValidator case insensitive

scottbommarito:
Currently it's just noise because all of the `RegistrationIdValidator` failures appear to be due to casing issues.

Also added unit tests (which was non-trivial because the testing framework did not support what I was trying to test).


joelverhagen:
In a Perfect World, this case insensitivity should not be done. There is a well known ID casing per version. This is defined by the package's .nusepc. The reason this PR is necessary is the casing of IDs in V2 API (and everywhere in the gallery) is determined by the _first_ version uploaded with that ID. This is what defines the `Id` string the `PackageRegistrations` table. This field is what is JOINed into the query that produces non-hijacked V2 results (which include that original `Id` string repeated over and over for all versions of that ID).

Ideally the `Package` table would have a package ID column which has the correct casing for that specific version.

I understand this is a limitation of our database schema today, but we need to create an issue tracking this schema improvement and the corresponding revert of this PR.

Note that catalog and registration have the correct casing per version because they are derived off of the .nuspec. I think the search service is wrong because db2lucene takes the DBs value, not the .nuspec's value 😢.

Example: SQLite 3.12.2-alpha
This version's casing: SQLite
Original casing: sqlite
Catalog: https://api.nuget.org/v3/catalog0/data/2016.05.26.23.26.16/sqlite.3.12.2-alpha.json
Registration: https://api.nuget.org/v3/registration2/sqlite/index.json
Search (wrong): https://api-v2v3search-0.nuget.org/search/query?q=packageid:sqlite&ignoreFilter=true

skofman1:
@jver, what is the change you are proposing to the DB? To make queries key sensitive? To enforce the initial casing on all subsequent versions?

joelverhagen:
@skofman, I am suggesting to add a column to Packages table which has the correct ID casing for that specific version.

scottbommarito:
Issue has already been filed at https://github.com/NuGet/NuGetGallery/issues/3349. Will add this additional discussion to it.

PR #198, opened 2017-06-06
Update telemetry processors

dtivel:
Resolve #4060.


src/NuGet.Services.BasicSearch/Startup.cs
```diff
@@ -54,23 +54,25 @@ public class Startup
...
+            // Create telemetry sink
+            _searchTelemetryClient = new SearchTelemetryClient();
+
             // Add telemetry processors
             var processorChain = TelemetryConfiguration.Active.TelemetryProcessorChainBuilder;
 
             processorChain.Use(next =>
             {
-                var responseCodeProcessor = new TelemetryResponseCodeProcessor(next);
+                var responseCodeProcessor = new RequestTelemetryProcessor(next);
```
cristinamanum: The type changed but the variable name is the same. Maybe change it to align with the new type?

PR #202, opened 2017-06-12
Quick fix for Db2lucene

shishirx34:
Fixes https://github.com/NuGet/NuGetGallery/issues/3976


src/NuGet.Indexing/Sql2Lucene.cs
```diff
@@ -223,6 +223,9 @@ public static void Export(string sourceConnectionString, string destinationPath,
             var logger = loggerFactory.CreateLogger<Sql2Lucene>();
             var stopwatch = new Stopwatch();
 
+            // Save the initial time to write to the lucene index
+            var initTime = DateTime.UtcNow;
```
joelverhagen: Spoke offline about this. We can't use this timestamp. _Most of the time_ this will work out since clock skew from job machine to feed2catalog is likely less than the run time of sql2lucene. However, this is really comparing apples to oranges. Timestamp from job machine has only a weak relationship to timestamp on feed2catalog (which is what catalog2lucene cursors are based off of).

The right value to get here is the `commitTimeStamp` from this document:
https://api.nuget.org/v3/catalog0/index.json

src/NuGet.Indexing/Sql2Lucene.cs
```diff
@@ -280,5 +284,30 @@ public static void Export(string sourceConnectionString, string destinationPath,
...
         }
+
+        private static DateTime GetCommitTimestampFromCatalog(Uri indexUrl, ILogger<Sql2Lucene> logger)
+        {
+            DateTime commitTime = DateTime.UtcNow;
+            try
+            {
+                using (var client = new System.Net.Http.HttpClient())
+                {
+                    var response = client.GetAsync(indexUrl).Result;
```
joelverhagen: nit: if this must be sync, make the method async and make the caller call `.Result` on the returned task. In other words, make as much of the stack async as possible.

joelverhagen: nit: `HttpResponseMessage` should be disposable.

src/NuGet.Indexing/Sql2Lucene.cs
```diff
@@ -280,5 +284,30 @@ public static void Export(string sourceConnectionString, string destinationPath,
...
+                    logger.LogInformation("Fetching catalog index page: {0}", response.StatusCode);
+                    response.EnsureSuccessStatusCode();
+
+                    string json = response.Content.ReadAsStringAsync().Result;
+                    JObject obj = JObject.Parse(json);
+                    commitTime = obj["commitTimeStamp"].ToObject<DateTime>();
```
joelverhagen: nit: `DateTimeOffset`, unless otherwise necessary

shishirx34: It is necessary. This time is passed to `DocumentCreator` which expects a `DateTime` object.

joelverhagen: Sounds good.

shishirx34:
Whoops, just noticed I opened this PR against master branch. Closing and opening one against dev.

shishirx34:
@scottbommarito saved the day. Rebased against dev.

PR #203, opened 2017-06-15
Eliminate redundant storage calls

dtivel:
Resolve NuGet/NuGetGallery#4208.

* Changed `FeedHelpers.GetCatalogProperty(...)` to `FeedHelpers.GetCatalogPropertiesAsync(...)` to batch storage reads.
* Made some changes to facilitate testing (e.g.:  change parameter types from `Storage` to `IStorage`, add protected test constructor, etc.).
* Moved FeedHelpersTests.cs from NgTests into CatalogTests, since FeedHelpers is in the catalog project.


tests/CatalogTests/Helpers/FeedHelpersTests.cs
```diff
@@ -267,10 +405,58 @@ private static IEnumerable<ODataPackage> ODataPackages
...
+            var now = DateTime.Now;
+            var datetime = new DateTime(
+                now.Year,
+                now.Month,
+                now.Day,
+                now.Hour,
+                now.Minute,
+                now.Second,
+                now.Millisecond,
+                DateTimeKind.Unspecified);
```
agr: Could be written shorter: `var datetime = new DateTime(DateTime.Now.Ticks, DateTimeKind.Unspecified);`

PR #204, opened 2017-06-19
China dnx

cristinamanum:
The script and changes for the catlog2dnx for China.

PR #209, opened 2017-06-27
Feed2catalog KeyNotFoundException hotfix

chenriksson:
Regression from https://github.com/NuGet/NuGet.Services.Metadata/pull/204


src/Ng/CommandHelpers.cs
```diff
@@ -178,7 +178,8 @@ public static StorageFactory CreateSuffixedStorageFactory(string suffix, IDictio
                 { Arguments.StorageAccountName, Arguments.StorageAccountName + suffix },
                 { Arguments.StorageKeyValue, Arguments.StorageKeyValue + suffix },
                 { Arguments.StorageContainer, Arguments.StorageContainer + suffix },
-                { Arguments.StoragePath, Arguments.StoragePath + suffix }
+                { Arguments.StoragePath, Arguments.StoragePath + suffix },
+                { Arguments.StorageSuffix, Arguments.StorageSuffix + suffix }
```
scottbommarito: IMO we should rename `Arguments.StorageSuffix` to `Arguments.StorageEndpointSuffix`.

Having an argument named `StorageSuffix` is confusing given that we have a method named `CreateSuffixedStorageFactory`.

cristinamanum: Yes, I think that would be better to change it. I can do that in a future change.

PR #211, opened 2017-06-28
Catalog2Reg for China

cristinamanum:
Changes to allow registration blobs to be added to China.

There are two changes in this PR:
1. The registration blobs will point to flat container instead of packages. (for this change FlatContainerPackagePathProvider was introduced)
2. The catalog2Reg waits for the catalog2dnx (for this now the job takes an extra parameter for the flatcontainer cursor)


src/Catalog/CatalogItem.cs
```diff
@@ -10,11 +10,25 @@ namespace NuGet.Services.Metadata.Catalog
 {
     public abstract class CatalogItem
     {
+        Uri _baseAddress;
```
skofman1: private

src/Catalog/CatalogWriterBase.cs
```diff
@@ -148,14 +154,16 @@ protected virtual ResourceSaveOperation CreateSaveOperationForItem(IStorage stor
...
+            var catalogItemUri = item.GetItemAddress();
 
             var operation = new ResourceSaveOperation();
-            operation.ResourceUri = resourceUri;
+            operation.ResourceUri = catalogItemUri;
 
             if (content != null)
             {
-                operation.SaveTask = storage.Save(resourceUri, content, cancellationToken);
+                //the save operation will not be done through the customDomain but directly through the storage
```
skofman1: nit: capital letter in the beginning of comment, like in the rest of the method

src/Catalog/CatalogItem.cs
```diff
@@ -37,5 +51,23 @@ public virtual IGraph CreateContentGraph(CatalogContext context)
         {
             return null;
         }
+
+        /// <summary>
+        /// Changes the host of the baseAddress with the customDomain
+        /// </summary>
+        /// <param name="baseAddress"></param>
```
skofman1: nit: remove empty documentation tags

src/Catalog/CatalogItem.cs
```diff
@@ -37,5 +51,23 @@ public virtual IGraph CreateContentGraph(CatalogContext context)
...
+        /// <param name="customDomain"></param>
+        /// <returns></returns>
+        public static  Uri AliasesAddress(Uri baseAddress, string customDomain)
```
skofman1: I would suggest a clearer name: RelaceHostForUri or something..

src/Catalog/CatalogItem.cs
```diff
@@ -10,11 +10,25 @@ namespace NuGet.Services.Metadata.Catalog
...
         public DateTime TimeStamp { get; set; }
 
         public Guid CommitId { get; set; }
 
-        public Uri BaseAddress { get; set; }
+        public Uri BaseAddress
+        {
+            get
+            {
+                return AliasesAddress(_baseAddress, CustomDomain);
```
skofman1: every time BaseAddress is called AliasesAddress will be recalculated. To improve perf you can calculate this ones, on _baseAddress set

src/Catalog/Registration/RegistrationMaker.cs
```diff
@@ -20,7 +20,8 @@ public static class RegistrationMaker
             Uri contentBaseAddress,
             int partitionSize,
             int packageCountThreshold,
-            CancellationToken cancellationToken)
+            CancellationToken cancellationToken,
+            string customDomain)
```
skofman1: nit: the convention in the code is that CancellationToken is the last parameter

src/Catalog/Registration/RegistrationMakerCatalogWriter.cs
```diff
@@ -86,18 +91,21 @@ protected override ResourceSaveOperation CreateSaveOperationForItem(IStorage sto
             if (registrationMakerCatalogItem != null)
             {
                 var content = item.CreateContent(Context); // note: always do this first
-                var resourceUri = item.GetItemAddress();
+                var itemAddress = item.GetItemAddress();
+                //the storage will indicate the location where the content will be saved
```
skofman1: nit: same comment for capital letter for comments

src/Catalog/Registration/RegistrationPersistence.cs
```diff
@@ -23,15 +23,23 @@ public class RegistrationPersistence : IRegistrationPersistence
         private readonly RecordingStorage _storage;
         private readonly Uri _registrationBaseAddress;
         private readonly Uri _contentBaseAddress;
-
-        public RegistrationPersistence(StorageFactory storageFactory, RegistrationKey registrationKey, int partitionSize, int packageCountThreshold, Uri contentBaseAddress)
+        private readonly string _customDomain ;
```
skofman1: nit: space

src/Catalog/CatalogWriterBase.cs
```diff
@@ -18,8 +18,13 @@ public abstract class CatalogWriterBase : IDisposable
     {
         protected List<CatalogItem> _batch;
         protected bool _open;
+        protected string _customDomain;
 
-        public CatalogWriterBase(IStorage storage, ICatalogGraphPersistence graphPersistence = null, CatalogContext context = null)
+        public CatalogWriterBase(
```
skofman1: do those changes impact Feed2Catalog job?

src/Catalog/CatalogItem.cs
skofman1: I didn't see any UTs for this method.

src/Catalog/CatalogWriterBase.cs
scottbommarito: also space between the text of the comment and the `//`

for example: `// The save...`

src/Catalog/CatalogWriterBase.cs
scottbommarito: I was confused at first as well, but `RegistrationMakerCatalogWriter` extends this class.

src/Catalog/Registration/FlatContainerPackagePathProvider.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace NuGet.Services.Metadata.Catalog.Registration
+{
+    public class FlatContainerPackagePathProvider : IPackagePathProvider
+    {
+        public string GetPackagePath(string id, string version)
+        {
+            version = NuGetVersionUtility.NormalizeVersion(version);
+
+            return $"v3-flatcontainer/{id.ToLowerInvariant()}/{version.ToLowerInvariant()}/{id.ToLowerInvariant()}.{version.ToLowerInvariant()}.nupkg";
```
scottbommarito: This should use the V3 index.json to determine the base address of the flat-container. There's not guarantee that we will always use `v3-flatcontainer` or that we will only have a single flat-container.

src/Ng/Jobs/Catalog2RegistrationJob.cs
```diff
@@ -78,12 +80,24 @@ protected override void Init(IDictionary<string, string> arguments, Cancellation
                 source,
                 storageFactories.LegacyStorageFactory);
 
-            RegistrationMakerCatalogItem.PackagePathProvider = new PackagesFolderPackagePathProvider();
+            if (isContentFlatContainer)
+            {
+                var flatContainerCursorUriString = arguments.GetOrThrow<string>(Arguments.FlatContainerCursorUri);
```
scottbommarito: You can very easily calculate the flat-container's cursor uri by using the flat-container's base address. You shouldn't need to pass it in via a command line argument.

src/Catalog/CatalogWriterBase.cs
```diff
@@ -183,7 +191,7 @@ protected virtual StorageContent CreateIndexContent(IGraph graph, Uri type)
         {
             IGraph graph = new Graph();
 
-            INode resourceNode = graph.CreateUriNode(resourceUri);
+            INode resourceNode = graph.CreateUriNode(CatalogItem.AliasesAddress(resourceUri, _customDomain));
```
scottbommarito: These changes seem very scary to me.

Basically, in the past, all had were resource uris with the same host, so we passed these uris around on their own. Now, however, we care about the host, and so we have added tons of places where we are replacing the host.

If we missed a place in which the host should have been replaced, we could very easily get into a scenario where some uri somewhere is being written with an incorrect host. Additionally, in the future, if we want to provide some additional modifications to the uri, we could potentially need to modify each of these places again, and could again miss.

I think we should rethink this design. I would much rather see something in which we stop passing around uris and start passing around a `ResourceIdentity` object that we can ask for the uri of which would perform the `AliasesAddress` transform. By changing interfaces and methods that previously used uris to `ResourceIdentity`s, we can guarantee through strong typing that the uris are being transformed when expected.

I understand doing that would probably be a much bigger change, but I think it would be a lot more stable in the long run.

src/Ng/Jobs/LightningJob.cs
```diff
@@ -401,7 +442,7 @@ private void InitStrike(IDictionary<string, string> arguments)
                             if (graph.Nodes.Count() > 3000 && sortedGraphs.Count >= 20)
                             {
                                 // Process graphs
-                                await ProcessGraphsAsync(packageId, sortedGraphs);
+                                await ProcessGraphsAsync(packageId, sortedGraphs, _customDomain);
```
scottbommarito: You shouldn't need to pass `_customDomain` around, given that it's a field on the object.

src/Catalog/CatalogItem.cs
scottbommarito: Given that you are using this method constantly outside of the context of the catalog, I would move this to another shared static class.

src/Catalog/CatalogItem.cs
cristinamanum: Done

src/Catalog/CatalogWriterBase.cs
cristinamanum: Done

src/Catalog/CatalogItem.cs
cristinamanum: Done

src/Catalog/CatalogItem.cs
cristinamanum: Done

src/Catalog/CatalogItem.cs
cristinamanum: It will be a small perf impact. If changed in the suggested way then the set of the base address needs to happen after the set of custom domain. In the current way the get will use the data at the current time.

src/Catalog/Registration/RegistrationMaker.cs
cristinamanum: Done

src/Catalog/Registration/RegistrationMakerCatalogWriter.cs
cristinamanum: Done

src/Catalog/Registration/RegistrationPersistence.cs
cristinamanum: Not sure what space i need to remove?

src/Catalog/CatalogWriterBase.cs
cristinamanum: The feed2catalog will not be impacted by these changes as long as AppendOnlyCatalogWriter (used by feed2catalog) will not subscribe to use a customDomain.
By default customDomain is null and the ReplaceHostForUri is noop for this case. I will go some feed2catalog more testing.

src/Catalog/CatalogItem.cs
cristinamanum: Good point. Added.

src/Catalog/CatalogWriterBase.cs
cristinamanum: We discussed offline and I agree that the changes could be risky.  I tried to be as less intrusive as possible  using the current design. The mental model that I followed was the separation of concerns between the storage and content. The host links in content (generated by the writers) can be different by the storage host. We can do more radical design changes but  that it will add to the time and risk.

src/Catalog/Registration/FlatContainerPackagePathProvider.cs
cristinamanum: Good suggestion to not have the v3-flatcontainer hard-coded. The code was updated to use a parameter for it.

src/Ng/Jobs/Catalog2RegistrationJob.cs
cristinamanum: The cursor is calculated now, but the flatcontainer uri is still passed as argument. I preferred this way instead of passing the index.json uri , open and read the flat container from there.

src/Ng/Jobs/LightningJob.cs
cristinamanum: Right... updated.

src/Catalog/CatalogItem.cs
cristinamanum: It is used in the catalog content generation. IMO is to have it here or catalogwriter base. I added in the writerbase. If any strong preference let me know.

src/Catalog/Registration/RegistrationPersistence.cs
scottbommarito: after the variable name
`_customDomain ;` -> `_customDomain;`

src/Catalog/Registration/FlatContainerPackagePathProvider.cs
scottbommarito: We came to the conclusion that it would be best not to use the V3 `index.json` because it's now important to have the registration blobs use the same flat-container, and using the `index.json` could lead to a mismatch.

src/Catalog/CatalogWriterBase.cs
cristinamanum: Manual tested the feed2catalog job.

src/Catalog/Registration/RegistrationPersistence.cs
cristinamanum: Done

src/Ng/lightning-template.txt
```diff
@@ -6,7 +6,7 @@ title [cursorFile]
...
+start /w ng.exe lightning -command strike -indexFile "[indexFile]" -cursorFile "[cursorFile]" -contentBaseAddress "[contentBaseAddress]" -verbose "[verbose]" -storageBaseAddress "[storageBaseAddress]" -storageAccountName "[storageAccountName]" -storageKeyValue "[storageKeyValue]" -storageContainer "[storageContainer]" -useCompressedStorage "[useCompressedStorage]" -compressedStorageBaseAddress "[compressedStorageBaseAddress]" -compressedStorageAccountName "[compressedStorageAccountName]" -compressedStorageKeyValue "[compressedStorageKeyValue]" -compressedStorageContainer "[compressedStorageContainer]" -useSemVer2Storage "[useSemVer2Storage]" -semVer2StorageBaseAddress "[semVer2StorageBaseAddress]" -semVer2StorageAccountName "[semVer2StorageAccountName]" -semVer2StorageKeyValue "[semVer2StorageKeyValue]" -semVer2StorageContainer "[semVer2StorageContainer]" [optionalArguments]
```
joelverhagen: We should format this such that each kvp is on its own line

joelverhagen: Like the file above 😄

src/Catalog/Registration/RegistrationPersistence.cs
```diff
@@ -200,24 +253,53 @@ static void AddExistingItem(IDictionary<RegistrationEntryKey, RegistrationCatalo
...
             Trace.TraceInformation("RegistrationPersistence.SaveLargeRegistration: registrationBaseAddress = {0} items: {1}", registrationBaseAddress, items.Count);
 
             IList<Uri> cleanUpList = new List<Uri>();
 
-            await SaveRegistration(storage, registrationBaseAddress, items, cleanUpList, null, partitionSize, contentBaseAddress, cancellationToken);
+            await SaveRegistration(storage,
+                                   registrationBaseAddress,
+                                   items,
+                                   cleanUpList,
+                                   null,
```
joelverhagen: name nulls

src/Catalog/CatalogWriterBase.cs
```diff
@@ -217,7 +241,7 @@ protected virtual StorageContent CreateIndexContent(IGraph graph, Uri type)
 
             if (parent != null)
             {
-                graph.Assert(resourceNode, graph.CreateUriNode(Schema.Predicates.CatalogParent), graph.CreateUriNode(parent));
+                graph.Assert(resourceNode, graph.CreateUriNode(Schema.Predicates.CatalogParent), graph.CreateUriNode(ReplaceHostForUri(parent,_customDomain)));
```
joelverhagen: nit space after comma

src/Ng/Jobs/LightningJob.cs
```diff
@@ -126,6 +133,7 @@ public override string GetUsage()
         private RegistrationStorageFactories _storageFactories;
         private ShouldIncludeRegistrationPackage _shouldIncludeSemVer2;
         private IDictionary<string, string> _arguments;
+        private string _customDomain;
```
joelverhagen: Unused?

tests/NgTests/RegistrationCollectorTests.cs
```diff
@@ -12,6 +12,9 @@
 using NgTests.Infrastructure;
 using NuGet.Services.Metadata.Catalog;
 using NuGet.Services.Metadata.Catalog.Registration;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
```
joelverhagen: Are these changes necessary?

src/Ng/Jobs/LightningJob.cs
```diff
@@ -147,19 +155,27 @@ protected override void Init(IDictionary<string, string> arguments, Cancellation
...
             arguments[Arguments.StorageType] = Arguments.AzureStorageType;
 
-            // Enable greater HTTP parallization.
+            // Enable greater HTTP parallelization.
             ServicePointManager.DefaultConnectionLimit = CatalogIndexReader.MaxDegreeOfParallelism;
 
             // Configure the package path provider.
-            RegistrationMakerCatalogItem.PackagePathProvider = new PackagesFolderPackagePathProvider();
+            var useFlatContainerAsPackageContent = arguments.GetOrDefault<bool>(Arguments.ContentIsFlatContainer, false);
+            if (!useFlatContainerAsPackageContent)
```
chenriksson: nit, would reverse the if/else so you don't need the '!'

src/Ng/Jobs/LightningJob.cs
```diff
@@ -73,6 +73,13 @@ public override string GetUsage()
             sw.WriteLine($"      Compressed only: Azure Storage account name.");
             sw.WriteLine($"  -{Arguments.CompressedStorageContainer} <container>");
             sw.WriteLine($"      Compressed only: Container to generate registrations in.");
+            sw.WriteLine($"  -{Arguments.ContentIsFlatContainer} true|false");
+            sw.WriteLine($"      Boolean to indicate if the registration blobs will have the content from the packages or flat container.");
+            sw.WriteLine($"  -{Arguments.FlatContainerName} <the flat container name>");
+            sw.WriteLine($"      It is required when the ContentIsFlatContainer flag is true.");
```
chenriksson: Assuming ContentIsFlatContainer is only necessary so we can have hard-coded default for FlatContainerName?

src/Ng/lightning-template.txt
cristinamanum: Done

src/Ng/Jobs/LightningJob.cs
cristinamanum: Updated

tests/NgTests/RegistrationCollectorTests.cs
cristinamanum: Not needed. Updated.

src/Catalog/Registration/FlatContainerPackagePathProvider.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            _container = container;
+        }
+        public string GetPackagePath(string id, string version)
+        {
+            version = NuGetVersionUtility.NormalizeVersion(version);
+
+            return $"{_container}/{id.ToLowerInvariant()}/{version.ToLowerInvariant()}/{id.ToLowerInvariant()}.{version.ToLowerInvariant()}.nupkg";
+        }
+
```
cristinamanum: Done

src/Ng/Jobs/LightningJob.cs
cristinamanum: It is passed as argument in order to be able to change the flatContainer container name.

PR #213, opened 2017-07-09
Registration3 script files

cristinamanum:
Add the script files for  registration3 (nupkg links are from flat container instead of packages).


src/Ng/Scripts/Catalog2RegistrationV3-Reg3.cmd
```diff
@@ -0,0 +1,40 @@
...
+    -clientId #{Deployment.Azure.KeyVault.ClientId} ^
+    -certificateThumbprint #{Deployment.Azure.KeyVault.CertificateThumbprint} ^
+    -verbose true ^
+    -interval #{Jobs.common.v3.Interval} ^
+    -storageBaseAddress #{Jobs.catalog2registrationv3reg3.StorageBaseAddress} ^
+    -storageAccountName #{Jobs.common.v3.c2r.StorageAccountName} ^
+    -storageKeyValue #{Jobs.common.v3.c2r.StorageAccountKey} ^
+    -storageContainer #{Jobs.catalog2registrationv3reg3.StorageContainer} ^
+    -useCompressedStorage #{Jobs.catalog2registrationv3reg3.UseCompressedStorage} ^
+    -compressedStorageBaseAddress #{Jobs.catalog2registrationv3reg3.StorageBaseAddressCompressed} ^
```
joelverhagen: Do we really need to have the `UseSemVer2Storage` etc parameterized?

cristinamanum: It is the same structure as the registration2 script. I would let it the same for now and we can change it later.

PR #217, opened 2017-07-19
catlog2lucene waits for one or more registration cursors

cristinamanum:
In order to support multiple regions registrations and one search service force catalog2Lucene to wait for all registrations.  The cursor used to advance catalog2lucene will have the value the min of all passed cursors.


src/Ng/Jobs/Catalog2LuceneJob.cs
```diff
@@ -107,6 +107,13 @@ protected override void Init(IDictionary<string, string> arguments, Cancellation
             }
         }
 
+        ReadCursor GetTheLeastAdvancedRegistrationCursor(string registrationArg, CancellationToken cancellationToken)
```
joelverhagen: `private`

src/Ng/Jobs/Catalog2LuceneJob.cs
```diff
@@ -107,6 +107,13 @@ protected override void Init(IDictionary<string, string> arguments, Cancellation
...
+        {
+            string[] registrations = registrationArg.Split(';');
```
scottbommarito: Any reason we can't just query index.json and get all of the registrations from there?

skofman1:
:shipit:

src/Ng/Jobs/Catalog2LuceneJob.cs
cristinamanum: Updated

src/Ng/Jobs/Catalog2LuceneJob.cs
cristinamanum: Updated

cristinamanum: There will be one registration in index.json ; its location will be masked by a dns. The job will need to read the cursor blob location.

scottbommarito: After offline discussion, I'll file an issue regarding how we can harden our services to change with index.json changes.

PR #219, opened 2017-07-20
LoopingJobFatalException

cristinamanum:
Looping job will log fatal exception before ending.


src/Ng/Jobs/LoopingNgJob.cs
```diff
@@ -48,6 +48,11 @@ protected LoopingNgJob(ILoggerFactory loggerFactory)
                 {
                     Logger.LogWarning("Job execution ended due to a transient exception. Exception: {1}, inner exception: {2}.", te, te.InnerException);
                 }
+                catch (Exception ex)
+                {
+                    Logger.LogError("Job execution ended due to a fatal exception. Exception: {1}, inner exception: {2}.", ex, ex.InnerException);
```
joelverhagen: Is this logging to `TrackException`? If not can we track an issue? (there might already be  one... not sure)

cristinamanum:
The ng.exe catches the unhandled exception and logs it. Close the PR without merge.

PR #221, opened 2017-07-31
Added missing exceptions to Search Service's logs

loic-sharma:
Fixes [#4473](https://github.com/NuGet/NuGetGallery/issues/4473)


src/NuGet.Services.BasicSearch/Startup.cs
```diff
@@ -246,7 +246,7 @@ private bool InitializeSearcherManager(IndexingConfiguration config, Directory d
                     shouldRetry: e =>
                     {
                         // Retry on any exception (but log it)
-                        _logger.LogError("Startup: An error occurred initializing searcher manager. Going to retry...",
+                        _logger.LogError("Startup: An error occurred initializing searcher manager. Going to retry... Exception: {Exception}",
                             e);
                         _searchTelemetryClient.TrackMetric(SearchTelemetryClient.MetricName.SearchIndexReopenFailed, 1);
```
joelverhagen: Does this write exceptions to the `exceptions` AI table? I thought you had to call `TrackException` or something like that.

loic-sharma: This goes to AI's trace table due to the configuration set up by ServerCommon on this [line](https://github.com/NuGet/NuGet.Services.Metadata/blob/0fc73fd1f138abce259e6e40ef98def52cd1a992/src/NuGet.Services.BasicSearch/Startup.cs#L86). From local testing, it does not go to the `exceptions` table though.

loic-sharma: Opened [this issue](https://github.com/NuGet/NuGetGallery/issues/4499) to duplicate exception logs as traces.

PR #222, opened 2017-07-31
Fix attempt for hangs in CloudBlockBlob.UploadFromStreamAsync and CloudBlockBlob.DownloadToStreamAsync 

cristinamanum:
Intermittent hangs have been observed in c2registration and c2dnx due to hangs in CloudBlockBlob.UploadFromStreamAsync and CloudBlockBlob.DownloadToStreamAsync calls. 

Per Azure Storage team recommendation add execution max timeout option for azure storage operations. Also update the version of the AzureStorage SDK from 7.0.0 to 8.2.1.

The current jobs do not need to use the new setting for max execution timeout. The jobs where the hang happened are for the china pipeline. They will be updated with the new parameter.


src/Ng/CommandHelpers.cs
```diff
@@ -119,7 +119,8 @@ public static StorageFactory CreateStorageFactory(IDictionary<string, string> ar
                 { Arguments.StorageKeyValue, Arguments.StorageKeyValue },
                 { Arguments.StorageContainer, Arguments.StorageContainer },
                 { Arguments.StoragePath, Arguments.StoragePath },
-                { Arguments.StorageSuffix, Arguments.StorageSuffix }
+                { Arguments.StorageSuffix, Arguments.StorageSuffix },
+                { Arguments.StorageOperationMaxExecutionTimeInSeconds, Arguments.StorageOperationMaxExecutionTimeInSeconds },
```
joelverhagen: Does this really need to be a parameter? If we make this a simple const in the code, I think it will be a lot easier. Asked another way, when would one environment need a different timeout than another. Do we have HTTP client timeouts set anywhere else on server side?

If it must be a parameter can we default it to a sane value then make all of the nullable ints `int?` into just `int`?

src/Ng/CommandHelpers.cs
```diff
@@ -179,7 +182,8 @@ public static StorageFactory CreateSuffixedStorageFactory(string suffix, IDictio
                 { Arguments.StorageKeyValue, Arguments.StorageKeyValue + suffix },
                 { Arguments.StorageContainer, Arguments.StorageContainer + suffix },
                 { Arguments.StoragePath, Arguments.StoragePath + suffix },
-                { Arguments.StorageSuffix, Arguments.StorageSuffix + suffix }
+                { Arguments.StorageSuffix, Arguments.StorageSuffix + suffix },
+                { Arguments.StorageOperationMaxExecutionTimeInSeconds, Arguments.StorageOperationMaxExecutionTimeInSeconds + suffix }
```
joelverhagen: nit: put a comma at the end of the last item in the dictionary initializer to avoid future diff noise (note it looks like there are two lines changes here instead of just one).

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -16,13 +17,21 @@ namespace NuGet.Services.Metadata.Catalog.Persistence
     public class AzureStorage : Storage
     {
         private readonly CloudBlobDirectory _directory;
-
-        public AzureStorage(CloudStorageAccount account, string containerName, string path, Uri baseAddress)
-            : this(account.CreateCloudBlobClient().GetContainerReference(containerName).GetDirectoryReference(path), baseAddress)
+        private readonly BlobRequestOptions _blobRequestOptions;
+        private readonly int _serverTimeoutSeconds = 30;
```
joelverhagen: `const`

src/Ng/CommandHelpers.cs
scottbommarito: IMO if it's going to be a parameter it should have a default

cristinamanum: Added a default, changed the int? to int  but kept the configuration of the maxtimeout. I think that having the be able to configure different timeouts for different environments can be beneficial.

src/Ng/CommandHelpers.cs
cristinamanum: Updated

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Updated

cristinamanum: Updated the scripts as well.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -112,21 +124,41 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
+                                                     cancellationToken: cancellationToken);
                     Trace.WriteLine(String.Format("Saved uncompressed blob {0} to container {1}", blob.Uri.ToString(), _directory.Container.Name));
                 }
             }
             await TryTakeBlobSnapshotAsync(blob);
         }
 
+        private BlobRequestOptions CreateBlobRequestOptions(int maxExecutionTimeInSeconds)
+        {
+            int internalMaxTimeout = maxExecutionTimeInSeconds <= 0 ? _defaultMaxTimeoutInSeconds : maxExecutionTimeInSeconds;
```
skofman1: >maxExecutionTimeInSeconds <= 0 [](start = 37, length = 30)

consider using nullable (int?) to specify no value in ctor, and throw on illegal value like maxExecutionTimeInSeconds <= 0 in the ctor

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -112,21 +124,41 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
+           
+            return new BlobRequestOptions
+            {
+                ServerTimeout = TimeSpan.FromSeconds(_serverTimeoutSeconds),
+                MaximumExecutionTime = TimeSpan.FromSeconds(internalMaxTimeout),
```
skofman1: >internalMaxTimeout [](start = 60, length = 18)

are those values identical to the azure defaults we have right now? 
I would like to make sure we don't change the timeout config for existing jobs.

src/Ng/CommandHelpers.cs
joelverhagen: Sound great!

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -15,14 +16,24 @@ namespace NuGet.Services.Metadata.Catalog.Persistence
 {
     public class AzureStorage : Storage
     {
-        private readonly CloudBlobDirectory _directory;
+        private const int _serverTimeoutSeconds = 30;
+        private const int _defaultMaxTimeoutInSeconds = 600;
```
joelverhagen: Nit could we pass a `TimeSpan` around instead of an int? In general we should use data types that represent the value as closely as possible. Leads to code that is easier to understand.

joelverhagen: How did we decide on these values? They seem like a reasonable start to me, I'm just wondering what the thinking is here.

src/Catalog/Persistence/AzureStorage.cs
joelverhagen: We spoke above about `int` vs. `int?`. I think another option is using `TimeSpan` and throwing `ArgumentOutOfRangeException` when the `TimeSpan` is negative.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -112,21 +124,41 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
...
 
+        private BlobRequestOptions CreateBlobRequestOptions(int maxExecutionTimeInSeconds)
+        {
+            int internalMaxTimeout = maxExecutionTimeInSeconds <= 0 ? _defaultMaxTimeoutInSeconds : maxExecutionTimeInSeconds;
+           
+            return new BlobRequestOptions
+            {
+                ServerTimeout = TimeSpan.FromSeconds(_serverTimeoutSeconds),
+                MaximumExecutionTime = TimeSpan.FromSeconds(internalMaxTimeout),
+                RetryPolicy = new ExponentialRetry()
```
joelverhagen: Does this just exponentially back off until `MaximumExecutionTime` is exhausted?

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Added timespan and validation for negative value arguments

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: This change will introduce a timeout as previously there was not any timeout. If timeouts will be encountered in the jobs we can either increase the default timeout or set a higher timeout on the jobs.

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Added the timespan

cristinamanum: Server timeout from samples and documentation . The max timeout it is just a guess.

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Yes

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -37,6 +55,15 @@ private AzureStorage(CloudBlobDirectory directory, Uri baseAddress)
...
             }
             ResetStatistics();
+            this._blobRequestOptions = CreateBlobRequestOptions(maxExecutionTime);
+        }
+
+        public static TimeSpan DefaultMaxExecutionTime
+        {
+            get
+            {
+                return TimeSpan.FromSeconds(600);
```
scottbommarito: Can this `TimeSpan` be a constant? Similarly to `_defaultServerTimeout`?

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -112,21 +138,39 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
                     }
 
                     destinationStream.Seek(0, SeekOrigin.Begin);
-                    await blob.UploadFromStreamAsync(destinationStream, cancellationToken);
+                    await blob.UploadFromStreamAsync(destinationStream,
+                                                     accessCondition: null,
```
scottbommarito: Why do we need to pass in `null`? Can we just not pass anything in for these parameters?

src/Ng/Jobs/Catalog2RegistrationJob.cs
```diff
@@ -57,7 +57,10 @@ public override string GetUsage()
                    + $"-{Arguments.SemVer2StorageAccountName} <azure-acc> "
                    + $"-{Arguments.SemVer2StorageKeyValue} <azure-key> "
                    + $"-{Arguments.SemVer2StorageContainer} <azure-container> "
-                   + $"-{Arguments.SemVer2StoragePath} <path>";
+                   + $"-{Arguments.SemVer2StoragePath} <path>"
+                   + Environment.NewLine
+                   + "Optional Arguments"
+                   + $"-{Arguments.StorageOperationMaxExecutionTimeInSeconds} <azureOperationMaxExecutionTimeInSeconds>";
```
scottbommarito: Technically, this is a parameter for all of our jobs that use storage accounts now.

We should add a work item to make the `GetUsage` strings generated in a more automated way.

One potential option:
```
public string GetStorageFactoryUsage()
{
    $"-{Arguments.StorageBaseAddress} <storage-base-address> " +
    $"-{Arguments.StorageType} file|azure " +
    ...
}
```

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Needs to be a compile time constant and the TimeSpan is not.

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Used the defaults for the arguments that we do not currently use.

src/Ng/Jobs/Catalog2RegistrationJob.cs
cristinamanum: Good suggestion . I will add a task.

PR #223, opened 2017-08-02
Ensure SSL v3 is disabled and that TLS v1.2 is enabled

loic-sharma:
This affects all ng.exe jobs and the Search Service. Related to [#647](https://github.com/NuGet/Engineering/issues/647).

- [X] Validate Search Service on dev
- [ ] Validate Ng.exe on dev


src/NuGet.Services.BasicSearch/Startup.cs
```diff
@@ -168,7 +168,11 @@ public void Configuration(IAppBuilder app)
             ServicePointManager.DefaultConnectionLimit = Int32.MaxValue;
             ServicePointManager.Expect100Continue = false;
             ServicePointManager.UseNagleAlgorithm = false;
-            
+
+            // Ensure that SSLv3 is disabled and that Tls v1.2 is enabled.
```
joelverhagen: Is there some place in `ServerCommon` we can put all of this `ServicePointManager` bootstrapping stuff?

loic-sharma: Talked to Joel offline. Most of the `ServicePointManager` settings are inconsistent across the Gallery, Search Service, and jobs - the only consistent setting would be the `SecurityProtocol`. Splitting this into another repository would be overkill for now, so we will keep it as it is for now.

PR #225, opened 2017-08-15
[WebScout] Disable content type sniffing

loic-sharma:
Fixes [#697](https://github.com/NuGet/Engineering/issues/697)


loic-sharma:
WebScout didn't raise any issues about this on the Gallery. I double-checked and tested a few Gallery API  endpoints - it seems that the `X-Content-Type-Options` is indeed properly set.

PR #226, opened 2017-08-22
[Prefix] Display Packages' Verified Status on Search Service

loic-sharma:
Fixes https://github.com/NuGet/NuGetGallery/issues/4382


src/NuGet.Indexing/ResponseFormatter.cs
```diff
@@ -340,6 +342,7 @@ private static void WriteDataV2(JsonWriter jsonWriter, NuGetIndexSearcher search
                 WriteDocumentValue(jsonWriter, "ReleaseNotes", document, "ReleaseNotes");
                 WriteDocumentValue(jsonWriter, "ProjectUrl", document, "ProjectUrl");
                 WriteDocumentValue(jsonWriter, "IconUrl", document, "IconUrl");
+                WriteProperty(jsonWriter, "IsVerified", isVerified);
```
shishirx34: For v2 this needs to be in `PackageRegistration` object, `WriteRegistrationV2` above. See the OneNote spec for the return result structure. Also the key should be "Verified" for V2 search result on the registration object.

src/NuGet.Indexing/VerifiedPackages.cs
```diff
@@ -0,0 +1,59 @@
...
+            reader.Read();
+            ThrowIfNotExpectedToken(reader, JsonToken.StartArray);
+
+            // Read all of the package ID strings from the array.
+            string packageId = reader.ReadAsString();
+            var result = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+            while (packageId != null)
+            {
+                result.Add(String.Intern(packageId));
```
chenriksson: Assuming String.Intern is intended to optimize lookup performance? Would the string for lookup need to be interned as well, and would the optimization work on a case insensitive collection?

Would measure if this code actually improves perf (i.e., with simple console app), and keep / add comment if it does.

tests/NuGet.Services.BasicSearchTests/V2SearchFunctionalTests.cs
```diff
@@ -6,6 +6,10 @@
 using NuGet.Services.BasicSearchTests.Models;
 using NuGet.Services.BasicSearchTests.TestSupport;
 using Xunit;
+using System.IO;
+using Newtonsoft.Json;
+using System;
```
chenriksson: nit: sort

src/NuGet.Indexing/VerifiedPackages.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System;
+using System.Collections.Generic;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+using FrameworkLogger = Microsoft.Extensions.Logging.ILogger;
+using System.IO;
+
+namespace NuGet.Indexing
+{
+    public static class VerifiedPackages
```
loic-sharma: I'm torn on this. You're correct that this could be internal, but should it? Only `NuGetSearcherManager` uses this `VerifiedPackages` class today, but this may be a useful helper class in the future. Thoughts?

src/NuGet.Indexing/VerifiedPackages.cs
loic-sharma: Interning strings is a pattern used throughout the search service. `String.Intern` is to reduce memory pressure, not to improve performance. Each reload allocates a bunch of strings (things like package IDs, versions, etc..) that already exist in memory from the previous index. By interning the string, we can reuse the memory that was allocated the first that string was encountered. The memory allocated during the index reload will then be collected on the next garbage collection. If we didn't intern the strings, the strings would be allocated on each reload and would need to be freed on the subsequent reload. Since the time between reloads is in the order of minutes, the strings would likely get promoted to the next generation of memory. Garbage collection is cheaper when memory is only used briefly.

chenriksson: Is this documented in code, so we don't accidentally remove it?

tests/NuGet.IndexingTests/VerifiedPackagesParsingTests.cs
```diff
@@ -0,0 +1,42 @@
...
+using Newtonsoft.Json;
+using NuGet.Indexing;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Xunit;
+
+namespace NuGet.IndexingTests
+{
+    public class VerifiedPackagesParsingTests
```
loic-sharma: Testing the `Load` method would require mocks that would complicate the test for little gain. Since the `Load` method is only a convenience method, I think testing the `Parse` method directly is fine. Thoughts?

chenriksson: No strong opinion here. Generally, I try to test the entry point for complete coverage.

I thought it'd just require a simple mock to have the loader return a reader... but if it's more complex, wouldn't worry about it.

src/NuGet.Indexing/NuGetSearcherManager.cs
```diff
@@ -206,7 +207,7 @@ protected override NuGetIndexSearcher CreateSearcher(IndexReader reader)
                 }
                 catch (Exception e)
                 {
-                    _logger.LogError("NuGetSearcherManager.CreateSearcher: Error loading auxiliary data.", e);
+                    _logger.LogError("NuGetSearcherManager.CreateSearcher - Exception thrown while loading auxiliary data: {Exception}", e);
```
shishirx34: if `verifiedPackages.json` is not found would this exception not reload any aux data? I am worried what would happen if we deployed search service before the job is deployed.

loic-sharma: If `verifiedPackages.json` is not found, the search service will fail to launch and all responses will return `UNINITIALIZED`. I believe this is the correct pattern to use - if there are issues, the search service should fail so that we are alerted quickly. We should deploy this change only when we believe the auxiliary data is ready to be consumed.

shishirx34: Please update the deployment task to make sure that the order of deployment is important during this deployment. https://github.com/NuGet/NuGetGallery/issues/4591

PR #228, opened 2017-08-25
Fix unit tests that failed on build machines

loic-sharma:


tests/NuGet.IndexingTests/ResponseFormatterTests.cs
```diff
@@ -324,7 +324,7 @@ public static IEnumerable<object[]> V2ResultData
...
+                    string.Format("{{\"totalHits\":2,\"index\":\"mockTakeDocuments\",\"indexTimestamp\":\"{0:G}\",\"data\":[{{\"PackageRegistration\":{{\"{2}\":\"{1}{2}0\",\"DownloadCount\":0,\"Verified\":false,\"Owners\":[]}},\"Version\":\"{1}{10}0\",\"NormalizedVersion\":\"{1}{3}0\",\"{4}\":\"{1}{4}0\",\"{5}\":\"{1}{5}0\",\"{6}\":\"{1}{6}0\",\"{7}\":\"{1}{7}0\",\"{8}\":\"{1}{8}0\",\"IsLatestStable\":false,\"IsLatest\":false,\"Listed\":true,\"DownloadCount\":0,\"Dependencies\":[],\"SupportedFrameworks\":[],\"PackageFileSize\":0,\"{9}\":\"{1}{9}0\",\"RequiresLicenseAcceptance\":false}},{{\"PackageRegistration\":{{\"{2}\":\"{1}{2}1\",\"DownloadCount\":0,\"Verified\":false,\"Owners\":[]}},\"Version\":\"{1}{10}1\",\"NormalizedVersion\":\"{1}{3}1\",\"{4}\":\"{1}{4}1\",\"{5}\":\"{1}{5}1\",\"{6}\":\"{1}{6}1\",\"{7}\":\"{1}{7}1\",\"{8}\":\"{1}{8}1\",\"IsLatestStable\":false,\"IsLatest\":false,\"Listed\":true,\"DownloadCount\":0,\"Dependencies\":[],\"SupportedFrameworks\":[],\"PackageFileSize\":0,\"{9}\":\"{1}{9}1\",\"RequiresLicenseAcceptance\":false}}]}}",
```
agr: Could it be turned into a C# class that is converted to the json string instead?

loic-sharma: Personally, I think the code that generates the expected results be as simple as possible. I prefer to avoid abstractions as long as the test is still "maintainable".

PR #231, opened 2017-09-08
V3 monitoring job - various fixes

scottbommarito:
A lot of various fixes in here done over the past couple months. I think it's time to get it merged in...

**Large changes:**

Modified the existing job to be queue-based.

This primarily involved splitting the existing "endpointmonitoring" job into two separate jobs and adding an additional job for retry logic.

- Catalog2Monitoring: reads the catalog and queues packages to be validated
- MonitoringProcessor: dequeues packages and validates them
- Monitoring2Monitoring: requeues packages that failed validation

I also added queue support to our storage APIs.

This major change provides two main benefits:

- easy retry flow
- we can run an unlimited number of `MonitoringProcessor`s to handle increased load (such as when we want to rebuild registration)

There are also 2 other jobs that I've included for ops purposes.

- Feed2Monitoring: reads the feed and queues packages that have not been scanned

This is both a consistency tool for the catalog and a consistency tool for the monitoring pipeline itself. I could run this on an interval but currently it is turned off.

- Catalog2MonitoringReprocessing: reads the entire catalog and queues the latest entries for each package

This can be used to rebuild the monitoring blobs if we desire (e.g. same use as lightning).

**Small fixes:**

1 - `CatalogIndexEntry`s are now correctly serialized in status blobs by using `CatalogTypesConverter`
2 - `FeedPackageIdentity`s have case insensitive comparisons (just like normal ids and versions)
3 - Added `SemVer2` support
4 - Removed the `packageContent` registration field validation
5 - Added ETag support to our `Storage` APIs to prevent status blobs from being overwritten and added that support to the `PackageMonitoringStatusService`.

...and potentially some others.


src/Catalog/CatalogTypeConverter.cs
```diff
@@ -0,0 +1,53 @@
...
+                } while (reader.TokenType != JsonToken.EndArray);
+            }
+            else
+            {
+                type = reader.Value;
+            }
+
+            if (type == null)
+            {
+                throw new ArgumentException("Failed to parse the type of a catalog entry!");
```
loic-sharma: >ArgumentException [](start = 26, length = 17)

I think a `InvalidDataException` would be more relevant here.

src/Catalog/CatalogTypeConverter.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            return objectType == typeof(IEnumerable<string>);
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            object type = null;
+            if (reader.TokenType == JsonToken.StartArray)
+            {
+                // If the type is stored in an array, use the first value and then discard the rest.
```
loic-sharma: >first [](start = 62, length = 5)

This comment seems wrong - doesn't this grab the *last* value?

src/Catalog/CatalogTypeConverter.cs
```diff
@@ -0,0 +1,53 @@
...
+                do
+                {
+                    reader.Read();
+                    if (reader.TokenType == JsonToken.String)
+                    {
+                        type = reader.Value;
```
loic-sharma: >Value [](start = 38, length = 5)

Is there a specific reason why `type` is an object and you cast `type` to string at the very end? If not, you could simplify the code a little by making `type` a string from the get-go and use `reader.ReadAsString` instead here.

src/Catalog/CatalogTypeConverter.cs
```diff
@@ -0,0 +1,53 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace NuGet.Services.Metadata.Catalog
+{
+    public class CatalogTypeConverter : JsonConverter
```
loic-sharma: >CatalogTypeConverter [](start = 17, length = 20)

Is this covered by tests? I can't seem to find any.

src/Catalog/Helpers/ParallelExecutionUtility.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.Metadata.Catalog.Helpers
+{
+    public static class ParallelExecutionUtility
+    {
+        public const int DefaultDegreeOfParallelism = 32;
+
+        public static Task ProcessInParallel(Func<Task> taskFactory, int degreeOfParallelism = DefaultDegreeOfParallelism)
```
loic-sharma: Add some comments as to what this does.

src/Catalog/Persistence/AzureStorageQueue.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        public override Task OnAdd(IStorageQueueMessage message, CancellationToken token)
+        {
+            var azureMessage = new AzureStorageQueueMessage(message.Contents);
+            return _queue.AddMessageAsync(azureMessage.Message, token);
+        }
+
+        public override async Task<IStorageQueueMessage> OnGetNext(CancellationToken token)
+        {
+            var nextMessage = await _queue.GetMessageAsync(_visibilityTimeout, null, null, token);
```
loic-sharma: >GetMessageAsync [](start = 43, length = 15)

named parameters.

src/Catalog/Persistence/AzureStorageQueue.cs
```diff
@@ -0,0 +1,44 @@
...
+            {
+                return null;
+            }
+
+            return new AzureStorageQueueMessage(nextMessage);
+        }
+
+        public override Task OnRemove(IStorageQueueMessage message, CancellationToken token)
+        {
+            return _queue.DeleteMessageAsync((message as AzureStorageQueueMessage).Message, token);
```
loic-sharma: >AzureStorageQueueMessage [](start = 57, length = 24)

This will throw a hard to debug null exception if the `serializer` returns a `IStorageQueueMessage` that isn't `AzureStorageQueueMessage`. Consider adding a check before the `as`, and if the check fails, throw an exception with a helpful message

src/Catalog/Persistence/FileStorage.cs
```diff
@@ -118,7 +128,7 @@ public string Path
             FileInfo fileInfo = new FileInfo(filename);
             if (fileInfo.Exists)
             {
-                return await Task.Run<StorageContent>(() => { return new StreamStorageContent(fileInfo.OpenRead()); }, cancellationToken);
+                return await Task.Run<StorageContent>(() => { return new StreamStorageContent(fileInfo.OpenRead()) { ETag = GetETag(fileInfo) }; }, cancellationToken);
```
loic-sharma: >StreamStorageContent [](start = 73, length = 20)

This is getting really dense. Please break this up into multiple lines.

src/Catalog/Registration/RecordingStorage.cs
```diff
@@ -30,18 +30,35 @@ public Task Save(Uri resourceUri, StorageContent content, CancellationToken canc
             return result;
         }
 
+        public Task SaveIfETag(Uri resourceUri, StorageContent content, string eTag, CancellationToken cancellationToken)
+        {
+            Task result = _innerStorage.SaveIfETag(resourceUri, content, eTag, cancellationToken);
```
loic-sharma: >Task [](start = 12, length = 4)

Use `var` here?

src/Catalog/CatalogTypeConverter.cs
scottbommarito: nice catch haha

src/Ng/Jobs/Feed2MonitoringJob.cs
```diff
@@ -0,0 +1,191 @@
...
+            await ParallelExecutionUtility.ProcessInParallel(() => ProcessPackages(validPackagesSet, invalidPackages, cancellationToken), _degreeOfParallelism);
+
+            if (!invalidPackages.IsEmpty)
+            {
+                var invalidPackagesList = new List<FeedPackageIdentity>();
+                Logger.LogError("The following packages are missing:");
+                while (invalidPackages.TryDequeue(out var invalidPackage))
+                {
+                    invalidPackagesList.Add(invalidPackage);
+                    Logger.LogError($"{invalidPackage.Id} {invalidPackage.Version}");
```
loic-sharma: >LogError [](start = 27, length = 8)

I would recommend concatenating all the packages first and then emitting a single error. These logs may be difficult to understand if there are multiple threads running all emitting their own logs in between.

src/Ng/Jobs/Feed2MonitoringJob.cs
```diff
@@ -0,0 +1,191 @@
...
+            }
+        }
+
+        private async Task<HashSet<FeedPackageIdentity>> GetValidPackages(CancellationToken cancellationToken)
+        {
+            var validPackagesList = await _statusService.ListAsync(cancellationToken);
+            return new HashSet<FeedPackageIdentity>(validPackagesList.Select(p => p.Package).ToList());
+        }
+
+        private static async Task<T> RepeatUntilSuccessOrMaxFailures<T>(Func<Task<T>> action) where T : class
```
loic-sharma: >RepeatUntilSuccessOrMaxFailures [](start = 37, length = 31)

Should this maybe record all exceptions that are caught, and then if `_maxRequestFailures` is reached, log them?

src/Catalog/CatalogTypeConverter.cs
scottbommarito: I think this is better...hard to explain with just text but using ReadAsString would require me to move around the rest of the flow and I think this is simpler.

agr:
Can we have some documentation for the new code?..

scottbommarito:
Closing to split this up more digestably--stay tuned.

src/Ng/Jobs/Feed2MonitoringJob.cs
scottbommarito: This part should be synchronous, after all threads have finished. Additionally, concatenating all packages could be a very very bad idea if there are a lot of packages.

src/NuGet.Services.Metadata.Catalog.Monitoring/MonitoringStatusVerificationCollector.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
```
scottbommarito: Delete this, no longer needed.

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusService.cs
```diff
@@ -28,40 +28,69 @@ public class PackageMonitoringStatusService : IPackageMonitoringStatusService
...
                 (await Task.WhenAll(statusTasks))
-                .SingleOrDefault(s => s != null);
+                .Where(s => s != null);
+            
+            if (statuses.Count() == 0)
+            {
+                return null;
+            }
+
+            // If more than one status exist for a single package, find the status with the latest timestamp.
```
scottbommarito: `If multiple statuses exist`

PR #233, opened 2017-09-09
FeedPackageIdentity comparisons should be case insensitive

scottbommarito:
`abcd 1.0.0-abc` should be equal to `ABCD 1.0.0-ABC`

(this is a fix for the V3 monitoring job)


src/Catalog/Helpers/FeedPackageIdentity.cs
```diff
@@ -30,12 +30,12 @@ public FeedPackageIdentity(PackageIdentity package)
 
         public bool Equals(FeedPackageIdentity other)
         {
-            return Id == other.Id && Version == other.Version;
+            return Id.ToLowerInvariant() == other.Id.ToLowerInvariant() && Version.ToLowerInvariant() == other.Version.ToLowerInvariant();
```
skofman1: please add UTs that make sure Equals and GetHashCode are equivalent. I.e. if Equals returns true for two FeePackageIdentity classes, their hash codes are also identical (and vice versa).

skofman1: please add UTs that make sure Equals and GetHashCode are equivalent. I.e. if Equals returns true for two FeePackageIdentity classes, their hash codes are also identical (and vice versa).

scottbommarito:
@skofman1 added tests as requested

scottbommarito:
@skofman1 added more tests as requested offline

PR #239, opened 2017-09-23
V3 monitoring job - remove packageContent validator and do not alert on leaf nodes that are left after delete

scottbommarito:
The `packageContent` validator no longer made any sense after `packageContent` was modified to link to the flat-container.

Additionally, for the `exists` validator, the job no longer considers leaf nodes that are left after a delete to be a failure.


src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/Registration/RegistrationExistsValidator.cs
```diff
@@ -30,6 +30,15 @@ public override Task<bool> ShouldRunLeaf(ValidationContext data, PackageRegistra
 
             if (v2Exists != v3Exists)
             {
+                // Currently, leaf nodes are not deleted after a package is deleted.
+                // This is a known bug. Do not fail validations because of it.
+                // See https://github.com/NuGet/NuGetGallery/issues/4475
+                if (!v2Exists && v3Exists && !(v2 is PackageRegistrationIndexMetadata) && !(v3 is PackageRegistrationIndexMetadata))
```
chenriksson: You already know v2Exists != v3Exists, so do you only need to check one of these here?

tests/NgTests/Validators/RegistrationExistsValidatorTestData.cs
```diff
@@ -21,18 +21,52 @@ public class RegistrationExistsValidatorTestData : RegistrationLeafValidatorTest
 
         public override IEnumerable<Func<PackageRegistrationIndexMetadata>> CreateIndexes => new Func<PackageRegistrationIndexMetadata>[]
         {
-            () => new PackageRegistrationIndexMetadata(),
             () => null
         };
 
         public override IEnumerable<Func<PackageRegistrationIndexMetadata>> CreateSkippedIndexes => new Func<PackageRegistrationIndexMetadata>[0];
 
+        public override IEnumerable<Func<Tuple<PackageRegistrationIndexMetadata, PackageRegistrationIndexMetadata, bool>>> CreateSpecialIndexes => new Func<Tuple<PackageRegistrationIndexMetadata, PackageRegistrationIndexMetadata, bool>>[]
```
chenriksson: nit: wrap long line

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/Registration/RegistrationExistsValidator.cs
joelverhagen: Why are we checking types here? Could you explain the meaning here?

scottbommarito: yes, very true

scottbommarito: When we are running validations against an index here, the type of `v2` and `v3` will be `PackageRegistrationIndexMetadata`. If we are running validations against a leaf, it will be a `PackageRegistrationLeafMetadata`, which `PackageRegistrationIndexMetadata` inherits from. Therefore, we are checking that `v3` is not a `PackageRegistrationIndexMetadata` because we only want to avoid failing validations if it is a `PackageRegistrationLeafMetadata`.

shishirx34: From what I gather, there would be only two types of metadata, 'index' and 'leaf, wouldn't this be simplified by:
```
if (v3Exists && !(v3 is PackageRegistrationIndexMetadata))
```
The third check `(!(v2 is PackageRegistrationIndexMetadata))` might altogether be unnecessary, because if it doesn't exist, wouldn't it be null?

scottbommarito: Correct. I will fix.

PR #242, opened 2017-09-26
V3 monitoring jobs - if an exception cannot be deserialized, return a default exception instead

scottbommarito:
This is a bug fix for the job that requeues failed validations. Basically, the job deserializes JSON that represents the status of a package and then requeues those packages using the data it has.

JSON.NET, for whatever reason, cannot deserialize some exceptions thrown by `HttpClient` and throws an exception itself when it tries to do so. To prevent this from blocking the requeueing job, this PR introduces a `JsonConverter` that attempts to deserialize the exception, and if it fails, replaces the exception with a default one that states that the exception could not be deserialized.


joelverhagen:
Why are we deserializing exceptions? A serialized exception should either be mapped to a more useful serialized type or should be treated as a `JObject` after parsing. What use is a deserialized exception?

scottbommarito:
@joelverhagen 

The exceptions are serialized as-is because we want the maximum traceability of them in the status blobs. I don't want to do any processing to the exceptions because we could lose vital debugging information.

These exceptions are then deserialized as a part of the validation retry logic, which must read the status blobs to discern what to do.

If we wanted to, these deserialized exceptions could be used to implement smarter validation retry logic based on what exception caused the validation to fail.

That, however, is not the case now, so this fix just is a temporary workaround to protect deserializing status blobs from failure due to exceptions failing to deserialize.

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/ValidationResultExceptionConverter.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            try
+            {
+                return serializer.Deserialize(reader, objectType);
+            }
+            catch
+            {
+                return new Exception("This exception could not be deserialized. Please try running validations again.");
```
skofman1: Never throw a general exception. You should throw Deserialization exception, and add the original exception caught as an inner exception so it can be logged.

skofman1:
@joelverhagen is correct. Exceptions shouldn't be part of a status blob. A status blob should contain a state enum indicating failure reason. Exceptions should be logged to AI.

scottbommarito:
@skofman1 
I disagree. Logging the exceptions to AI makes it much more difficult to debug a live site issue, requiring the user to switch from viewing the storage blobs to checking AI and writing a query to find the exception that caused the failure. Including the exception, which is essentially the reason for the failure, in the status blob allows much easier use of the monitoring pipeline.

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/ValidationResultExceptionConverter.cs
cristinamanum: Maybe add few comments ... In what scenarios the ExceptionConverter will be used?  How the returned exception will be used ?  Why not let it throw?

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/ValidationResultExceptionConverter.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
```
chenriksson: header

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/ValidationResultExceptionConverter.cs
```diff
@@ -0,0 +1,30 @@
...
+using Newtonsoft.Json;
+
+namespace NuGet.Services.Metadata.Catalog.Monitoring
+{
+    public class ValidationResultExceptionConverter : JsonConverter
+    {
+        public override bool CanConvert(Type objectType)
+        {
+            return typeof(Exception).IsAssignableFrom(objectType);
```
chenriksson: Do we have to use the converter for all exceptions, when only certain HttpClient exceptions are failing to deserialize?

scottbommarito: there's a default one (by JSON.NET) that this calls

```
return serializer.Deserialize(reader, objectType);
```

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/SafeExceptionConverter.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            try
+            {
+                return serializer.Deserialize(reader, objectType);
+            }
+            catch
+            {
+                // When deserializing the exception fails, we don't want to fail deserialization of the entire object.
+                // Return an ExceptionDeserializationException instead.
+                return new ExceptionDeserializationException();
```
loic-sharma: >ExceptionDeserializationException [](start = 27, length = 33)

Why would deserializing an exception fail? Is this a common case? If so, explain how. If not, should we maybe log something so that we can debug these issues?

scottbommarito: Basically, some exceptions don't implement the correct structure to be successfully deserialized by JSON.NET.

I encountered it in the past, and I found a StackOverflow post explaining the exact issue, but I can't seem to find it now.

scottbommarito: It's not something we can really debug, so I didn't add any logging here.

loic-sharma: Fair enough - please add a comment explaining this! :)

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/SafeExceptionConverter.cs
```diff
@@ -0,0 +1,39 @@
...
+            return typeof(Exception).IsAssignableFrom(objectType);
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            try
+            {
+                return serializer.Deserialize(reader, objectType);
+            }
+            catch
```
chenriksson: Seems like this class is really an attempt at adding a 'TryDeserialize' to Json.NET for exception types?

Is it so bad to handle this from the caller?

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/ExceptionDeserializationException.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace NuGet.Services.Metadata.Catalog.Monitoring
+{
+    [Serializable]
+    public class ExceptionDeserializationException : Exception
+    {
+        public ExceptionDeserializationException()
+            : base("This exception could not be deserialized.")
+        {
+        }
```
chenriksson: Could this be a more general nuget serialization exception, similar to what JsonException is to JSON.Net? This allows for reuse by any of our custom converters. Maybe NuGetJsonException?

Also, I would support passing a JsonException as an inner exception.

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/SafeExceptionConverter.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// If the exception fails to deserialize, returns an <see cref="ExceptionDeserializationException"/> instead of failing.
+    /// </summary>
+    public class SafeExceptionConverter : JsonConverter
+    {
+        public override bool CanConvert(Type objectType)
+        {
+            return typeof(Exception).IsAssignableFrom(objectType);
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
```
chenriksson: Arg validation? Throw if objectType isn't an exception?

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/SafeExceptionConverter.cs
scottbommarito: You can't handle it from the caller. This deserialization is done as a part of deserializing a larger object, which fails when this field fails to deserialize.

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/SafeExceptionConverter.cs
scottbommarito: JSON.NET should never pass us an object that is not an `Exception` because of the `CanConvert` function above.

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/ExceptionDeserializationException.cs
scottbommarito: I don't think sharing exceptions between converters provides much value. If anything we'd want them to be separate types so we can catch individual issues.

I've actually decided to get rid of this exception type and just return the `Exception` that was originally thrown.

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/SafeExceptionConverter.cs
cristinamanum: I agree with @skofman1 comments above regarding the exceptions. In my opinion the exception's data store needs to be easy to be queried and blob storage is not. We also would probably like to have an common exception data storage and we currently use AI.

scottbommarito: @cristinamanum Regardless of that, the infrastructure already uses exceptions in blob storage. This is merely a bug fix to having exceptions in blob storage. I think changes to the infrastructure are outside of the scope of this PR.

joelverhagen: @scottbommarito, understood. So we're saying "out of scope for this PR". I'm fine with that.

chenriksson: Still think we need to improve validation of our public APIs, but seems to be a common theme across jobs and servercommon.

PR #243, opened 2017-09-26
V3 monitoring jobs - updates to the framework to support the new queue based infrastructure

scottbommarito:
This PR consists of a series of smaller changes to support the new queue based infrastructure jobs

- Remove all trace of `EndpointMonitoringJob`, the old, now deprecated job
    - A future PR will add back the new jobs
- Remove `IList<JObject>` from `IMonitoringNotificationService`'s `OnPackageValidationFailedAsync`
    - This field is no longer used is any implementation
    - Due to refactoring, we may not have the catalog entries of an object when validations fail to run
- Improve `PackageMonitoringStatusService`
    - Added `ListAsync` to the interface as a lightweight solution for getting all packages that have been validated, because `GetAsync(PackageStatus)` is impossibly slow when restoring from `PackageStatus.Valid`
    - Added logic to handle when saving a package's status fails and leaves status blobs in multiple package states
- Add `JsonSerializerUtility` so that json serialization settings are stored in a shared place
- `PackageValidator` now operates on `PackageValidatorContext` objects
- `ValidationCollector` no longer performs any validation logic and instead add packages to a queue of `PackageValidatorContext`s
- Refactored the factories for `ValidationCollector` and `ValidationFactory` so that they can be more easily used independently of each other


src/Catalog/Helpers/ParallelExecutionUtility.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Threading.Tasks;
+
+namespace NuGet.Services.Metadata.Catalog.Helpers
+{
+    public static class ParallelExecutionUtility
+    {
+        public const int DefaultDegreeOfParallelism = 32;
+
+        /// <summary>
+        /// Runs a task created by <paramref name="taskFactory"/> on a number of threads specified by <paramref name="degreeOfParallelism"/> and waits for all tasks to complete.
```
skofman1: >umber of threads spe [](start = 72, length = 20)

This code creates "degreeOfParallelism" number of tasks, and waits for them to complete. We have no control over how many threads are actually used for the execution, because we don't control the thread pool.

src/Catalog/Helpers/ParallelExecutionUtility.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// </summary>
+        /// <param name="taskFactory">Creates tasks for each thread to run.</param>
+        /// <param name="degreeOfParallelism">The number of threads to run tasks on.</param>
+        /// <returns>A task that completes when all threads have completed their tasks.</returns>
+        public static Task ProcessInParallel(Func<Task> taskFactory, int degreeOfParallelism = DefaultDegreeOfParallelism)
```
skofman1: >ProcessInParallel [](start = 27, length = 17)

couldn't find where this code is used..

src/Ng/Ng.csproj
```diff
@@ -353,32 +352,13 @@
...
-    <None Include="Scripts\Catalog2RegistrationV3-Reg1.cmd" />
-    <None Include="Scripts\DB2Catalog.cmd" />
-    <None Include="Scripts\EndpointMonitoring.cmd" />
-    <None Include="Scripts\Feed2CatalogV3-Secondary.cmd" />
-    <None Include="Scripts\Feed2CatalogV3.cmd" />
-    <None Include="Scripts\Functions.ps1" />
-    <None Include="Scripts\PreDeploy.ps1" />
-    <None Include="Scripts\PostDeploy.ps1" />
-    <None Include="Scripts\nssm.exe" />
+    <None Include="Scripts\**" />
```
skofman1: nice!

skofman1:
> using System.Collections.Generic;

file header

---
Refers to: src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/PackageValidatorContext.cs:1 in 341cc6a. [](commit_id = 341cc6a1b9b32a5253ddc4beb3049bc666ea8900, deletion_comment = False)

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/ValidatorFactoryFactory.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System;
+using System.Collections.Generic;
```
skofman1: file header

src/Catalog/Helpers/ParallelExecutionUtility.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            var bagTakers = new List<Task>();
+            for (int i = 0; i < degreeOfParallelism; i++)
```
cristinamanum: Is there any reason for not using Parallel.For or Paralle.Foreach instead of ProcessInParallel method?

src/Catalog/Helpers/ParallelExecutionUtility.cs
cristinamanum: Need a way to cancel the operation.

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/IPackageMonitoringStatusService.cs
```diff
@@ -14,6 +14,11 @@ namespace NuGet.Services.Metadata.Catalog.Monitoring
     public interface IPackageMonitoringStatusService
     {
         /// <summary>
+        /// Returns a list of every package that has been monitored and its <see cref="PackageState"/>.
+        /// </summary>
+        Task<IEnumerable<PackageMonitoringStatusListItem>> ListAsync(CancellationToken token);
+
+        /// <summary>
```
cristinamanum: For the UpdateAsync method below - maybe it should return Task<bool> to capture the result of the operation?

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusListItem.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using NuGet.Services.Metadata.Catalog.Helpers;
+
+namespace NuGet.Services.Metadata.Catalog.Monitoring
+{
+    public class PackageMonitoringStatusListItem
+    {
```
cristinamanum: Why PackageMonitoringStatusListItem and not only PackageMonitoringStatus

src/Catalog/Helpers/ParallelExecutionUtility.cs
scottbommarito: Wasn't aware of it--I'll get rid of this class and use it

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusListItem.cs
scottbommarito: The actual `PackageMonitoringStatus` object is much more complex. This object is designed to be returned by the `IPackageMonitoringStatusService.ListAsync`, which does not have the full information from the status blobs.

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusService.cs
```diff
@@ -28,40 +26,69 @@ public class PackageMonitoringStatusService : IPackageMonitoringStatusService
...
-
         public PackageMonitoringStatusService(IStorageFactory storageFactory, ILogger<PackageMonitoringStatusService> logger)
         {
             _logger = logger;
             _storageFactory = storageFactory;
         }
 
-        public async Task<PackageMonitoringStatus> GetAsync(FeedPackageIdentity package, CancellationToken token)
+        public async Task<IEnumerable<PackageMonitoringStatusListItem>> ListAsync(CancellationToken token)
         {
```
cristinamanum: The token seems that  it is not used .

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusService.cs
```diff
@@ -28,40 +26,69 @@ public class PackageMonitoringStatusService : IPackageMonitoringStatusService
...
-                        return await GetPackageAsync(GetStorage(state), package, token);
-                    })
-                );
+                        try
+                        {
+                            return new PackageMonitoringStatusListItem(ParsePackageUri(item.Uri), (PackageState)Enum.Parse(typeof(PackageState), state));
+                        }
+                        catch
+                        {
+                            return null;
```
cristinamanum: Do these exception matter.

src/Catalog/Helpers/ParallelExecutionUtility.cs
scottbommarito: Will replace with `Parallel.ForEach`

src/Catalog/Helpers/ParallelExecutionUtility.cs
scottbommarito: It's used in https://github.com/NuGet/NuGet.Services.Metadata/pull/244 and https://github.com/NuGet/NuGet.Services.Metadata/pull/245

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusService.cs
```diff
@@ -70,60 +97,68 @@ public PackageMonitoringStatusService(IStorageFactory storageFactory, ILogger<Pa
...
         {
-            // Guarantee that we never have the same package in multiple states by deleting it first.
-            await DeleteAsync(status.Package, token);
+            foreach (int stateInt in Enum.GetValues(typeof(PackageState)))
+            {
+                var state = (PackageState)stateInt;
+                if (state == status.State)
+                {
+                    // Save the new status.
+                    await SaveAsync(status, token);
```
cristinamanum: SaveAsync and DeleteAsync have different signature. Any reason for this?

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusService.cs
```diff
@@ -70,60 +97,68 @@ public PackageMonitoringStatusService(IStorageFactory storageFactory, ILogger<Pa
...
+                }
+                else
+                {
+                    // Delete the existing status.
+                    await DeleteAsync(status.Package, state, token);
```
cristinamanum: What would be the state id the DeleteAsync fails ? Will be future update operation impacted?

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/IPackageMonitoringStatusService.cs
scottbommarito: If it fails, the task will throw an exception. Therefore, if the method is called and returns successfully, it can be assumed to had succeeded.

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusService.cs
scottbommarito: it's used by `var list = await storage.List(token);`

i will add it to the `Task.Run` call as well.

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusService.cs
scottbommarito: I don't think they do, but I'll add a log just in case

src/Catalog/Helpers/ParallelExecutionUtility.cs
scottbommarito: Deleting class so no longer necessary

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusService.cs
scottbommarito: To clarify, this service stores status in two separate locations based on whether or not validations succeeded or failed. When a package is being moved from one status to the other, it needs to delete the package from the other location.

`SaveAsync` requires the exact `PackageMonitoringStatus` that is to be saved.

`DeleteAsync` just needs the package and version that it's deleting and the status folder that it's deleting from.

I could rewrite these functions to take in the same parameters, but I don't see a benefit.

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusService.cs
scottbommarito: This method would throw, the queue message for this package to be validated would not be removed from the queue, and the validation would run again in approximately 5 minutes.

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusListItem.cs
```diff
@@ -0,0 +1,20 @@
...
+        public FeedPackageIdentity Package { get; private set; }
+
+        public PackageState State { get; private set; }
```
joelverhagen: `{ get; }`

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusService.cs
```diff
@@ -28,40 +26,70 @@ public class PackageMonitoringStatusService : IPackageMonitoringStatusService
...
 
-        public async Task<PackageMonitoringStatus> GetAsync(FeedPackageIdentity package, CancellationToken token)
+        public async Task<IEnumerable<PackageMonitoringStatusListItem>> ListAsync(CancellationToken token)
         {
-            var statusTasks =
+            var listTasks =
                 Enum.GetNames(typeof(PackageState))
-                .Select(state =>
-                    Task.Run(async () =>
+                .Select(state => Task.Run(async () =>
```
joelverhagen: Why `Task.Run`? Presumably you want to immediately run on the thread pool? If so, `Task.Yield()` is for this.

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusService.cs
```diff
@@ -152,11 +203,21 @@ private Task<PackageMonitoringStatus> GetPackageAsync(Storage storage, string fi
...
+        private async Task<PackageMonitoringStatus> GetPackageAsync(Persistence.Storage storage, Uri packageUri, CancellationToken token)
         {
             try
             {
-                return JsonConvert.DeserializeObject<PackageMonitoringStatus>(await GetStorageContentsAsync(storage, packageUri, token), SerializerSettings);
+                var content = await storage.Load(packageUri, token);
+                string statusString = null;
+                using (Stream stream = content.GetContentStream())
+                {
+                    StreamReader reader = new StreamReader(stream);
```
joelverhagen: dispose `StreamReader`

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/JsonSerializerUtility.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace NuGet.Services.Metadata.Catalog.Monitoring
+{
+    public static class JsonSerializerUtility
+    {
+        /// <summary>
+        /// The <see cref="JsonSerializerSettings"/> to use.
+        /// </summary>
+        public static JsonSerializerSettings SerializerSettings => _serializerSettings.Value;
+        private static Lazy<JsonSerializerSettings> _serializerSettings = new Lazy<JsonSerializerSettings>(() =>
```
joelverhagen: Why lazy?

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/PackageValidator.cs
```diff
@@ -20,22 +20,23 @@ namespace NuGet.Services.Metadata.Catalog.Monitoring
...
         private readonly StorageFactory _auditingStorageFactory;
 
         public PackageValidator(
-            IEnumerable<IAggregateValidator> endpointValidators,
+            IEnumerable<IAggregateValidator> aggregateValidators,
             StorageFactory auditingStorageFactory,
             ILogger<PackageValidator> logger)
         {
-            if (endpointValidators.Count() < 1)
+            if (aggregateValidators.Count() < 1)
```
joelverhagen: nit: `!Any()`

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/PackageValidatorContext.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Collections.Generic;
+using Newtonsoft.Json;
+using NuGet.Services.Metadata.Catalog.Helpers;
+
+namespace NuGet.Services.Metadata.Catalog.Monitoring
+{
+    public class PackageValidatorContext
```
joelverhagen: This seems like a lot of state to be going into a queue message. What if queue messages are lost? How do we recover?

joelverhagen:
I am unfamiliar with the infrastructure of this system now. It seems to have evolved a lot since the initial design meeting. Do you have an up-to-date design document?

scottbommarito:
@joelverhagen Will respond over email (it's in the OneNote).

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/PackageValidatorContext.cs
scottbommarito: Once a message is in the queue, I think the chances of the message disappearing are probably roughly similar to that of a storage blob disappearing, which is not a situation we plan for right now. Please correct me if I'm wrong however.

Regarding failures of the message being pushed into the queue, the job pushing that message will retry again.

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/JsonSerializerUtility.cs
scottbommarito: No huge reason, just thought it was prettier syntactically than the other options.

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusService.cs
scottbommarito: TIL how to make `Task`s correctly

loic-sharma:
>     /// Runs a <see cref="PackageValidator"/> against collected packages.

Could you update this comment? `ValidationCollector` no longer uses a `PackageValidator` directly.

---
Refers to: src/NuGet.Services.Metadata.Catalog.Monitoring/ValidationCollector.cs:18 in eabae63. [](commit_id = eabae63c8ccac3bc9c4bf873493960bb05d24cdd, deletion_comment = False)

src/NuGet.Services.Metadata.Catalog.Monitoring/ValidationCollectorFactory.cs
```diff
@@ -42,63 +40,28 @@ public ValidationCollectorFactory(ILoggerFactory loggerFactory)
         }
 
         public Result Create(
```
loic-sharma: >Create [](start = 22, length = 6)

Could we add documentation for this method?

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusListItem.cs
loic-sharma: Add a comment about this.

---
In reply to: [141684494](https://github.com/NuGet/NuGet.Services.Metadata/pull/243#discussion_r141684494) [](ancestors = 141684494)

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusService.cs
```diff
@@ -28,40 +26,74 @@ public class PackageMonitoringStatusService : IPackageMonitoringStatusService
...
+            _logger = logger;
+            _storageFactory = storageFactory;
+        }
 
-            settings.Converters.Add(new NuGetVersionConverter());
-            settings.Converters.Add(new StringEnumConverter());
+        public async Task<IEnumerable<PackageMonitoringStatusListItem>> ListAsync(CancellationToken token)
+        {
+            var listTasks =
+                Enum.GetNames(typeof(PackageState))
```
loic-sharma: `GetNames` uses reflection and is an expensive call. Cache this in the constructor.

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusService.cs
```diff
@@ -28,40 +26,74 @@ public class PackageMonitoringStatusService : IPackageMonitoringStatusService
...
+                .Aggregate((current, next) => current.Concat(next))
+                .Where(i => i != null);
+        }
 
-        public PackageMonitoringStatusService(IStorageFactory storageFactory, ILogger<PackageMonitoringStatusService> logger)
+        private async Task<IEnumerable<PackageMonitoringStatusListItem>> GetListItems(string state, CancellationToken token)
         {
-            _logger = logger;
-            _storageFactory = storageFactory;
+            await Task.Yield();
```
loic-sharma: >Yield [](start = 23, length = 5)

There should be no need for the `Task.Yield` since you await on `storage.List` anyways, right?

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusService.cs
```diff
@@ -137,12 +177,27 @@ private string GetPackageFileName(FeedPackageIdentity package)
                 $"{idString}.{versionString}.json";
         }
 
-        private Task<PackageMonitoringStatus> GetPackageAsync(Storage storage, FeedPackageIdentity package, CancellationToken token)
+        private FeedPackageIdentity ParsePackageUri(Uri packageUri)
+        {
+            var uriSegments = packageUri.Segments;
```
loic-sharma: >Segments [](start = 41, length = 8)

Could you provide an example URI in a comment here?

scottbommarito: good idea

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusService.cs
scottbommarito: yup, nice catch

src/NuGet.Services.Metadata.Catalog.Monitoring/ValidationCollectorFactory.cs
scottbommarito: added

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusService.cs
scottbommarito: done, and also cached `Enum.GetValues`

src/NuGet.Services.Metadata.Catalog.Monitoring/ValidationCollectorFactory.cs
loic-sharma: This needs more documentation. What's the `messageHandlerFactory`? What gets written to the `monitoringStorageFactory`? etc..

---
In reply to: [142288714](https://github.com/NuGet/NuGet.Services.Metadata/pull/243#discussion_r142288714) [](ancestors = 142288714)

src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/JsonSerializerUtility.cs
loic-sharma: I agree it's prettier syntactically, but, lazy has thread-safety overhead. I would use a static constructor instead.

src/Catalog/Helpers/ParallelExecutionUtility.cs
scottbommarito: Decided to keep it because `Parallel.ForEach` is not async.

src/Catalog/Helpers/ParallelExecutionUtility.cs
scottbommarito: The user can cancel the tasks they provide to this, which should provide the same functionality, but I will add a cancellation token

src/Catalog/Helpers/ParallelExecutionUtility.cs
scottbommarito: Decided to keep it because `Parallel.ForEach` is not async.

src/Catalog/Helpers/ParallelExecutionUtility.cs
scottbommarito: `Task.WhenAll` doesn't support passing a `CancellationToken`, so all `CancellationToken`s must be set up by the user in the `Func<Task> taskFactory`.

src/Catalog/Helpers/ParallelExecutionUtility.cs
chenriksson: See Stephen Toub's [ForEachAsync](https://blogs.msdn.microsoft.com/pfxteam/2012/03/04/implementing-a-simple-foreachasync) blog posts. You want something similar, but based on Parallel.For... maybe below is cleaner?

```
public static Task ForAsync(int dop, Func<Task> body, CancellationToken ct)
{
    return Task.WhenAll(
        from i in Enumerable.Repeat(body, dop)
        select Task.Run(i, ct)
        );
}
```

chenriksson: Maybe just name the class Parallel? So invocation looks like:
```
  Parallel.ForAsync(32, ProcessPackages);
```

src/Catalog/Helpers/ParallelExecutionUtility.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    public static class ParallelExecutionUtility
+    {
+        public const int DefaultDegreeOfParallelism = 32;
+
+        /// <summary>
+        /// Runs a task created by <paramref name="taskFactory"/> on a number of threads specified by <paramref name="degreeOfParallelism"/> and waits for all tasks to complete.
+        /// </summary>
+        /// <param name="taskFactory">Creates tasks for each thread to run.</param>
+        /// <param name="degreeOfParallelism">The number of threads to run tasks on.</param>
```
loic-sharma: >threads [](start = 60, length = 7)

This isn't strictly correct as creating a `Task` does not necessarily create a new thread. I wouldn't mention threads here.

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusService.cs
```diff
@@ -9,59 +9,96 @@
...
+using CatalogStorage = NuGet.Services.Metadata.Catalog.Persistence.Storage;
+
 namespace NuGet.Services.Metadata.Catalog.Monitoring
 {
     /// <summary>
     /// Manages the storage and access of the status of packages that validation has run against.
     /// </summary>
     public class PackageMonitoringStatusService : IPackageMonitoringStatusService
     {
+        private string[] _packageStateNames;
```
loic-sharma: >_packageStateNames [](start = 25, length = 18)

I would make `_packageStateNames` and `_packageStateValues` static since they only need to be calculated once per process, instead of once per instance of this class.

src/Catalog/Helpers/ParallelExecutionUtility.cs
scottbommarito: I really like using `Enumerable.Repeat`, although I've reduced it down to
```
return Task.WhenAll(
    Enumerable
        .Repeat(taskFactory, degreeOfParallelism)
        .Select(f => f()));
```

scottbommarito: can't make it an extension of `Parallel`, you can't make extension methods on static classes

scottbommarito: I like `ParallelAsync` however

PR #244, opened 2017-09-26
V3 monitoring jobs - add new queue-based jobs

scottbommarito:
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/243

Added `Catalog2Monitoring`, `MonitoringProcessor`, and `Monitoring2Monitoring`, the three jobs that make up the V3 monitoring jobs.

`Catalog2Monitoring`:
Runs on the catalog and pushes new commits to a queue.

`MonitoringProcessor`:
Listens to the queue and runs validations on each commit.

`Monitoring2Monitoring`:
Requeues packages that have failed validation.

PR #245, opened 2017-09-26
V3 monitoring jobs - add new queue based ops tools

scottbommarito:
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/244

Added `Feed2Monitoring` and `Catalog2MonitoringReprocessing`, two ops tools that can alternatively be set up to run regularly.

`Feed2Monitoring`:
Reads all packages from the feed and verifies that they have been checked by the V3 monitoring jobs. Can be run if there are concerns that the V3 monitoring jobs have skipped packages.

`Catalog2MonitoringReprocessing`:
Reads the entire catalog (using lightning) and queues the latest commit for each package. Should be run to regenerate the monitoring status blobs (instead of running `Catalog2Monitoring`).

### DIFF
--- a/src/Ng/Jobs/Catalog2MonitoringReprocessingJob.cs
+++ b/src/Ng/Jobs/Catalog2MonitoringReprocessingJob.cs
@@ -54,7 +54,6 @@ namespace Ng.Jobs
             var entries = await _catalogIndexReader.GetEntries();
 
             var packageEntries = entries
-                .Where(c => !c.IsDelete())
                 .GroupBy(c => new PackageIdentity(c.Id, c.Version))
                 .Select(g => g.OrderByDescending(c => c.CommitTimeStamp).First());
 
--- a/src/Ng/Jobs/Catalog2MonitoringReprocessingJob.cs
+++ b/src/Ng/Jobs/Catalog2MonitoringReprocessingJob.cs
@@ -18,7 +18,6 @@ namespace Ng.Jobs
         private IPackageMonitoringStatusService _statusService;
         private IStorageQueue<PackageValidatorContext> _queue;
         private CatalogIndexReader _catalogIndexReader;
-        private ReadCursor _front;
 
         public Catalog2MonitoringReprocessingJob(ILoggerFactory loggerFactory)
             : base(loggerFactory)
@@ -43,9 +42,6 @@ namespace Ng.Jobs
 
             _queue = CommandHelpers.CreateStorageQueue<PackageValidatorContext>(arguments);
 
-            var storage = monitoringStorageFactory.Create();
-            _front = ValidationCollectorFactory.GetFront(monitoringStorageFactory);
-
             _catalogIndexReader = new CatalogIndexReader(new Uri(source), new CollectorHttpClient(messageHandlerFactory()));
         }
 
@@ -57,11 +53,7 @@ namespace Ng.Jobs
             Logger.LogInformation("Parsing catalog for all entries.");
             var entries = await _catalogIndexReader.GetEntries();
 
-            await _front.Load(cancellationToken);
-            var currentTimestamp = _front.Value;
-
             var packageEntries = entries
-                .Where(c => c.CommitTimeStamp < currentTimestamp)
                 .GroupBy(c => new PackageIdentity(c.Id, c.Version))
                 .Select(g => g.OrderByDescending(c => c.CommitTimeStamp).First());
 
--- a/src/Ng/Jobs/Catalog2MonitoringReprocessingJob.cs
+++ b/src/Ng/Jobs/Catalog2MonitoringReprocessingJob.cs
@@ -1,0 +1,90 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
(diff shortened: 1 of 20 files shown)


src/Ng/Jobs/Catalog2MonitoringReprocessingJob.cs
```diff
@@ -0,0 +1,82 @@
...
+using NuGet.Packaging.Core;
+using NuGet.Services.Configuration;
+using NuGet.Services.Metadata.Catalog;
+using NuGet.Services.Metadata.Catalog.Helpers;
+using NuGet.Services.Metadata.Catalog.Monitoring;
+using NuGet.Services.Storage;
+
+namespace Ng.Jobs
+{
+    public class Catalog2MonitoringReprocessingJob : LoopingNgJob
```
loic-sharma: >Catalog2MonitoringReprocessingJob [](start = 17, length = 33)

Please add a description to what these jobs do here.

src/Ng/Jobs/Catalog2MonitoringReprocessingJob.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            Logger.LogInformation("Processing packages.");
+            foreach (var packageEntry in packageEntries)
+            {
+                var feedPackage = new FeedPackageIdentity(new PackageIdentity(packageEntry.Id, packageEntry.Version));
+                Logger.LogInformation("Processing {Id} {Version}", feedPackage.Id, feedPackage.Version);
+
+                if (!validatedPackages.Contains(feedPackage))
+                {
+                    Logger.LogError("{Id} {Version} was skipped!", feedPackage.Id, feedPackage.Version);
```
loic-sharma: >LogError [](start = 27, length = 8)

This isn't an error, right? Considering changing this to `LogWarning`. Also, maybe add something like "Reprocessing package" to the log message.

src/Ng/Jobs/Catalog2MonitoringReprocessingJob.cs
```diff
@@ -0,0 +1,82 @@
...
+                        new PackageValidatorContext(feedPackage, new CatalogIndexEntry[] { packageEntry }),
+                        cancellationToken);
+                }
+            }
+        }
+
+        private async Task<HashSet<FeedPackageIdentity>> GetValidatedPackages(CancellationToken cancellationToken)
+        {
+            var validatedPackagesList = await _statusService.ListAsync(cancellationToken);
+            return new HashSet<FeedPackageIdentity>(validatedPackagesList.Select(p => p.Package).ToList());
```
loic-sharma: >ToList [](start = 97, length = 6)

Is the `ToList` necessary? This forces the `validatedPackagesList` to be iterated twice: once for the `ToList`, and a second time to add each package to the `HashSet`.

src/Ng/Jobs/Feed2MonitoringJob.cs
```diff
@@ -0,0 +1,190 @@
...
+
+            var invalidPackages = new ConcurrentQueue<FeedPackageIdentity>();
+            _pagesProcessed = 0;
+
+            await ParallelExecutionUtility.ProcessInParallel(() => ProcessPackages(validPackagesSet, invalidPackages, cancellationToken), _degreeOfParallelism);
+
+            if (!invalidPackages.IsEmpty)
+            {
+                var invalidPackagesList = new List<FeedPackageIdentity>();
+                Logger.LogError("The following packages are missing:");
```
loic-sharma: >LogError [](start = 23, length = 8)

I think that seeing a bunch of error logs with just ids and versions may be confusing. Maybe change the first `LogError` to "<X> packages are missing." and the second `LogError` message to "{id} {version} is missing."

src/Ng/Jobs/Feed2MonitoringJob.cs
```diff
@@ -0,0 +1,190 @@
...
+                while (invalidPackages.TryDequeue(out var invalidPackage))
+                {
+                    invalidPackagesList.Add(invalidPackage);
+                    Logger.LogError($"{invalidPackage.Id} {invalidPackage.Version}");
```
loic-sharma: >LogError [](start = 27, length = 8)

I wouldn't format the log strings - loggers often have special optimizations/laziness for string formatting.

src/Ng/Jobs/Feed2MonitoringJob.cs
```diff
@@ -0,0 +1,190 @@
...
+        {
+            var requestSuccessful = false;
+            var failures = 0;
+
+            T result = null;
+            while (!requestSuccessful || failures > _maxRequestFailures)
+            {
+                try
+                {
+                    result = await action();
```
loic-sharma: >result [](start = 20, length = 6)

Nit: you could just return the result of the action directly, get rid of the `requestSuccessful` variable, and return `null` on line 187.

scottbommarito:
Decided to close this PR because it's not currently necessary and would require work to fix. In the future when these ops tools are needed, we can look back into this branch and fix it.

PR #246, opened 2017-10-05
Cmanuissue823

cristinamanum:
Add logging for:  correct storage and operation execution time. 

CI build:
https://nuget.visualstudio.com/NuGetBuild/_build/index?buildId=15260&_a=summary


src/Catalog/Persistence/Storage.cs
```diff
@@ -191,8 +205,23 @@ protected void TraceMethod(string method, Uri resourceUri)
...
+        private string TraceException(string method, Uri resourceUri, Exception exception)
+        {
+            string message = $"{method} EXCEPTION: {GetUri(GetName(resourceUri))} {exception.Message}";
+            Trace.WriteLine(message);
+            return message;
+        }
+
+        private void TraceExecutionTime(string method, Uri resourceUri, long executionTimeInMilliseconds)
+        {
+            string message = $"ExecutionTime_Method:{method}_Uri:{GetUri(GetName(resourceUri))}_Milliseconds:{executionTimeInMilliseconds}";
```
joelverhagen: Why is this encoded so?

cristinamanum: The goal is to have a string format to be parsed in AI. The goal is to have a string format to be parsed in AI.

joelverhagen: Spoke offline. @cristinamanum is going to check is JSON serialization is easier from the AI side.

src/Catalog/Persistence/Storage.cs
```diff
@@ -191,8 +205,23 @@ protected void TraceMethod(string method, Uri resourceUri)
...
-                Trace.WriteLine(String.Format("{0} {1}", method, resourceUri));
+                //The Uri depends on the storage implementation.
+                Uri storageUri = GetUri(GetName(resourceUri));
+                Trace.WriteLine(String.Format("{0} {1}", method, storageUri));
             }
         }
+
+        private string TraceException(string method, Uri resourceUri, Exception exception)
+        {
+            string message = $"{method} EXCEPTION: {GetUri(GetName(resourceUri))} {exception.Message}";
```
loic-sharma: You should log the full exception instead of just its `Message` so that `InnerException`s are logged too.

cristinamanum: Updated.

src/Catalog/Persistence/Storage.cs
cristinamanum: Updated.

src/Catalog/Persistence/Storage.cs
```diff
@@ -191,8 +205,24 @@ protected void TraceMethod(string method, Uri resourceUri)
...
+        {
+            string message = $"{method} EXCEPTION: {GetUri(GetName(resourceUri))} {exception.ToString()}";
+            Trace.WriteLine(message);
+            return message;
+        }
+
+        private void TraceExecutionTime(string method, Uri resourceUri, long executionTimeInMilliseconds)
+        {
+            string message =  Newtonsoft.Json.JsonConvert.SerializeObject(new { MethodName = $"{method}", StreamUri = $"{GetUri(GetName(resourceUri))}", ExecutionTimeInMilliseconds = executionTimeInMilliseconds });
+           // string message = $"ExecutionTime_Method:{method}_Uri:{GetUri(GetName(resourceUri))}_Milliseconds:{executionTimeInMilliseconds}";
```
joelverhagen: Delete commented code.
Add `using Newtonsoft.Json`.

src/Catalog/Persistence/Storage.cs
```diff
@@ -93,10 +105,12 @@ public override string ToString()
             }
             catch (Exception e)
             {
-                string message = String.Format("DELETE EXCEPTION: {0} {1}", resourceUri, e.Message);
-                Trace.WriteLine(message);
+                var message = TraceException(nameof(Delete), resourceUri, e);
                 throw new Exception(message, e);
```
joelverhagen: Can we just rethrow here `throw;`? The stack on the outermost exception will be wrong if we don't rethrow.

src/Catalog/Persistence/Storage.cs
```diff
@@ -191,8 +205,23 @@ protected void TraceMethod(string method, Uri resourceUri)
...
-                Trace.WriteLine(String.Format("{0} {1}", method, resourceUri));
+                //The Uri depends on the storage implementation.
+                Uri storageUri = GetUri(GetName(resourceUri));
+                Trace.WriteLine(String.Format("{0} {1}", method, storageUri));
             }
         }
+
+        private string TraceException(string method, Uri resourceUri, Exception exception)
+        {
+            string message = $"{method} EXCEPTION: {GetUri(GetName(resourceUri))} {exception.ToString()}";
```
loic-sharma: I personally prefer letting C#'s string formatting implicitly convert the exception to a string (`{exception}` instead of `{exception.ToString()}`). Feel free to ignore though :)

PR #249, opened 2017-10-18
[HotFix]Catalog2lucene fix for index out of date

shishirx34:


src/NuGet.Indexing/Extraction/CatalogNuspecReader.cs
```diff
@@ -41,6 +41,11 @@ private static XDocument CreateXDocument(JObject catalogItem)
                 foreach (var dependency in group.GetJArray("dependencies"))
                 {
                     var dependencyElement = new XElement("dependency");
+                    if (string.IsNullOrWhiteSpace((string)dependency["id"]))
```
skofman1: can u add a comment on the behavior and connect to an issue in GitHub

PR #251, opened 2017-10-20
In feed2catalog, ignore dependencies that have a missing, empty, or whitespace ID

joelverhagen:
Fix https://github.com/NuGet/NuGetGallery/issues/4494


edouardp:
Operational question on behalf of package maintainers: Will this fix be automatically applied to existing feeds, or will package maintainers (that have specs with empty id's, for example) have to push a new version of their package to have this code executed for their package feed on nuget.org?

For example https://api.nuget.org/v3/registration3-gz-semver2/baseline/index.json still had empty id's - will this be regenerated automatically, or only when a new package is released?

(I'm probably getting some of those terms wrong, but hopefully you understand the question - I'll try to clarify if not)

joelverhagen:
@edouardp, great question and thanks for the attention to this detail. NuGet.org team typically takes the necessary action to fully "deploy" the fix. In this case, we will be reflowing the four effected packages through the pipeline which will result in the a corrected registration JSON file (the one linked above).

No further action is needed on the part of the package maintainers.

PR #253, opened 2017-10-24
 In feed2catalog, default to empty string if the version range is missing or invalid

joelverhagen:
Fix https://github.com/NuGet/NuGetGallery/issues/3482.


joelverhagen:
Before I take this to PROD, I am going to do a scan of all invalid version ranges so that we know what we have to reflow.

PR #256, opened 2017-11-09
Add CaseInsensitiveVersion field to the Lucene index for case-insensitive version comparison

joelverhagen:
This is step 1 in fixing https://github.com/NuGet/NuGetGallery/issues/4583.

Currently, the `PackageVersion` analyzer normalizes the package version but does not lowercase it! This means that `version:` queries (most notably hijacked V2 request for a specific ID/version).

This fix enables case insensitivity on the package version by adding a new field to the Lucene documents which uses `LowerCaseFilter` (which is the de facto way to do case insensitivity in Lucene).

Note that this PR does not enable the use of the new field in the search service. This must be done in a subsequent deployment. As this is not high priority, I am taking this more risk-adverse way of rolling out the change which supports easy roll-back.

After `catalog2lucene` is deployed, we must rebuild the index using `db2lucene`.

A subsequent PR and deployment will make use of this new field in the search service.

Also, I used our Lucene field name constants in more places.


joelverhagen:
This new field makes increases the index size from 3.36 to 3.41.

src/NuGet.Indexing/Extraction/MetadataConstants.cs
```diff
@@ -17,6 +17,7 @@ public static class LuceneMetadata
             public const string NormalizedVersionPropertyName = "Version";
             public const string FullVersionPropertyName = "FullVersion";
             public const string VerbatimVersionPropertyName = "OriginalVersion";
+            public const string CaseInsensitiveNormalizedVersionPropertyName = "CaseInsensitiveVersion";
```
scottbommarito: Could `CaseInsensitiveNormalizedVersion` be a better string value here?

src/NuGet.Indexing/PackageAnalyzer.cs
```diff
@@ -15,17 +17,18 @@ static PackageAnalyzer()
...
+                { LuceneConstants.IdAutocompletePropertyName, new IdentifierAutocompleteAnalyzer() },
+                { LuceneConstants.TokenizedIdPropertyName, new IdentifierAnalyzer() },
+                { LuceneConstants.ShingledIdPropertyName, new ShingledIdentifierAnalyzer() },
+                { LuceneConstants.NormalizedVersionPropertyName, new VersionAnalyzer(caseSensitive: true) },
+                { LuceneConstants.CaseInsensitiveNormalizedVersionPropertyName, new VersionAnalyzer(caseSensitive: false) },
+                { LuceneConstants.TitlePropertyName, new DescriptionAnalyzer() },
+                { LuceneConstants.DescriptionPropertyName, new DescriptionAnalyzer() },
+                { LuceneConstants.SummaryPropertyName, new DescriptionAnalyzer() },
+                { LuceneConstants.AuthorsPropertyName, new DescriptionAnalyzer() },
                 { "Owner", new OwnerAnalyzer() },
```
scottbommarito: Any reason for not adding a constant for `Owner`?

src/NuGet.Indexing/VersionAnalyzer.cs
```diff
@@ -1,15 +1,30 @@
...
+
+        public VersionAnalyzer(bool caseSensitive)
+        {
+            _caseSensitive = caseSensitive;
+        }
+
         public override TokenStream TokenStream(string fieldName, TextReader reader)
         {
-            return new SemanticVersionFilter(new KeywordTokenizer(reader));
+            TokenStream stream =  new SemanticVersionFilter(new KeywordTokenizer(reader));
```
scottbommarito: nit: `var`?

src/NuGet.Indexing/PackageAnalyzer.cs
joelverhagen: I could not find any other places this was used (in the same way). Investigation is out of scope.

src/NuGet.Indexing/Extraction/MetadataConstants.cs
joelverhagen: Yes, I like that better.

src/NuGet.Indexing/NuGetQuery.cs
```diff
@@ -38,7 +39,7 @@ public static Query MakeAutoCompleteQuery(string q)
                 return new MatchAllDocsQuery();
             }
 
-            return ExecuteAnalyzer(new PackageAnalyzer(), "IdAutocomplete", q);
+            return ExecuteAnalyzer(new PackageAnalyzer(), LuceneConstants.IdAutocompletePropertyName, q);
```
shishirx34: 👍 on constants 😄

src/NuGet.Indexing/VersionAnalyzer.cs
joelverhagen: `LowerCaseFilter` cannot be assigned to `SemanticVersionFilter`.

src/NuGet.Indexing/PackageAnalyzer.cs
joelverhagen: Turns out it's unused. I removed it.

PR #257, opened 2017-11-21
Merge dev into master

joelverhagen:
This brings the phase 1 for case insensitive versions.


joelverhagen:
@skofman1, I split the related issue into two pieces so that we can track `Verified-Dev` at a more granular level:
https://github.com/NuGet/NuGetGallery/issues/5052 - addressed by this merge
https://github.com/NuGet/NuGetGallery/issues/5051 - not done yet

PR #258, opened 2017-11-22
When searching, use the case-insensitive package version field in the Lucene document

joelverhagen:
Fix https://github.com/NuGet/NuGetGallery/issues/5051.

Added some tests, and fixed a previously skipped test that was just plain wrong.

I will only merge this after https://github.com/NuGet/NuGetGallery/issues/5052 has been deployed (along with https://github.com/NuGet/NuGetGallery/issues/5010).

### DIFF
--- a/tests/NuGet.Services.BasicSearchTests/V2SearchFunctionalTests.cs
+++ b/tests/NuGet.Services.BasicSearchTests/V2SearchFunctionalTests.cs
@@ -13,7 +13,7 @@ using Xunit;
 
 namespace NuGet.Services.BasicSearchTests
 {
-    public class V2SearchSkipFunctionalTests
+    public class V2SearchFunctionalTests
     {
         [Fact]
         public async Task CanReturnEmptyResult()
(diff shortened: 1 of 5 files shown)


tests/NuGet.Services.BasicSearchTests/V2SearchFunctionalTests.cs
```diff
@@ -37,7 +37,47 @@ public class V2SearchSkipFunctionalTests
...
+                var partialIdResponse = await app.Client.GetAsync(new V2SearchBuilder
+                {
+                    Query = "version:1.6.0 id:Web",
+                    Prerelease = true,
+                }.RequestUri);
+                var partialId = await partialIdResponse.Content.ReadAsAsync<V2SearchResult>();
+
+                var wrongExactIdResponse = await app.Client.GetAsync(new V2SearchBuilder
+                {
+                    Query = "id:Web version:1.6.0",
```
loic-sharma: Why don't we support partial id matches here?

joelverhagen: Gallery expects this to work as exact match for V2 hijacking, e.g. `/api/v2/Packages(Id='Web',Version='1.6.0')`

PR #259, opened 2017-11-27
 Fix all analyzers to use LowerInvariantFilter (ToLowerInvariant instead of ToLower)

joelverhagen:
Fix https://github.com/NuGet/Engineering/issues/958.

Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/258.


joelverhagen:
Deployment process:

1. Make copy old catalog2lucene on jobs machine
2. New db2lucene to generate a new index
3. Upload new index to blob storage
4. Deploy new catalog2lucene pointed at new index
5. Run backed up catalog2lucene against old index location
6. Deploy new search pointed at new index
7. Wait 24 hours and deploy ussc
8. Wait 24 hours and turn off old catalog2lucene

PR #263, opened 2017-12-08
Perf improvement for catalog2dnx

cristinamanum:


src/Catalog/Persistence/Storage.cs
```diff
@@ -170,6 +170,18 @@ public Uri ResolveUri(string relativeUri)
...
+        /// <summary>
+        /// It will return true if there are changes between the source and destination.
+        /// </summary>
+        /// <param name="resourceUri">The uri for the file source.</param>
+        /// <param name="id">The id of the source.</param>
+        /// <param name="version">The version of the source.</param>
+        /// <returns></returns>
+        public virtual Task<bool> IsUpdateRequiredAsync(Uri resourceUri, string id, string version)
+        {
+            return Task<bool>.FromResult(true);
```
joelverhagen: Type parameter is not required.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -283,5 +283,24 @@ public override Uri GetUri(string name)
         {
             return new Uri(_directory.Uri, name);
         }
+
+        public override async Task<bool> IsUpdateRequiredAsync(Uri resourceUri, string id, string version)
+        {
+            //the destination blob
+            var blobName = resourceUri.Segments.Last();
+            var blobRelativePath = $"{id}/{version}/{blobName}";
```
joelverhagen: This is leaking knowledge of the destination URL from `DnxMaker`. The URL create should be shared.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -283,5 +283,24 @@ public override Uri GetUri(string name)
...
+            var destination = _directory.GetBlockBlobReference(blobRelativePath);
+            if (await destination.ExistsAsync())
+            {
+                await destination.FetchAttributesAsync();
```
joelverhagen: Do we really need `FetchAttributesAsync` if we just called `ExistsAsync`? They should both be possible with a single HTTP request.

src/Catalog/Dnx/DnxCatalogCollector.cs
```diff
@@ -41,13 +44,21 @@ public DnxCatalogCollector(Uri index, StorageFactory storageFactory, Func<HttpMe
...
                     {
-                        var requestUri = Utilities.GetNugetCacheBustingUri(new Uri(ContentBaseAddress, string.Format("{0}.{1}.nupkg", id, version)));
-                        using (Stream stream = await client.GetStreamAsync(requestUri))
+                        Uri sourceUri= new Uri(ContentBaseAddress, string.Format("{0}.{1}.nupkg", id, version));
+                        var requestUri = Utilities.GetNugetCacheBustingUri(sourceUri);
+                        if (await storage.IsUpdateRequiredAsync(sourceUri, id, version))
                         {
-                            await _dnxMaker.AddPackage(stream, nuspec, id, version, cancellationToken);
-                        }
+                            using (Stream stream = await client.GetStreamAsync(requestUri))
```
joelverhagen: `GetStreamAsync` should also determine `ExistsAsync` and the `Content-MD5` header. We don't need this many HTTP requests.

joelverhagen:
This needs unit/functional tests. Since this is a hotfix, that can wait but should be done after the hotfix is out... otherwise it will never happen.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -124,7 +124,7 @@ private StorageListItem GetStorageListItem(IListBlobItem listBlobItem)
             CloudBlockBlob blob = _directory.GetBlockBlobReference(name);
             blob.Properties.ContentType = content.ContentType;
             blob.Properties.CacheControl = content.CacheControl;
-
+
```
scottbommarito: nit: pls remove excess whitespace

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -283,5 +283,24 @@ public override Uri GetUri(string name)
         {
             return new Uri(_directory.Uri, name);
         }
+
+        public override async Task<bool> IsUpdateRequiredAsync(Uri resourceUri, string id, string version)
```
scottbommarito: These APIs are meant to be a general purpose storage API--not specific to NuGet or packaging. `AzureStorage` should know nothing about `id`s or `version`s--it should just accept a `resourceUri`. Additionally, by adding package information to these calls, you get issues such as the one that @joelverhagen calls out below where we're duplicating code from `DnxMaker`.

I would suggest rewriting this method to take in just `resourceUri` and rewriting `DnxMaker` to construct the `resourceUri`.

src/Catalog/Persistence/Storage.cs
```diff
@@ -170,6 +170,18 @@ public Uri ResolveUri(string relativeUri)
...
         }
 
+        /// <summary>
+        /// It will return true if there are changes between the source and destination.
+        /// </summary>
+        /// <param name="resourceUri">The uri for the file source.</param>
+        /// <param name="id">The id of the source.</param>
+        /// <param name="version">The version of the source.</param>
+        /// <returns></returns>
+        public virtual Task<bool> IsUpdateRequiredAsync(Uri resourceUri, string id, string version)
```
scottbommarito: I'm not a fan of this method. It seems very specific to a Catalog2Dnx scenario and I'm not sure where else it would be used.

IMO we should make a method that returns the results of `FetchAttributesAsync` and then have the logic for comparing the hashes inside Catalog2Dnx. If we discover that logic is needed elsewhere, we can create some sort of helper/utility method, but I don't think comparing MD5 hashes belongs in this storage API.

src/Catalog/Persistence/Storage.cs
cristinamanum: Updated

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Updated

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Updated

src/Catalog/Dnx/DnxCatalogCollector.cs
cristinamanum: As we discussed offline I will add a task to track the reduction of the http requests. The current change will add one more in case that changes are required.

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Updated to use DnxMarker

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Updated.

src/Catalog/Dnx/DnxCatalogCollector.cs
cristinamanum: https://github.com/NuGet/Engineering/issues/1026

cristinamanum: Agree with the below.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -283,5 +282,20 @@ public override Uri GetUri(string name)
         {
             return new Uri(_directory.Uri, name);
         }
+
+        public override async Task<bool> AreSyncronized(Uri sourceUri, Uri destinationUri)
+        {
+            var destination = new CloudBlockBlob(destinationUri);
```
joelverhagen: This assumes public blob right? I think you should be calling `GetBlockBlobReference` instead of the constructor. You can fix this easily-ish by accepting `sourceUri`'s `Storage` as a parameter.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -283,5 +282,20 @@ public override Uri GetUri(string name)
...
+            if (await destination.ExistsAsync())
+            {
+                var source = new CloudBlockBlob(sourceUri);
+                if (await source.ExistsAsync())
+                {
+                    return !string.IsNullOrEmpty(source.Properties.ContentMD5) && source.Properties.ContentMD5 == destination.Properties.ContentMD5;
+                }
+                return true;
+            }
+            return false;
```
joelverhagen: If both don't exist, then they are synchronized right?

src/Catalog/Dnx/DnxCatalogCollector.cs
```diff
@@ -34,24 +36,33 @@ public DnxCatalogCollector(Uri index, StorageFactory storageFactory, Func<HttpMe
...
-                    string nuspec = await LoadNuspec(client, id, version, cancellationToken);
-                    if (nuspec != null)
+                    var sourceUri = new Uri(ContentBaseAddress, string.Format("{0}.{1}.nupkg", id, version));
+                    var destinationUri = destinationStorage.GetUri(DnxMaker.GetRelativeAddressNupkg(id, version));
+                    if (!await destinationStorage.AreSyncronized(sourceUri, destinationUri))
                     {
-                        var requestUri = Utilities.GetNugetCacheBustingUri(new Uri(ContentBaseAddress, string.Format("{0}.{1}.nupkg", id, version)));
-                        using (Stream stream = await client.GetStreamAsync(requestUri))
+                        // Add/update package
+                        string nuspec = await LoadNuspec(client, id, version, cancellationToken);
```
joelverhagen: Can we combine the `LoadNuspec` and `GetStreamAsync` download? If we save to a temp file on disk, then we should only need to download once.

src/Catalog/Persistence/Storage.cs
cristinamanum: Catalog2DNX needs to support multiple Storage types. Currently these are FiltSystem and Azure.
For this reason I think that the logic for deciding if the two resources are out of sync and the destination needs to be updated needs to stay at the Storage level. I updated the method name and signature.

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: For destination we can do that. For source we cannot with the current design - Catalog2DNX works under the assumption that the source is public.

src/Catalog/Persistence/Storage.cs
```diff
@@ -170,6 +170,17 @@ public Uri ResolveUri(string relativeUri)
             return new Uri(BaseAddress, relativeUri);
         }
 
+        /// <summary>
+        /// It will return true if there are changes between the source and destination.
```
scottbommarito: Isn't it the opposite? The implementation above returns true if no changes and false if changes.

src/Catalog/Persistence/Storage.cs
```diff
@@ -170,6 +170,17 @@ public Uri ResolveUri(string relativeUri)
...
+        /// </summary>
+        /// <param name="sourceUri">The soource uri.</param>
+        /// <param name="destinationUri">The destination uri.</param>
+        /// <returns></returns>
+        public virtual Task<bool> AreSyncronized(Uri sourceUri, Uri destinationUri)
```
scottbommarito: `AreSyncronized` -> `AreSynchronized`

src/Catalog/Persistence/AzureStorage.cs
joelverhagen: Okay, sounds good.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -283,5 +282,20 @@ public override Uri GetUri(string name)
         {
             return new Uri(_directory.Uri, name);
         }
+
+        public override async Task<bool> AreSyncronized(Uri sourceUri, Uri destinationUri)
+        {
+            var destination = new CloudBlockBlob(destinationUri);
+            if (await destination.ExistsAsync())
```
scottbommarito: You can write this in a way that unnests it a lot

```
var source = new CloudBlockBlob(sourceUri);
var destination = new CloudBlockBlob(destinationUri);

var sourceExists = await source.ExistsAsync();
if (!await destination.ExistsAsync())
{
    return !sourceExists;
}

return sourceExists && 
    !string.IsNullOrEmpty(source.Properties.ContentMD5) && 
    source.Properties.ContentMD5 == destination.Properties.ContentMD5;
```

src/Catalog/Persistence/Storage.cs
scottbommarito: Add this to `IStorage` as well.

src/Catalog/Persistence/Storage.cs
```diff
@@ -170,6 +170,17 @@ public Uri ResolveUri(string relativeUri)
...
+        {
+            return Task.FromResult(true);
```
scottbommarito: This should be false by default. Having this as true breaks Catalog2Dnx with `FileStorage` because it will always think that it doesn't need to save the file because it's synchronized.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -283,5 +282,20 @@ public override Uri GetUri(string name)
         {
             return new Uri(_directory.Uri, name);
         }
+
+        public override async Task<bool> AreSyncronized(Uri sourceUri, Uri destinationUri)
```
scottbommarito: These aren't really a `sourceUri` and a `destinationUri`. For Catalog2Dnx, you're copying one to the other, but outside of that context, this method is identical regardless of which is which.

IMO something like `Uri firstResourceUri` and `Uri secondResourceUri` would make more sense.

scottbommarito:
We need strong unit tests on this. The V3 monitoring pipeline doesn't run on flat-container so not copying blobs could lead to some really cryptic bugs that are not monitored.

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Good catch - updated.

src/Catalog/Dnx/DnxCatalogCollector.cs
cristinamanum: This changes will introduce more risk that I propose to not take in this hotfix. I opened a task for future improvements. I added this suggestion to the task/issue.

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: Updated with the suggested names.

src/Catalog/Persistence/Storage.cs
cristinamanum: Correct - updated.

PR #266, opened 2018-02-06
Merge dev into master

joelverhagen:
This fixes:

- Search service treat versions as case sensitive strings (https://github.com/NuGet/NuGetGallery/issues/4583)
- Package with Turkish "i" collides with ASCII "i" in V2 search results (https://github.com/NuGet/Engineering/issues/958)

This will require an out-of-band deployment I will be driving:

1. Make copy old catalog2lucene on jobs machine
2. Use new db2lucene to generate a new index
    - Download the new Ng package from Octopus and extract it to C:\nuget\LuceneIndexCreation
    - Follow the standard index rebuild steps
    - The script uploads the new indexes (both primary and secondary) to blob storage
4. Deploy new catalog2lucene pointed at new index
    - This requires updating the catalog2lucene (both primary and secondary) configuration to point to the new index location
5. Run backed up catalog2lucene against old index location
    - Remember, both primary and secondary! This step should be done very quickly after the deployment of step 4 completes (during e2e tests).
6. Deploy new search pointed at new index
7. Wait 24 hours and deploy ussc
8. Wait 24 hours and turn off old catalog2lucene

### DIFF
--- a/src/NuGet.Indexing/DescriptionAnalyzer.cs
+++ b/src/NuGet.Indexing/DescriptionAnalyzer.cs
@@ -1,8 +1,8 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-using Lucene.Net.Analysis;
-using System.Collections.Generic;
+
 using System.IO;
+using Lucene.Net.Analysis;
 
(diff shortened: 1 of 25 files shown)


joelverhagen:
I've verified the changes and deployment process on DEV.

joelverhagen:
🔔

PR #269, opened 2018-02-21
Add ITelemetryService and TelemetryHandler for logging HTTP request durations

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/5527.

Sorry about the huge size. This project does not have DI so passing things down requires a lot of constructor changes. Also, I removed dead code that would otherwise need to be changed. I figured removing dead code right now was better than changing it to match API changes.

I made the dead code removal its own commit.

This PR adds a HTTP message handler which emits the duration of an HTTP request as a AI metric. These metrics will tell us how long it takes to download the catalog index. I will follow up with another PR for how long it takes to deserialize the catalog index and how long it takes to write the catalog index.


src/Catalog/CollectorHttpClient.cs
```diff
@@ -93,6 +89,7 @@ public virtual Task<IGraph> GetGraphAsync(Uri address, bool readOnly, Cancellati
 
         public virtual Task<string> GetStringAsync(Uri address, CancellationToken token)
         {
+            var stopwatch = Stopwatch.StartNew();
```
loic-sharma: >var stopwatch = Stopwatch.StartNew(); [](start = 12, length = 37)

Looks unused.

joelverhagen: Yup, nice catch.

src/Catalog/CollectorBase.cs
```diff
@@ -12,17 +12,14 @@ namespace NuGet.Services.Metadata.Catalog
 {
     public abstract class CollectorBase
     {
+        private readonly ITelemetryService _telemetryService;
```
loic-sharma: >private [](start = 8, length = 7)

Talked to Joel offline, this will become protected later.

joelverhagen: Basically, this needs to be protected to be useful. In a follow-up PR I have done this.

joelverhagen: Fixed.

src/Catalog/CollectorHttpClient.cs
joelverhagen: Fixed.

src/Catalog/CollectorBase.cs
```diff
@@ -12,17 +12,14 @@ namespace NuGet.Services.Metadata.Catalog
...
 
-        public CollectorBase(Uri index, Func<HttpMessageHandler> handlerFunc = null)
+        public CollectorBase(Uri index, ITelemetryService telemetryService, Func<HttpMessageHandler> handlerFunc = null)
         {
-            if (index == null)
-            {
-                throw new ArgumentNullException(nameof(index));
-            }
-
+            _telemetryService = telemetryService;
```
skofman1: null check

src/Catalog/TelemetryHandler.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace NuGet.Services.Metadata.Catalog
+{
+    public class TelemetryHandler : DelegatingHandler
+    {
+        private readonly ITelemetryService _telemetryService;
+
+        public TelemetryHandler(ITelemetryService telemetryService, HttpMessageHandler innerHandler) : base(innerHandler)
+        {
+            _telemetryService = telemetryService;
```
skofman1: null check

src/Catalog/CollectorBase.cs
joelverhagen: Fixed.

src/Catalog/TelemetryHandler.cs
joelverhagen: Fixed.

PR #270, opened 2018-02-22
Add TrackCatalogIndexReadDuration and TrackCatalogIndexWriteDuration

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/5527

Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/269.


src/Ng/Program.cs
```diff
@@ -47,6 +46,14 @@ public static void Main(string[] args)
                 ServicePointManager.SecurityProtocol &= ~SecurityProtocolType.Ssl3;
                 ServicePointManager.SecurityProtocol |= SecurityProtocolType.Tls12;
 
+                // Determine the job name
+                if (args.Length == 0)
+                {
+                    throw new ArgumentException("Missing tool specification");
```
loic-sharma: >tool specification [](start = 57, length = 18)

Maybe this would be an better error message: "Missing job name argument"?

src/Ng/Program.cs
```diff
@@ -59,17 +66,10 @@ public static void Main(string[] args)
...
-                if (args.Length == 0)
-                {
-                    throw new ArgumentException("Missing tool specification");
-                }
 
                 // Create an ITelemetryService
                 var telemetryService = new TelemetryService(new TelemetryClient());
 
-                var jobName = args[0];
-                LogContext.PushProperty("JobName", jobName);
```
loic-sharma: >LogContext [](start = 16, length = 10)

Why was this removed? Does it not work?

joelverhagen: Yes, it works, but only on traces. I added a more general solution for all AI events.

src/Ng/Program.cs
joelverhagen: Sure

joelverhagen: Fixed.

PR #276, opened 2018-04-03
[SearchService] Return 503 for uninitialized

shishirx34:
A quick fix for uninitialized state of the service.


skofman1:
Do we know how this endpoint is used what will be the impact of this change?

shishirx34:
Fixes https://github.com/NuGet/Engineering/issues/1174

shishirx34:
@skofman1 - This will return 503 when the service is not running("Uninitialized"). There shouldn't be any impact of this change. I will validate it on DEV search environment.

PR #278, opened 2018-04-06
Increase HTTP client timeout in catalog2dnx

joelverhagen:
Address https://github.com/NuGet/Engineering/issues/1261.

- Add `-httpClientTimeoutInSeconds` option to catalog2dnx.
- Only download the .nupkg once
- Download the .nupkg to disk and use this stream for getting .nuspec and for uploading to flat container
- Improve the tests a lot
- Only accept the HTTP response if it is 200 OK. We previously allowed any 2XX.

### DIFF
--- a/src/Catalog/Dnx/DnxCatalogCollector.cs
+++ b/src/Catalog/Dnx/DnxCatalogCollector.cs
@@ -9,6 +9,7 @@ using System.IO.Compression;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
 using Newtonsoft.Json.Linq;
 using NuGet.Services.Metadata.Catalog.Helpers;
 using NuGet.Services.Metadata.Catalog.Persistence;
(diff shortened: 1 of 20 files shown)


joelverhagen:
Oh, I set the HTTP client timeout to 900 seconds. This is because the download of a 250 megabyte file takes approximately 7 minutes on our jobs machine. I doubled-ish this value to 15 minutes.

src/Catalog/PackageDownloader.cs
```diff
@@ -0,0 +1,96 @@
...
+namespace NuGet.Services.Metadata.Catalog
+{
+    public class PackageDownloader
+    {
+        private const int BufferSize = 8192;
+
+        private readonly HttpClient _httpClient;
+        private readonly ILogger _logger;
+
+        public PackageDownloader(HttpClient httpClient, ILogger logger)
```
joelverhagen: This is a copy of `PackageDownloader` from NuGet.Jobs. I have removed some custom telemetry events and returned `null` on 404.

src/Catalog/PackageDownloader.cs
```diff
@@ -0,0 +1,96 @@
...
+                    packageStream.Length,
+                    stopwatch.Elapsed.TotalSeconds,
+                    packageUri);
+
+                return packageStream;
+            }
+            catch (Exception e)
+            {
+                _logger.LogError(
+                    0,
```
shishirx34: parameter name

src/Catalog/Dnx/DnxCatalogCollector.cs
```diff
@@ -36,67 +45,67 @@ public DnxCatalogCollector(Uri index, StorageFactory storageFactory, ITelemetryS
...
+
+            var packageDownloader = new PackageDownloader(client, _logger);
+            var requestUri = Utilities.GetNugetCacheBustingUri(sourceUri);
+
+            using (var stream = await packageDownloader.DownloadAsync(requestUri, cancellationToken))
             {
-                httpResponseMessage.EnsureSuccessStatusCode();
+                if (stream == null)
+                {
+                    Trace.TraceInformation("package not found...");
```
loic-sharma: >TraceInformation [](start = 26, length = 16)

Should this be a warning?

src/Catalog/PackageDownloader.cs
joelverhagen: I'll do `(EventId)0`. This seems like a good middle ground to know what that `0` means and not having to name all further parameters.

src/Catalog/Dnx/DnxCatalogCollector.cs
joelverhagen: Wasn't before, but I can switch this.

PR #279, opened 2018-04-07
Don't no-op added package in catalog2dnx if version is missing

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/5473.
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/278.

If the added package version is not in the `{id}/index.json`, then don't no-op the add. This can happen if on first attempt that .nupkg is uploaded properly but the .nuspec upload or index.json upload fails and the job retries.

This does make catalog2dnx slower but ideally the index.json download is relatively small compared to the .nupkg upload so we still get a big win.

This is one instance of this happening: https://github.com/NuGet/NuGetGallery/issues/5720

### DIFF
--- a/src/Catalog/Dnx/DnxCatalogCollector.cs
+++ b/src/Catalog/Dnx/DnxCatalogCollector.cs
@@ -70,35 +70,58 @@ namespace NuGet.Services.Metadata.Catalog.Dnx
             var destinationStorage = _storageFactory.Create(id);
             var destinationUri = destinationStorage.GetUri(DnxMaker.GetRelativeAddressNupkg(id, version));
 
-            if (await destinationStorage.AreSyncronized(sourceUri, destinationUri))
+            var isNupkgSynchronized = await destinationStorage.AreSynchronized(sourceUri, destinationUri);
+            if (isNupkgSynchronized
+                && await _dnxMaker.HasPackageInIndex(destinationStorage, id, version, cancellationToken))
             {
                 _logger.LogInformation("No changes detected: {Id}/{Version}", id, version);
                 return;
             }
 
-            var packageDownloader = new PackageDownloader(client, _logger);
-            var requestUri = Utilities.GetNugetCacheBustingUri(sourceUri);
-
-            using (var stream = await packageDownloader.DownloadAsync(requestUri, cancellationToken))
+            if (isNupkgSynchronized)
             {
-                if (stream == null)
-                {
-                    _logger.LogWarning("Package {Id}/{Version} not found.", id, version);
-                    return;
-                }
+                _logger.LogInformation(
+                    "The .nuspec and .nupkg for {Id}/{Version} are already uploaded. Updating just the index.json.",
+                    id,
+                    version);
 
-                var nuspec = GetNuspec(stream, id);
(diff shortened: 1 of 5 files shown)


src/Catalog/Dnx/DnxCatalogCollector.cs
```diff
@@ -70,7 +70,8 @@ public class DnxCatalogCollector : CommitCollector
             var destinationStorage = _storageFactory.Create(id);
             var destinationUri = destinationStorage.GetUri(DnxMaker.GetRelativeAddressNupkg(id, version));
 
-            if (await destinationStorage.AreSyncronized(sourceUri, destinationUri))
+            if (await destinationStorage.AreSyncronized(sourceUri, destinationUri)
+                && await _dnxMaker.HasPackageInIndex(destinationStorage, id, version, cancellationToken))
```
shishirx34: I think we can further optimize and simply call into the `UpdateMetadata` only as needed, rather than re-download the nupkg and nuspec which don't need an update. Store the `AreSynchronized` result and skip `nupkg` update based on this value in the `AddPackage`, you can refactor the `AddPackage` method for this if needed. I think it would be a good optimization to have, given we have seen so many incidents around this code path.

joelverhagen: > I think it would be a good optimization to have, given we have seen so many incidents around this code path.

I'm not sure if I follow. The only time this optimization will make a difference is when then .nupkg is there but the version is not in the index.json. This has only ever happened once in PROD so I'd prefer to keep the code simpler.

shishirx34: Oh, I meant the downloads timing out for large size package. If the issue for failure to update index.json were to happen for a large sized package, we would essentially be spending another 10-15 mins for downloading a nupkg which doesn't need to be updated.

joelverhagen: Yeah, that's a good point. If we can save this, we should. I'll fix this up.

joelverhagen: Fixed.

src/Catalog/Dnx/DnxCatalogCollector.cs
```diff
@@ -70,35 +70,58 @@ public class DnxCatalogCollector : CommitCollector
             var destinationStorage = _storageFactory.Create(id);
             var destinationUri = destinationStorage.GetUri(DnxMaker.GetRelativeAddressNupkg(id, version));
 
-            if (await destinationStorage.AreSyncronized(sourceUri, destinationUri))
+            var isNupkgSynchronized = await destinationStorage.AreSyncronized(sourceUri, destinationUri);
```
loic-sharma: >AreSyncronized [](start = 63, length = 14)

Could you rename this method: `Syncronized` -> `Synchronized` :)

joelverhagen: Sure

joelverhagen: Done.

PR #284, opened 2018-04-17
[SDL] Validate secrets' certificate

loic-sharma:
Changes the default for validating secrets to `true`. None of the ng jobs nor the search service override the default value.

Part of https://github.com/NuGet/Engineering/issues/1379
See https://devdiv.visualstudio.com/DevDiv/_workitems/edit/594980


joelverhagen:
Validating a certificate means... it must chain to a trusted root? Is this the case today? Are these KeyVault certificates/SSL certificates/what else?

loic-sharma:
This validates certificates that are loaded from VMs'/services' certificate store. This change only affects the secrets certificate that is used to authenticate to Key Vault. This flag is passed to the [`X509Certificate2Collection.Find` method](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificate2collection.find?view=netframework-4.7.1) which will ensure that an x509 certificate chain can be built (see [here](https://github.com/dotnet/corefx/blob/ab6b810680742a4d021792ff1f8a3cc9ee882f3d/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Windows/FindPal.cs#L358)).

PR #286, opened 2018-04-18
Polish V3PerPackage now that it's tested against all packages

joelverhagen:
1. Don't process deleted packages
1. Allow messages to be batches if they have 10 or fewer attempts. There is no automatic DLQ on Azure Storage Queues. This allows us to quickly get through messages that failed due to transient issues.

I have run the tool over all of PROD and these are two changes I made locally to get through everything quickly.

### DIFF
--- a/src/V3PerPackage/PerBatchProcessor.cs
+++ b/src/V3PerPackage/PerBatchProcessor.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Net;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


src/V3PerPackage/PerBatchProcessor.cs
```diff
@@ -56,6 +67,39 @@ public class PerBatchProcessor
...
+
+            return tasks
+                .Where(x => x.Task.Result)
+                .Select(x => x.Context)
+                .ToList();
+        }
+
+        private async Task<bool> DoesPackageExistAsync(PerPackageContext packageContext)
+        {
+            using (var request = new HttpRequestMessage(HttpMethod.Head, packageContext.PackageUri))
```
scottbommarito: Note that this skips packages that exist in V2 but for whatever reason are missing from the packages container. I think it might be useful to add that in a comment.

joelverhagen: Sure, sounds good.

joelverhagen: Done.

PR #288, opened 2018-04-18
[Search] Add config for Azure Directory's cache path

loic-sharma:
By default, Lucene will store the Azure Directory cache inside of the temp path. On Azure, this is the same drive as the operating system, which only has 20GB of free space. This configuration will allow us to move the Azure Directory cache to a drive with more free space. If `AzureDirectoryCachePath` configuration isn't set, the search service will default to using the temp path.


src/NuGet.Indexing/IndexDirectoryProvider/CloudIndexDirectoryProvider.cs
```diff
@@ -75,7 +79,24 @@ public bool Reload(IndexingConfiguration config)
...
+            {
+                sourceDirectory = new AzureDirectory(storageAccount, _indexContainerName);
+            }
+            else
+            {
+                _logger.LogInformation("Using custom azure directory cache path: {AzureDirectoryCachePath}", config.AzureDirectoryCachePath);
+
+                var azureDirectoryCachePath = Path.Combine(config.AzureDirectoryCachePath, _indexContainerName);
+
+                IODirectory.CreateDirectory(azureDirectoryCachePath);
```
agr: What will happen if path exists?

src/NuGet.Indexing/IndexDirectoryProvider/CloudIndexDirectoryProvider.cs
```diff
@@ -75,7 +79,24 @@ public bool Reload(IndexingConfiguration config)
...
-            var sourceDirectory = new AzureDirectory(storageAccount, _indexContainerName);
+            AzureDirectory sourceDirectory;
+
+            if (string.IsNullOrEmpty(config.AzureDirectoryCachePath))
+            {
+                sourceDirectory = new AzureDirectory(storageAccount, _indexContainerName);
+            }
+            else
+            {
+                _logger.LogInformation("Using custom azure directory cache path: {AzureDirectoryCachePath}", config.AzureDirectoryCachePath);
```
joelverhagen: I would either also log the absolute path with `Path.GetFullPath`, pass the full path to `AzureDirectory` ctor, or require `Path.IsPathRooted`.

As it stands today, a configured relative path has unclear behavior.

src/NuGet.Indexing/IndexDirectoryProvider/CloudIndexDirectoryProvider.cs
agr: Should we use `DirectoryInfo.Create` instead?

joelverhagen: `CreateDirectory` no-ops if the directory already exists.

loic-sharma: Nothing will happen :)

Source: https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.createdirectory?view=netframework-4.7.1#System_IO_Directory_CreateDirectory_System_String_

src/NuGet.Indexing/IndexDirectoryProvider/CloudIndexDirectoryProvider.cs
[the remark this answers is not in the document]
loic-sharma: Fixed

agr: FYI, `IsPathRooted` does not guarantee path is absolute. Try
    
     [System.IO.Path]::IsPathRooted('c:a.txt');

in powershell.

joelverhagen: Oh good point. In this case, I would just avoid `IsPathRooted`...

loic-sharma: I went ahead and did `Path.GetFullPath`

src/NuGet.Services.BasicSearch/SafeRoleEnvironment.cs
```diff
@@ -62,6 +62,11 @@ public static string GetDeploymentId()
             return string.Empty;
         }
 
+        public static string GetLocalResourceRootPath(string name)
+        {
+            return RoleEnvironment.GetLocalResource(name).RootPath;
```
loic-sharma: This throws if a local resource with that `name` doesn't exist.

src/NuGet.Services.BasicSearch/Startup.cs
```diff
@@ -91,6 +91,26 @@ public class Startup
...
+            {
+                if (!SafeRoleEnvironment.IsAvailable)
+                {
+                    _logger.LogWarning(
+                        "Cannot use Azure Local Resource {LocalResourceName} for caching when the RoleEnvironment is not available",
+                        config.AzureDirectoryCacheLocalResourceName);
+                }
+                else
+                {
+                    config.AzureDirectoryCachePath = SafeRoleEnvironment.GetLocalResourceRootPath(config.AzureDirectoryCacheLocalResourceName);
```
joelverhagen: Let's document on `AzureDirectoryCachePath` that it gets overwritten at runtime if `AzureDirectoryCacheLocalResourceName` is set.

src/NuGet.Services.BasicSearch/SafeRoleEnvironment.cs
```diff
@@ -62,6 +62,11 @@ public static string GetDeploymentId()
             return string.Empty;
         }
 
+        public static string GetLocalResourceRootPath(string name)
```
joelverhagen: We need to test what this does when:

1. Size is "small", e.g. 50 GB.
1. Size is large enough to be less than max disk size but too big given already consumed disk space, e.g. 199 GB.
1. Size is exactly what our VM SKU advertises, e.g. 200 GB.
1. Size is great than VM disk size, e.g. 300 GB.

joelverhagen: Also, what does `cleanOnRoleRecycle` do for us? What will we set it to on our resource?

loic-sharma:
See Local Storage: https://github.com/Huachao/azure-content/blob/master/articles/cloud-services/cloud-services-configure-local-storage-resources.md

src/NuGet.Services.BasicSearch/Startup.cs
loic-sharma: Fixed

PR #289, opened 2018-04-18
Use new DequeueCount property from NuGet.Services.Storage

joelverhagen:
Pull in https://github.com/NuGet/ServerCommon/pull/148.

This just makes the code less nasty.


joelverhagen:
This PR only affects the V3PerPackage tool, which is not deployed. It is run from your devbox if you need it.

PR #294, opened 2018-05-04
Telemetry for V3 - DO NOT MERGE

skofman1:


src/Ng/Telemetry/JobPropertiesTelemetryInitializer.cs
```diff
@@ -0,0 +1,31 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.ApplicationInsights.Channel;
+using Microsoft.ApplicationInsights.Extensibility;
+using NuGet.Services.Metadata.Catalog;
+using System;
+
+namespace Ng
+{
+    public class JobPropertiesTelemetryInitializer : ITelemetryInitializer
```
skofman1: The idea here is that a job can define at runtime which properties it wants to add to all telemetry.

src/Catalog/Telemetry/TelemetryService.cs
```diff
@@ -91,5 +64,10 @@ public void TrackCatalogIndexWriteDuration(TimeSpan duration, Uri uri)
                     { Uri, uri.AbsoluteUri },
                 });
         }
+        
+        public virtual DurationMetric TrackDuration(string name, IDictionary<string, string> properties = null)
```
skofman1: DurationMetric is a class Loic added in NuGet.Services.Logging. Very useful. Given more time I would add the ability to add/set properties at runtime. Right now the list of properties is given in ctor, and in order to modify them u need to maintain a reference to the dictionary.

src/Catalog/Telemetry/TelemetryService.cs
```diff
@@ -20,44 +23,14 @@ public class TelemetryService : ITelemetryService
...
-                    { Method, method.ToString() },
-                    { Uri, uri.AbsoluteUri },
-                    { StatusCode, ((int?)statusCode)?.ToString() },
-                    { Success, success.ToString() },
-                    { ContentLength, contentLength?.ToString() }
-                });
+            GlobalDimensions = new Dictionary<string, string>();
         }
 
         public void TrackCatalogIndexReadDuration(TimeSpan duration, Uri uri)
```
skofman1: Given more time I would replace the usage of TrackCatalogIndexReadDuration, TrackCatalogIndexWriteDuration with TrackDuration so we won't have to initilize Stopwatch in the code.

src/Ng/Jobs/Feed2CatalogJob.cs
```diff
@@ -75,12 +75,22 @@ protected override void Init(IDictionary<string, string> arguments, Cancellation
             CatalogStorage = catalogStorageFactory.Create();
             AuditingStorage = auditingStorageFactory.Create();
 
+            TelemetryService.GlobalDimensions[TelemetryConstants.Destination] = catalogStorageFactory.BaseAddress.AbsoluteUri;
```
skofman1: this gives the "region" on which the job operates

skofman1:
>             bool run;

This is the main execution loop. I think it will be useful to add TelemetryService.GlobalDimensions[TelemetryConstants.CorrelationId] = Guid.NewGuid() inside the loop to correlate packages processed in a single iteration (might be useful in other places as well)

---
Refers to: src/Ng/Jobs/Catalog2RegistrationJob.cs:120 in bf8300b. [](commit_id = bf8300b31f0d2ae04d963fad9441a76a4a9a739a, deletion_comment = False)

src/Catalog/Helpers/FeedHelpers.cs
```diff
@@ -261,10 +262,22 @@ private static DateTime ForceUtc(DateTime date)
                 return lastDate;
             }
 
+            var packageProcessingDurationTelemetryList = new List<DurationMetric>();
```
skofman1: Here I measure the processing time of a package from being read from the feed to written to catalog. 
It's a bit tricky since they are read one by one, by commited in batch, hence the list of DuartionMetric that need to be disposed together. If we see this pattern in other places in the code I suggest having a class to manage the list creation and dispose... :)

src/Catalog/Registration/RegistrationCollector.cs
```diff
@@ -66,6 +66,8 @@ protected override Task<IEnumerable<CatalogItemBatch>> CreateBatches(IEnumerable
             // from the correct location (even though we may have
             // a little rework).
 
+
+            // TODO: this is where processing of a package starts...
```
skofman1: If we want to measure the processing duration of a single package in Catalog2Registration, I recommend this point to start the measuring

src/Catalog/Registration/RegistrationCollector.cs
```diff
@@ -88,6 +90,8 @@ private string GetKey(JToken item)
         {
             var tasks = new List<Task>();
 
+            // TODO: this is where processing of a package ends.
```
skofman1: This is where the Registration blobs are saved. Perhaps use a Task.ContinueWith to complete the measurement?

dtivel:
Abandoning.  Implemented through https://github.com/NuGet/NuGet.Services.Metadata/pull/301.

PR #295, opened 2018-05-16
TSA: Update MS.AI.JS to version using strict mode

chenriksson:
Addresses TSA #608742, 608743


chenriksson:
Note that this did enable strict mode, but moderncop still isn't happy. This is the best we can do for now, and I've opened the following issue against AI.js:
https://github.com/Microsoft/ApplicationInsights-JS/issues/604

`
At (1,1) "Code should run in strict mode wherever possible. Strict mode was set for this file but was enabled at global scope (which can lead to unexpected behaviors if this file is concatenated as part of a post-processing step). Consider using the JavaScript module pattern to wrap all code contained in the file within a function expression. Enable strict mode as the first statement of the containing function. To disable enforcement of the module pattern and permit setting strict mode at global file level, pass /setting:JS2085.EnableStrictMode.EnforceModulePattern=false on the command-line."
`

PR #299, opened 2018-05-22
For DB2Lucene, check that package status key is 0, not that the package is not deleted

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/5912

Use `PackageStatusKey`, which is aware of the validating state, and not `Deleted`, which is `0` regardless of whether or not the package is validating or not.


src/NuGet.Indexing/Sql2Lucene.cs
```diff
@@ -89,7 +89,7 @@ FROM Packages
                     INNER JOIN PackageRegistrations ON Packages.PackageRegistrationKey = PackageRegistrations.[Key]
                       AND Packages.[Key] >= @BeginKey
                       AND Packages.[Key] < @EndKey
-                    WHERE Packages.Deleted = 0
+                    WHERE Packages.PackageStatusKey = 0
```
agr: Add a comment somewhere please that `0` is `Available`.

scottbommarito:
Tested this on DEV and it works as expected.
Merging now.

PR #303, opened 2018-05-24
V3:  increase default HTTP connection limit

dtivel:
Resolve https://github.com/NuGet/NuGetGallery/issues/5996.

Catalog2Registration implements a single "save" operation as an [async blob upload plus an async blob snapshot](https://github.com/NuGet/NuGet.Services.Metadata/blob/master/src/Catalog/Persistence/AzureStorage.cs#L120).  Mimicking this in a simple console app, I "saved" 10,000 small (~3.7) blobs from stateside to China blob storage.  End-to-end times improved dramatically as I increased .NET's [`DefaultConnectionLimit`](https://msdn.microsoft.com/en-us/library/system.net.servicepointmanager.defaultconnectionlimit(v=vs.110).aspx) value.

image

This change updates all Ng jobs' `DefaultConnectionLimit` to 64.  The previous limit for Feed2Catalog was 2.  The previous limit for Catalog2Dnx, Catalog2Registration, and Catalog2Lucene was 4.  The previous limit for LightningJob was 32.


dtivel:
@scottbommarito, ready for re-review.

src/Ng/Jobs/Feed2CatalogJob.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
```
chenriksson: necessary?

PR #306, opened 2018-05-30
V3 monitoring: don't requeue any packages if the queue is too big

scottbommarito:
If there are a lot of packages in the monitoring queue, and some failures (especially intermittent ones) have occurred, the queue would spiral out of control because `Monitoring2Monitoring` will requeue the failed packages over and over before `MonitoringProcessor` can process them. This change stops `Monitoring2Monitoring` from requeuing packages if the queue has too many messages, so the queue can no longer spiral out of control.


src/Catalog/Helpers/ParallelAsync.cs
```diff
@@ -3,25 +3,24 @@
...
-        /// <param name="degreeOfParallelism">The number of tasks to create.</param>
+        /// <param name="degreeOfParallelism">The number of tasks to create. Defaults to <see cref="ServicePointManager.DefaultConnectionLimit"/></param>
         /// <returns>A task that completes when all tasks have completed.</returns>
-        public static Task Repeat(Func<Task> taskFactory, int degreeOfParallelism = DefaultDegreeOfParallelism)
+        public static Task Repeat(Func<Task> taskFactory, int? degreeOfParallelism = null)
         {
             return Task.WhenAll(
                 Enumerable
-                    .Repeat(taskFactory, degreeOfParallelism)
+                    .Repeat(taskFactory, degreeOfParallelism ?? ServicePointManager.DefaultConnectionLimit)
```
chenriksson: Should we be using `ServicePointManager.DefaultConnectionLimit`? This setting is a limit on TCP connections, but it looks like you're using it for limiting the number of storage queue messages which is completely different.

Based on your description, is the real issue that the same failed packages keep getting requeued? Should we instead introduce a cache to avoid requeing the same packages?

scottbommarito: This was an additional parallelism improvement I added, unrelated to the primary purpose of this PR. My thoughts are that the number of messages being processed simultaneously should be allowed to be as great as the number of web connections we can make at the same time. I can revert this change if you want, I didn't think it was very risky/notable.

dtivel:
Don't you need to also update https://github.com/NuGet/NuGet.Services.Metadata/blob/dev/src/Ng/Scripts/Monitoring2Monitoring.cmd#L10?

src/Ng/Jobs/Monitoring2MonitoringJob.cs
```diff
@@ -42,6 +46,12 @@ protected override void Init(IDictionary<string, string> arguments, Cancellation
 
         protected override async Task RunInternal(CancellationToken cancellationToken)
         {
+            var currentMessageCount = await _queue.GetMessageCount(cancellationToken);
+            if (currentMessageCount > _maxRequeueQueueSize)
+            {
+                Logger.LogInformation("Can't requeue any invalid packages because the queue has too many messages!");
```
dtivel: Shouldn't this return early?

scottbommarito: lol yup, fixed

scottbommarito:
@dtivel We can update `Monitoring2Monitoring.cmd` when we decide we want to configure the value to be anything but the default. Right now I don't see the need to do that.

src/Ng/Arguments.cs
```diff
@@ -133,6 +133,11 @@ public static class Arguments
         /// Default value of <see cref="PackageStatusFolder"/>.
         /// </summary>
         public const string PackageStatusFolderDefault = "status";
+
+        /// <summary>
+        /// If the queue contains more messages than this, the job will not requeue any invalid packages.
+        /// </summary>
+        public const string MaxRequeueQueueSize = "maxQueueSize";
```
dtivel: `MaxQueueSize`, to be consistent with the parameter name?

scottbommarito: The pattern in this file is that the arguments start with lowercase. For example, `StorageAccountName` maps to `storageAccountName`.

src/Ng/Jobs/Monitoring2MonitoringJob.cs
shishirx34: may be log this as warning.. so that we can easily look it up in the logs?

scottbommarito: I don't think it counts as a warning-- the fact that the queue has a lot of packages doesn't signify anything is wrong with the system. We can also already find the message very easily in the logs by looking at the message directly.

src/Ng/Jobs/Monitoring2MonitoringJob.cs
```diff
@@ -42,6 +46,13 @@ protected override void Init(IDictionary<string, string> arguments, Cancellation
...
```
loic-sharma: >LogInformation [](start = 23, length = 14)

Also, it could be useful to log the value of `currentMessageCount`

scottbommarito: I'll log both the current message count and the max size, good idea!

src/Ng/Jobs/Monitoring2MonitoringJob.cs
loic-sharma: It *may* indicate something wrong, right? If so, I would make it a warning.

scottbommarito:
@shishirx34 I think it's still worth knowing if the queue size is large, especially if it's not decreasing, because that suggests that `MonitoringProcessor` might not be running. We shouldn't be above 100 messages at any time but it's certainly possible that someone might push/unlist a huge number of packages at once and this will be the case.

src/Ng/Jobs/Monitoring2MonitoringJob.cs
shishirx34: Filtering is far easier on warnings than on information and like @loic-sharma said it could possibly be an issue(like it happened this week)

src/Ng/Jobs/Monitoring2MonitoringJob.cs
shishirx34: also, would it be easier to log the count of `invalid` blobs somewhere in AI when not requeueing? don't need to go to the blob storage to look it up?

PR #309, opened 2018-06-01
[ReleasePrep][06.01.2018]FI of master into dev

shishirx34:


shishirx34:
Not needed I guess, will do an RI

shishirx34:
Can't do the FI with the strict check setting enabled.

PR #311, opened 2018-06-01
[ReleasePrep][06.01.2018]FI of master into dev

shishirx34:


shishirx34:
VSTS: NuGet.Services.Metadata - Queue NuGetDeployment CI Expected — Waiting for status to be reported - Required

This status check is preventing an FI. I will do an RI directly.

PR #313, opened 2018-06-05
V3:  add telemetry for processing a catalog batch

dtivel:
Resolve https://github.com/NuGet/NuGetGallery/issues/6032.


src/Catalog/CommitCollector.cs
```diff
@@ -71,13 +71,16 @@ public abstract class CommitCollector : CollectorBase
...
                     }
 
-                    acceptNextBatch = await OnProcessBatch(
-                        client,
-                        batch.Items.Select(item => item.Value),
-                        context,
-                        batch.CommitTimeStamp,
-                        batch.CommitTimeStamp == lastBatch.CommitTimeStamp,
-                        cancellationToken);
+                    using (_telemetryService.TrackDuration(TelemetryConstants.ProcessBatchSeconds))
```
shishirx34: do you think it might be useful to track the number of items in `batch` along with the duration as the metric?(As to average out the sampling per package)

dtivel: It could.  Custom metrics aren't a great fit though as additional numbers need to be stored as a custom dimension (string).

PR #319, opened 2018-06-26
Catalog2Dnx:  prepare for parallelization

dtivel:
Progress on https://github.com/NuGet/NuGetGallery/issues/6101.

* Pull out package version index update from .nupkg/.nuspec add/delete operations.  This is to enable parallelized .nupkg/.nuspec add/delete operations followed by parallelized package version index updates (after grouping by package ID).
* Add and update tests.


src/Catalog/Dnx/DnxMaker.cs
```diff
@@ -17,78 +18,57 @@ public class DnxMaker
...
             var relativeAddress = GetRelativeAddressNuspec(id, version);
             var nuspecUri = new Uri(storage.BaseAddress, relativeAddress);
+
             await storage.Save(nuspecUri, new StringStorageContent(nuspec, "text/xml", "max-age=120"), cancellationToken);
+
             return nuspecUri;
         }
 
-        private async Task UpdateMetadata(Storage storage, Action<HashSet<NuGetVersion>> updateAction, CancellationToken cancellationToken)
+        public async Task UpdatePackageVersionIndexAsync(string id, Action<HashSet<NuGetVersion>> updateAction, CancellationToken cancellationToken)
```
agr: Add argument validation since it was made public?

src/Catalog/Dnx/DnxCatalogCollector.cs
```diff
@@ -135,7 +141,12 @@ public class DnxCatalogCollector : CommitCollector
 
         private async Task ProcessPackageDeleteAsync(string id, string version, CancellationToken cancellationToken)
         {
-            await _dnxMaker.DeletePackage(id, version, cancellationToken);
+            await _dnxMaker.DeletePackageAsync(id, version, cancellationToken);
+
+            await _dnxMaker.UpdatePackageVersionIndexAsync(
```
cristinamanum: The order of operations changed from the previous implementation(previously it was Update metadata followed by the  nuspec, nupkg deletes now the update metadata is the last). This could lead to cases when the nupkg is deleted and the metadata is not.

dtivel: Good catch.

src/Catalog/Dnx/DnxMaker.cs
```diff
@@ -101,7 +144,8 @@ public DnxMaker(StorageFactory storageFactory)
             {
                 // Store versions (sorted)
                 result.Sort();
-                await storage.Save(resourceUri, CreateContent(result.Select((v) => v.ToString())), cancellationToken);
+
+                await storage.Save(resourceUri, CreateContent(result.Select((v) => v.ToNormalizedString())), cancellationToken);
```
loic-sharma: >(v) [](start = 76, length = 3)

`(v) => ...` can be `v => ...`

src/Catalog/Dnx/DnxMaker.cs
```diff
@@ -177,12 +221,24 @@ private StorageContent CreateContent(IEnumerable<string> versions)
...
+            {
+                throw new ArgumentException(Strings.ArgumentMustNotBeNullOrEmpty, nameof(id));
+            }
+
+            if (string.IsNullOrEmpty(version))
+            {
+                throw new ArgumentException(Strings.ArgumentMustNotBeNullOrEmpty, nameof(version));
+            }
+
+            var normalizedVersion = NuGetVersion.Parse(version).ToNormalizedString();
```
loic-sharma: >NuGetVersion.Parse(version).ToNormalizedString() [](start = 36, length = 48)

You could use `NuGetVersionUtility.NormalizeVersion` here

dtivel: That would introduce undesirable leniency in this API.  Currently, if you call this method with an unparseable version:

    DnxMaker.GetRelativeAddressNupkg(id: "a", version: "b")

...it will throw with `System.ArgumentException: 'b' is not a valid version string.`  I think that is correct behavior.  If `GetRelativeAddressNupkg(...)` instead used `NuGetVersionUtility.NormalizeVersion(...)`, the above method call would not throw and would return:

    b/a.b.nupkg

The previous implementation threw on invalid version and I'm preserving that behavior.

PR #320, opened 2018-06-28
Catalog2Dnx:  parallelize operations within a catalog batch

dtivel:
Resolve https://github.com/NuGet/NuGetGallery/issues/6101.

This change parallelizes flat container operations within a single catalog batch.

1.  .nupkg and .nuspec uploads are parallelized.
2.  Package version index changes are parallelized on package ID.  If a catalog batch contains multiple entries for the same package ID but different versions, Catalog2Dnx will batch index updates for the same package ID.  Prior to this change, if a catalog batch contained _n_ entries for the same package ID, Catalog2Dnx would update the version index _n_ times.  With this change, Catalog2Dnx will update the same version index only once.

I tested this in the DEV environment, and in terms of catalog entries per hour:

| Job Instance | Before | After |
| --- | --- | --- |
| catalog2dnx (U.S.) | 4,715 | 12,882 |
| catalog2dnx (China) | 896 | 1,890 |


src/Catalog/Dnx/DnxCatalogCollector.cs
```diff
@@ -21,122 +23,182 @@ public class DnxCatalogCollector : CommitCollector
...
             ILogger logger,
+            int maxDegreeOfParallelism,
             Func<HttpMessageHandler> handlerFunc = null,
             TimeSpan? httpClientTimeout = null)
             : base(index, telemetryService, handlerFunc, httpClientTimeout)
         {
             _storageFactory = storageFactory;
             _dnxMaker = new DnxMaker(storageFactory);
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _maxDegreeOfParallelism = maxDegreeOfParallelism;
```
loic-sharma: It may be useful to validate `maxDegreeOfParallelism` here and throw an `ArgumentOutOfRangeException` as early as possible

src/Catalog/Dnx/DnxCatalogCollector.cs
```diff
@@ -21,122 +23,182 @@ public class DnxCatalogCollector : CommitCollector
...
-        protected override async Task<bool> OnProcessBatch(CollectorHttpClient client, IEnumerable<JToken> items, JToken context, DateTime commitTimeStamp, bool isLastBatch, CancellationToken cancellationToken)
+        protected override async Task<bool> OnProcessBatch(
+            CollectorHttpClient client,
+            IEnumerable<JToken> items,
+            JToken context,
+            DateTime commitTimeStamp,
+            bool isLastBatch,
+            CancellationToken cancellationToken)
+        {
+            var catalogEntries = items.Select(item =>
```
loic-sharma: This LINQ query is going to be run once for `AssertNoMultipleEntriesForSamePackageIdentity` and once for `ProcessCatalogEntriesAsync`. Should it be executed once with `.ToList()` instead?

PR #322, opened 2018-07-11
Feed2Catalog:  permit package retrieval via cloud blob storage instead of V2

dtivel:
Resolve https://github.com/NuGet/NuGetGallery/issues/6127.

Accessing packages via Azure cloud blob storage provides a few advantages:

1.  The package hash is directly available as blob metadata.
2.  Azure storage API's provide a seekable stream (vs. `System.Net.Http.HttpClient`'s non-seekable stream).  `System.IO.Compression.ZipArchive` can use this seekable network stream to read select package metadata.

These changes obviate the need for Feed2Catalog to download the package.  This results in a huge time savings on larger packages.

On a test machine I observed an ~8% improvement in processing throughput given other json-ld fixes and packages following production distributions within 1% with respect to package size and entry counts.


src/Catalog/PackageCatalogItemCreator.cs
```diff
@@ -0,0 +1,234 @@
...
+            else
+            {
+                await blob.FetchAttributesAsync(cancellationToken);
+
+                string packageHash = null;
+                var etag = blob.ETag;
+
+                var metadata = await blob.GetMetadataAsync(cancellationToken);
+
+                if (metadata.TryGetValue("SHA512", out packageHash))
```
scottbommarito: nit: put `"SHA512"` in a constant

src/Catalog/PackageCatalogItemCreator.cs
```diff
@@ -0,0 +1,234 @@
...
+
+                        if (item == null)
+                        {
+                            _logger.LogWarning("Unable to extract metadata from: {PackageDetailsContentUri}", packageItem.ContentUri);
+                        }
+                    }
+
+                    if (item != null)
+                    {
+                        await blob.FetchAttributesAsync(cancellationToken);
```
scottbommarito: Why are we fetching the blob again just for the purpose of checking its `ETag`?

Do we have any reason to believe that the `.nupkg` associated with a package will change as Feed2Catalog is processing it and the change to that `.nupkg` will not be reflected in a future commit to the catalog?

I would believe that if a package's `.nupkg` is modified while Feed2Catalog is processing it, one of these two scenarios has happened:
1. the package was modified by a revalidation (e.g. repo signing) that finished just as Feed2Catalog started processing the package
2. the package was deleted right when Feed2Catalog started processing it

Either cases will result in an additional catalog entry (because the packages container is modified BEFORE the feed is changed) so we don't need to check the blob again.

scottbommarito: As a side note (not explicitly related to this PR), what happens if:
1. A package is pushed.
2. Every V3 job except Feed2Catalog stops running.
3. Feed2Catalog commits this package to the catalog.
4. The package is deleted.
5. Feed2Catalog commits this package's delete to the catalog.
6. The V3 jobs start running again and process the package's creation and deletion commits.

Have we verified that all jobs that download the package from the packages container skip a commit if the package associated with that commit doesn't exist in the packages container?

src/Catalog/PackageCatalogItemCreator.cs
scottbommarito: Is there any way that we can verify that the package we're downloading is the same package that we just read from the feed?

Is there a package hash in the feed that we can compare this to?

src/Catalog/PackageCatalogItemCreator.cs
```diff
@@ -0,0 +1,234 @@
...
+
+        public static PackageCatalogItemCreator Create(
+            HttpClient httpClient,
+            ITelemetryService telemetryService,
+            ILogger logger,
+            IStorage storage)
+        {
+            if (httpClient == null)
+            {
+                throw new ArgumentNullException(nameof(httpClient));
```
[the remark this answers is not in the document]
dtivel: Please reread [this discussion](https://github.com/NuGet/NuGetGallery/pull/5833#discussion_r184444532).

src/Catalog/PackageCatalogItemCreator.cs
dtivel: 1.  We are _not_ fetching the blob again just for the purpose of checking its ETag.  We are fetching the blob's _properties and metadata_ again for the purpose of checking its ETag.  There's a difference, which becomes more obvious with large blobs.  See [`CloudBlob.FetchAttributesAsync(...)`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.windowsazure.storage.blob.cloudblob.fetchattributesasync?view=azure-dotnet#Microsoft_WindowsAzure_Storage_Blob_CloudBlob_FetchAttributesAsync_System_Threading_CancellationToken_) for details.

2.  The original code fetched the package, wrote it locally, and performed multiple actions on the local file stream.  Implicitly, this provided a guarantee that the input for multiple subsequent actions was constant.  It's theoretically possible (though rare, I think) that in your first scenario, Feed2Catalog has read the package hash from blob v1.0 and a revalidation updates the blob while Feed2Catalog is reading it.  Now the package hash doesn't match the package, and the subsequent catalog entry will be incorrect.  This is more likely with large packages.  The catalog will be updated with correct information in time, but why break an existing consistency guarantee?

dtivel: That question is unrelated to this PR.  :)

src/Catalog/PackageCatalogItemCreator.cs
dtivel: That's a good idea, and yes we can read the package hash from the feed and compare it to the package hash stashed in the blob's metadata.  That would provide even stronger consistency guarantees for package reads.  See [this discussion](https://github.com/NuGet/NuGet.Services.Metadata/pull/322#discussion_r201877706).

src/Catalog/PackageCatalogItemCreator.cs
```diff
@@ -0,0 +1,234 @@
...
+                item = await GetPackageViaHttpAsync(packageItem, timestamp, item, cancellationToken);
+            }
+
+            return item;
+        }
+
+        private async Task<PackageCatalogItem> GetPackageViaStorageAsync(FeedPackageDetails packageItem, CancellationToken cancellationToken)
+        {
+            PackageCatalogItem item = null;
+            var packageName = $"{packageItem.PackageId.ToLowerInvariant()}.{packageItem.PackageVersion.ToLowerInvariant()}.nupkg";
```
skofman1: >{packageItem.PackageId.ToLowerInvariant()}.{packageItem.PackageVersion.ToLowerInvariant()}.nupkg [](start = 32, length = 96)

Not sure, but I suspect that generating package name from id and version is duplicated in a few places in this code base.. consider putting in a shared location.

src/Catalog/PackageCatalogItemCreator.cs
skofman1: when will this situation happen? Why should we be concerned that the blob is modified?

src/Catalog/PackageCatalogItemCreator.cs
```diff
@@ -0,0 +1,234 @@
...
+using NuGet.Services.Metadata.Catalog.Helpers;
+using NuGet.Services.Metadata.Catalog.Persistence;
+
+namespace NuGet.Services.Metadata.Catalog
+{
+    public sealed class PackageCatalogItemCreator
+    {
+        private readonly HttpClient _httpClient;
+        private readonly ILogger _logger;
+        private readonly IAzureStorage _storage;
```
skofman1: this is the first time in the V3 jobs we introduce a strong dependency on Azure. Up until now they could have functioned the same way on file system as well. Can we avoid this? Can we hide all the Azure specific functionality behind interfaces?

src/Catalog/Helpers/Utils.cs
```diff
@@ -476,36 +476,47 @@ public static IEnumerable<PackageEntry> GetEntries(ZipArchive package)
...
+
+            packageHash = packageHash ?? GenerateHash(stream);
 
             stream.Seek(0, SeekOrigin.Begin);
 
-            using (ZipArchive package = new ZipArchive(stream, ZipArchiveMode.Read, true))
+            using (var package = new ZipArchive(stream, ZipArchiveMode.Read, leaveOpen: true))
             {
-                nupkgMetadata.Nuspec = GetNuspec(package);
+                var nuspec = GetNuspec(package);
```
skofman1: we can do the same for Catalog2Dnx :)

src/Catalog/PackageCatalogItemCreator.cs
dtivel: See [this discussion](https://github.com/NuGet/NuGet.Services.Metadata/pull/322#discussion_r201877706).

src/Catalog/PackageCatalogItemCreator.cs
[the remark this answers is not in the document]
scottbommarito: In that discussion, you state that a static factory makes sense when "the initialization involves a lot of code other than simple argument validation and field/property initialization".

I would consider throwing if the arguments are null to be "simple argument validation".

src/Catalog/PackageCatalogItemCreator.cs
dtivel: `IAzureStorage` is an interface, and there's nothing in the interface that is Azure-specific except the presence of `Azure` in the name.  I could rename it to `IBlobStorage` if you prefer.

I could add file system support here, but I don't want to.  [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).  The goal of this PR is to improve V3 throughput for repository signing, and we don't need file system support here.  This new code path is opt-in only.  If there is a strong case to add file system support, we can add it to the backlog.

src/Catalog/Helpers/Utils.cs
dtivel: Do what exactly?  Sorry, I'm not following.

src/Catalog/PackageCatalogItemCreator.cs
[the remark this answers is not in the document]
scottbommarito: Forgot `HttpClient` was disposable, woops.

src/Catalog/PackageCatalogItemCreator.cs
scottbommarito: Ok, so to clarify:

1. when you are reading a blob's stream, there's no guarantee that the contents of the blob are not modified while you are reading it
2. by fetching the blob's metadata again after the operation has completed, we guarantee that the stream we read data from was not modified while we were reading it

This makes sense, but you should definitely add a comment in the code to explain why we are fetching the metadata again.

loic-sharma:
>             using (HashAlgorithm hashAlgorithm = HashAlgorithm.Create("SHA512"))

Should this be using your new Constant.Sha512?

---
Refers to: src/Catalog/Helpers/Utils.cs:446 in ac135ea. [](commit_id = ac135ead6785920ae74fcff6c7ad95de55d757f2, deletion_comment = False)

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -291,5 +291,26 @@ public override Uri GetUri(string name)
...
             return !(await source.ExistsAsync());
         }
+
+        public ICloudBlockBlob GetCloudBlockBlobReference(string name)
+        {
+            Uri uri = ResolveUri(name);
+            string blobName = GetName(uri);
+            CloudBlockBlob blob = _directory.GetBlockBlobReference(blobName);
+
+            if (blob.Exists())
```
loic-sharma: >Exists [](start = 21, length = 6)

There's an async version of `Exists`, can you use that?

loic-sharma:
>                 Trace.TraceError("Exception: {0} {1} {2}", origin, e.GetType().Name, e.Message);

Could you change this to log the full exception instead of just the type/message?

---
Refers to: src/Catalog/Helpers/Utils.cs:524 in ac135ea. [](commit_id = ac135ead6785920ae74fcff6c7ad95de55d757f2, deletion_comment = False)

src/Catalog/PackageCatalogItemCreator.cs
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        private async Task<PackageCatalogItem> GetPackageViaStorageAsync(FeedPackageDetails packageItem, CancellationToken cancellationToken)
+        {
+            PackageCatalogItem item = null;
+            var packageFileName = PackageUtility.GetPackageFileNameLowercase(packageItem.PackageId, packageItem.PackageVersion);
+            var blob = _storage.GetCloudBlockBlobReference(packageFileName);
+
+            if (blob == null)
+            {
```
loic-sharma: Minor suggestion: I think this would read slightly better if this condition returned null. Then, the rest of this method wouldn't need to be nested in an `else` block.

src/Catalog/PackageCatalogItemCreator.cs
```diff
@@ -0,0 +1,240 @@
...
+                        packageItem.ContentUri.ToString(),
+                        stream,
+                        packageItem.CreatedDate,
+                        packageItem.LastEditedDate,
+                        packageItem.PublishedDate);
+
+                    if (item == null)
+                    {
+                        _logger.LogWarning("Unable to extract metadata from: {PackageDetailsContentUri}", packageItem.ContentUri);
+                    }
```
loic-sharma: Is this the right behavior? The package will be in v2 but not v3.

src/Catalog/PackageCatalogItemCreator.cs
```diff
@@ -0,0 +1,240 @@
...
+        private async Task<PackageCatalogItem> GetPackageViaStorageAsync(FeedPackageDetails packageItem, CancellationToken cancellationToken)
+        {
+            PackageCatalogItem item = null;
+            var packageFileName = PackageUtility.GetPackageFileNameLowercase(packageItem.PackageId, packageItem.PackageVersion);
+            var blob = _storage.GetCloudBlockBlobReference(packageFileName);
+
+            if (blob == null)
+            {
+                _telemetryService.TrackMetric(
+                    TelemetryConstants.NonExistentBlob,
```
loic-sharma: The pattern we've taken to in NuGet.Jobs is making high-level methods for our metrics. Things like `_telemetryService.TrackNonExistenBlob`

src/Catalog/Helpers/PackageUtility.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace NuGet.Services.Metadata.Catalog.Helpers
+{
+    public static class PackageUtility
+    {
+        public static string GetPackageFileNameLowercase(string packageId, string packageVersion)
```
cristinamanum: Is this package version or normalized version? If normalized it would be great to have this in the argument name.

src/Catalog/Persistence/AzureStorage.cs
dtivel: Good catch.

src/Catalog/PackageCatalogItemCreator.cs
dtivel: It is a non-goal as part of my V3 throughput work to make behavioral changes, especially not in this PR.  Please open a new issue if you think this should change.

src/Catalog/PackageCatalogItemCreator.cs
dtivel: Nice to have.

src/Catalog/Helpers/PackageUtility.cs
dtivel: It's indeterminate.  The caller is responsible for normalizing.

Adding an option for normalizing would be a nice to have refactoring, I'll punt it from this PR.

src/Catalog/PackageCatalogItemCreator.cs
dtivel: I won't be implementing this suggestion as part of this PR but feel free to open an issue.  This check doesn't exist today, and in general, as part of my V3 work I'm reluctant to make non-critical behavioral changes like this.

src/Catalog/PackageCatalogItem.cs
```diff
@@ -12,25 +12,27 @@ namespace NuGet.Services.Metadata.Catalog
...
-        private readonly NupkgMetadata _nupkgMetadata;
-        private readonly DateTime? _createdDate;
-        private readonly DateTime? _lastEditedDate;
-        private readonly DateTime? _publishedDate;
         private string _id;
         private string _fullVersion;
         private string _normalizedVersion;
 
+        // These properties are public only to facilitate testing.
+        public NupkgMetadata NupkgMetadata { get; }
```
agr: Make them `internal` then maybe, and add test assembly in `InternalsVisibleTo`?

src/Catalog/PackageCatalogItemCreator.cs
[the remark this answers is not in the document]
agr: Isn't `HttpClient` an exception to it? As in it is better to have single instance of `HttpClient` in the app, pass it around and never dispose of it?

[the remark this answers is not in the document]
dtivel: Yes, it's an [anti-pattern](https://docs.microsoft.com/en-us/azure/architecture/antipatterns/improper-instantiation/) to create and dispose of an `HttpClient` with each request.

src/Catalog/PackageCatalogItem.cs
dtivel: I didn't do it because it's not worth the effort now.

PR #323, opened 2018-07-13
Feed2Catalog:  parallelize package downloads

dtivel:
Resolve https://github.com/NuGet/NuGetGallery/issues/6126.

Parallelize package downloads to improve Feed2Catalog throughput.

On a test machine I observed a ~57% improvement in processing throughput given other json-ld fixes and 2,000 packages following production package size and package entry count distributions to the nearest percentile.

| Trial | Before (seconds) | After (seconds) |
| --- | --- | --- |
| 1 | 2,627.11 | 1,200.86 |
| 2 | 2,591.97 | 1,100.23 |
| 3 | 2,847.53 | 1,173.27 |
| Average | 2,688.87 | 1,158.12 |


src/Ng/Jobs/Package2CatalogJob.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
 using System.Net.Http;
```
loic-sharma: Is this `using` necessary? It doesn't look like you introduced any new types in this file

src/Catalog/Helpers/FeedHelpers.cs
```diff
@@ -355,6 +308,78 @@ private static DateTime ForceUtc(DateTime date)
...
+                if (response.StatusCode == System.Net.HttpStatusCode.NotFound)
+                {
+                    //  the feed is out of sync with the actual package storage - if we don't have the package there is nothing to be done we might as well move onto the next package
+                    logger?.LogWarning("Unable to download: {PackageDetailsContentUri}. Http status: {HttpStatusCode}", feedPackageDetails.ContentUri, response.StatusCode);
+                }
+                else
+                {
+                    //  this should trigger a restart - of this program - and not move the cursor forward
+                    logger?.LogError("Unable to download: {PackageDetailsContentUri}. Http status: {HttpStatusCode}", feedPackageDetails.ContentUri, response.StatusCode);
+                    throw new Exception(
```
loic-sharma: How big is the batch of packages given to `DownloadMetadata2CatalogAsync`? If the batch is big, this behavior of throwing without retrying may be expensive.

src/Catalog/Helpers/FeedHelpers.cs
```diff
@@ -355,6 +308,78 @@ private static DateTime ForceUtc(DateTime date)
...
+                        feedPackageDetails.PublishedDate);
+
+                    if (workItem.PackageCatalogItem != null)
+                    {
+                        logger?.LogInformation("Add metadata from: {PackageDetailsContentUri}", feedPackageDetails.ContentUri);
+                    }
+                    else
+                    {
+                        logger?.LogWarning("Unable to extract metadata from: {PackageDetailsContentUri}", feedPackageDetails.ContentUri);
+                    }
```
loic-sharma: Is this the right behavior? This silent failure means that we have a package on v2 but not v3.

Also, could you change this line to log the full exception instead of just its type and message: [https://github.com/NuGet/NuGet.Services.Metadata/blob/master/src/Catalog/Helpers/Utils.cs#L513](https://github.com/NuGet/NuGet.Services.Metadata/blob/master/src/Catalog/Helpers/Utils.cs#L513)

loic-sharma: Never mind this comment as well.

src/Catalog/Helpers/FeedHelpers.cs
```diff
@@ -247,18 +247,24 @@ private static DateTime ForceUtc(DateTime date)
...
             DateTime lastDeleted,
+            int maxDegreeOfParallelism,
             bool? createdPackages,
             CancellationToken cancellationToken,
             ITelemetryService telemetryService,
             ILogger logger)
         {
+            if (maxDegreeOfParallelism < 1)
+            {
+                throw new ArgumentOutOfRangeException(nameof(maxDegreeOfParallelism));
```
agr: Human friendly message about what exactly is wrong with the value?

src/Catalog/Helpers/FeedHelpers.cs
```diff
@@ -268,76 +274,23 @@ private static DateTime ForceUtc(DateTime date)
                 return lastDate;
             }
 
-            foreach (var entry in packages)
+            // Flatten the sorted list.
+            var workItems = packages.SelectMany(
+                    pair => pair.Value.Select(
+                        details => new PackageWorkItem(pair.Key, details)))
+                .ToArray();
```
agr: Unnecessary `ToArray()`? `ForEachAsync` call below should work fine with any `IEnumerable`.

dtivel: Not unnecessary.  The enumerable is used [below](https://github.com/NuGet/NuGet.Services.Metadata/pull/323/files/1e39f6ca9a5a2dd713f8bca15fb7415fb199b2a7#diff-ef9bbac124ec39795cf902ad6b03bfc2R291) again.

PR #324, opened 2018-07-13
Update json-ld.net version

dtivel:
Resolve https://github.com/NuGet/Engineering/issues/1525 and progress on https://github.com/NuGet/NuGetGallery/issues/6128.


dtivel:
Need to strong name sign first.

PR #329, opened 2018-07-18
Feed2Catalog:  replace inner loops with one outer loop

dtivel:
Resolve https://github.com/NuGet/Engineering/issues/1526.

This PR replaces the 3 inner loops in Feed2Catalog's main job loop with one outer loop.  This will enable Feed2Catalog to process deleted/created/edited packages without being "stuck" in a single loop for extended periods of time when there is a sustained flow of packages in one particular query (e.g.:  edited).


dtivel:
Investigating test failure.  Doesn't repro on local machine.

loic-sharma:
>             using (Logger.BeginScope($"Logging for {TelemetryConstants.Destination}", Destination.AbsoluteUri))

Does this log line work? I don't think the `Destination.AbsoluteUri` will be put anywhere...

---
Refers to: src/Ng/Jobs/Feed2CatalogJob.cs:114 in dc68473. [](commit_id = dc6847361046e1a38cfa1cd1790c28e49c4e4344, deletion_comment = False)

joelverhagen:
One nice to have is some tests that assert the order of processing.

loic-sharma:
It'd also be nice to have a test that, with the previous code, would've gotten stuck and verifying each loop gets an equal chance.

PR #335, opened 2018-08-09
Catalog2Dnx:  improve throughput

dtivel:
Progress on https://github.com/NuGet/NuGetGallery/issues/6267.

This is an optimization for Azure storage-based usage of Catalog2Dnx.  This change:

1.  Downloads a package's .nuspec file from blob storage without downloading the entire package.
1.  Performs server-side package copies instead of client-side copies.

This results in huge savings in time and disk I/O.

Catalog2Dnx will fall back to downloading a package via HTTP if necessary.

On a local test machine with a China storage destination I observed a ~60% improvement in processing throughput with ~4,800 packages following production package size and package entry count distributions to the nearest percentile.

Notes:

1.  Previously, Catalog2Dnx downloaded a package, extracted the .nuspec, then uploaded both to flat container.  The download and upload were atomic, especially around blob properties.  To achieve the same eventual results, Catalog2Dnx's package sync/no-op check now includes a destination blob properties check.
1.  Microsoft.Azure.Storage.DataMovement 0.6.0 requires WindowsAzure.Storage 8.1.4 or later.  All projects in the solution reference WindowsAzure.Storage 8.2.1, except NuGet.Services.Metadata.Catalog.Monitoring which referenced WindowsAzure.Storage 7.1.2.  However, NuGet.Services.Metadata.Catalog.Monitoring references another project in the solution, so I updated NuGet.Services.Metadata.Catalog.Monitoring to use the same WindowsAzure.Storage 8.2.1.  This updated some other dependencies.


src/Catalog/Dnx/DnxCatalogCollector.cs
```diff
@@ -36,6 +40,8 @@ public class DnxCatalogCollector : CommitCollector
             : base(index, telemetryService, handlerFunc, httpClientTimeout)
         {
             _storageFactory = storageFactory ?? throw new ArgumentNullException(nameof(storageFactory));
+            _sourceStorage = preferredPackageSourceStorage;
```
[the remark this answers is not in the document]
dtivel: It's OK to be `null`.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -140,26 +142,45 @@ protected override async Task OnSave(Uri resourceUri, StorageContent content, Ca
...
-                    await blob.UploadFromStreamAsync(stream,
-                                                     accessCondition: null,
-                                                     options: _blobRequestOptions,
-                                                     operationContext: null,
-                                                     cancellationToken: cancellationToken);
-                    Trace.WriteLine(String.Format("Saved uncompressed blob {0} to container {1}", blob.Uri.ToString(), _directory.Container.Name));
+                    await TransferManager.CopyAsync(copiableContent.SourceBlob, blob, isServiceCopy: true);
+
+                    // Copying (above) did not apply these properties so apply them again.
+                    blob.Properties.ContentType = content.ContentType;
```
skofman1: do we need to apply all properties? (future proofing)

src/Catalog/Dnx/DnxCatalogCollector.cs
```diff
@@ -94,20 +98,23 @@ private async Task<IEnumerable<CatalogEntry>> ProcessCatalogEntriesAsync(Collect
...
                     using (_telemetryService.TrackDuration(TelemetryConstants.ProcessPackageDetailsSeconds, properties))
                     {
-                        var sourceUri = new Uri(ContentBaseAddress, $"{packageId}.{packageVersion}.nupkg");
+                        var sourceUri = new Uri(_contentBaseAddress, $"{packageId}.{packageVersion}.nupkg");
                         var destinationStorage = _storageFactory.Create(packageId);
                         var destinationUri = destinationStorage.GetUri(DnxMaker.GetRelativeAddressNupkg(packageId, packageVersion));
 
                         var isNupkgSynchronized = await destinationStorage.AreSynchronized(sourceUri, destinationUri);
                         var isPackageInIndex = await _dnxMaker.HasPackageInIndexAsync(destinationStorage, packageId, packageVersion, cancellationToken);
+                        var areRequiredPropertiesPresent = await AreRequiredPropertiesPresentAsync(destinationStorage, destinationUri);
```
loic-sharma: >AreRequiredPropertiesPresentAsync [](start = 65, length = 33)

Out of curiosity, why is this needed? Is this so allow us to fix packages that are missing properties by reflowing? Do we have cases of blobs in the flat container without these properties?

src/Catalog/Dnx/DnxCatalogCollector.cs
```diff
@@ -173,6 +195,70 @@ private async Task UpdatePackageVersionIndexAsync(IEnumerable<CatalogEntry> cata
             string version,
             Uri sourceUri,
             CancellationToken cancellationToken)
+        {
+            var shouldUseSourceStorage = _sourceStorage != null;
+            var sourceBlob = await GetBlockBlobAsync(id, version) as AzureCloudBlockBlob;
```
loic-sharma: >GetBlockBlobAsync [](start = 35, length = 17)

What do you think of renaming `GetBlockBlobAsync` to something like `GetSourcePackageBlobAsync` to clearly indicate it gets the source blob?

src/Catalog/Dnx/DnxMaker.cs
```diff
@@ -195,7 +241,27 @@ private StorageContent CreateContent(IEnumerable<string> versions)
...
+        private async Task<Uri> CopyNupkgAsync(CloudBlockBlob sourceBlob, Storage storage, string id, string version, CancellationToken cancellationToken)
+        {
+            var relativeUri = GetRelativeAddressNupkg(id, version);
+            var nupkgUri = new Uri(storage.BaseAddress, relativeUri);
+            var content = new AzureServerSideCopiableContent(
+                sourceBlob,
+                DnxConstants.ApplicationOctetStreamContentType,
+                DnxConstants.DefaultCacheControl);
+
+            await storage.Save(nupkgUri, content, cancellationToken);
```
joelverhagen: Why is this `Save`? The implementation must check types to make this work. Seems like a broken abstraction. Can't we have a `Copy` method that only knows how to copy? And let `Save` only upload?

src/Catalog/Dnx/DnxCatalogCollector.cs
```diff
@@ -173,6 +195,70 @@ private async Task UpdatePackageVersionIndexAsync(IEnumerable<CatalogEntry> cata
...
+
+            if (sourceBlob != null)
+            {
+                await sourceBlob.FetchAttributesAsync(cancellationToken);
+
+                // It's possible (though unlikely) that the blob may change between reads.  Reading a blob with a
+                // single GET request returns the whole blob in a consistent state, but we're reading the blob many
+                // different times.  To detect the blob changing between reads, we check the ETag again later.
+                // If the ETag's differ, we'll fall back to using a single HTTP GET request.
+                var etag = sourceBlob.ETag;
```
joelverhagen: This is only necessary in the off chance that we have to catalog2dnx instances running? Or is there some other way to have someone else updating the blob?

src/Catalog/Dnx/DnxCatalogCollector.cs
```diff
@@ -173,6 +195,70 @@ private async Task UpdatePackageVersionIndexAsync(IEnumerable<CatalogEntry> cata
...
+
+                string nuspec = await GetNuspecAsync(sourceBlob, id, cancellationToken);
+
+                if (string.IsNullOrEmpty(nuspec))
+                {
+                    _logger.LogWarning("No .nuspec available for {Id}/{Version}. Skipping.", id, version);
```
loic-sharma: >Skipping [](start = 81, length = 8)

Is `Skipping` correct?  It looks like this package will be processed by `ProcessPackageDetailsViaHttpAsync`

src/Catalog/Persistence/AzureCloudBlockBlob.cs
```diff
@@ -13,35 +13,30 @@ namespace NuGet.Services.Metadata.Catalog.Persistence
 {
     public sealed class AzureCloudBlockBlob : ICloudBlockBlob
     {
-        private readonly CloudBlockBlob _blob;
+        public CloudBlockBlob Blob { get; }
```
loic-sharma: Super nit: usually properties are placed after the constructors. Same for `ETag` and `Uri`.

src/Catalog/Dnx/DnxCatalogCollector.cs
dtivel: Currently we set blob properties and upload blob content in one operation.  As part of my changes I don't want to introduce scenarios where a blob could be uploaded but not have the correct blob properties (say, because of an exception).

The properties may be different or may have different values in flat container.

src/Catalog/Persistence/AzureStorage.cs
dtivel: No, the only properties we set today are ContentType and CacheControl.

PR #338, opened 2018-08-16
Add expanded signing targets probing

joelverhagen:
Based off of orchestrator:

https://github.com/NuGet/NuGet.Jobs/blob/0d7b59a547891bfe57283a5c6aaad343824e252f/src/NuGet.Services.Validation.Orchestrator/NuGet.Services.Validation.Orchestrator.csproj#L175-L180

Progress on https://github.com/NuGet/Engineering/issues/1644


joelverhagen:
`BUILD_SOURCESDIRECTORY` is used during `NuGetDeployment` builds.

PR #339, opened 2018-08-17
Rewrite SafeRoleEnvironment to not use Assembly.LoadWithPartialName

scottbommarito:
`Assembly.LoadWithPartialName` is insecure (as caught by Roslyn).

Fortunately, we don't have to use reflection at all, because we can simply `try-catch` the `RoleEnvironment` accesses instead.


scottbommarito:
Deploying to DEV test.

src/NuGet.Services.BasicSearch/SafeRoleEnvironment.cs
```diff
@@ -1,93 +1,60 @@
...
+                return true;
             }
             catch (Exception e)
             {
+                // If the assembly is not available, a file load exception will be thrown.
+                // Otherwise, it must be an exception thrown by the loaded assembly itself.
                 if (!(e is FileNotFoundException || e is FileLoadException || e is BadImageFormatException))
                 {
                     throw;
                 }
```
loic-sharma: If these methods are called often, this catching will be expensive. We should try to call once, and if we get one of these exceptions, we should enable a circuit breaker.

scottbommarito: added...deploying to DEV again to test

src/NuGet.Services.BasicSearch/SafeRoleEnvironment.cs
```diff
@@ -1,93 +1,66 @@
...
-
-            return string.Empty;
+            return TryGetField(() => RoleEnvironment.DeploymentId, out id);
         }
 
-        public static string GetLocalResourceRootPath(string name)
+        public static bool TryGetLocalResourceRootPath(string name, out string path)
         {
-            return RoleEnvironment.GetLocalResource(name).RootPath;
+            return TryGetField(() => RoleEnvironment.GetLocalResource(name).RootPath, out path);
```
ryuyu: Can GetLocalResource return null? What is the expected behaviour in that case?

scottbommarito: If `GetLocalResource` is null, it will throw a `NullReferenceException`, which will be bubbled up to the caller of this method. This is the same behavior as before.

src/NuGet.Services.BasicSearch/SafeRoleEnvironment.cs
```diff
@@ -1,93 +1,66 @@
...
         }
 
-        /// <summary>
-        /// Loads and returns the latest available version of the service runtime assembly.
-        /// </summary>
-        /// <returns>Loaded assembly, if any.</returns>
-        private static Assembly GetServiceRuntimeAssembly()
+        private static bool _assemblyIsAvailable = true;
+        private static bool TryGetField<T>(Func<T> getValue, out T value)
```
skofman1: no need for generics here. The return value is always string.

skofman1:
How did you test the code (besides deploying to dev)?

src/NuGet.Services.BasicSearch/SafeRoleEnvironment.cs
scottbommarito: Good catch.

The idea, however, is that this will be a proxy for every use of `RoleEnvironment`, so keeping it like this will make it easier for us to add additional methods if we need to access fields of `RoleEnvironment` that do not return a `string`.

If I had noticed that all of our uses of `RoleEnvironment` were `string`s initially I would have made it a `Func<string>`, but at this point there's no reason for the change.

skofman1: This introduces an unnecessary complexity to the code. Agree that it's not a bug deal.

PR #342, opened 2018-08-21
MonitoringProcessor:  improve throughput

dtivel:
Progress on https://github.com/NuGet/NuGetGallery/issues/6327.

The MonitoringProcessor job has 5 validators.  Each validator makes its own calls to retrieve V2/V3 resources, and many of these calls are duplicates.

This change memoizes a V2/V3 resource fetch across all validators for a single package validation.

Before this change, one package validation made the following requests:

Call Count | URL
--- | :---
5 | https://dev.nugettest.org/api/v2/Packages?$filter=true and Id eq 'NHibernate' and NormalizedVersion eq '3.0.0.4000'&semVerLevel=2.0.0
5 | https://az635243.vo.msecnd.net/v3-catalog0/data/2015.12.18.18.31.00/nhibernate.3.0.0.4000.json
10 | https://dev.nugettest.org/api/v2/FindPackagesById()?id='NHibernate'&semVerLevel=2.0.0
10 | https://apidev.nugettest.org/v3-registration3-gz-semver2/nhibernate/index.json
4 | https://dev.nugettest.org/api/v2/Packages(Id='NHibernate',Version='3.0.0.4000')
4 | https://apidev.nugettest.org/v3-registration3-gz-semver2/nhibernate/3.0.0.4000.json

After this change, each URL is hit once.  On my local dev machine I observed a ~23% average improvement, but to be fair there's a lot of variation in execution times.  Still, the reduction in web calls should also help alleviate performance issues from high garbage collection.  (During one performance trace I observed ~20% CPU time spent in garbage collection.)

Also:

* fixed a minor bug where validators were created twice
* renamed some stuff
* reformatted documents


src/Ng/Scripts/MonitoringProcessor.cmd
```diff
@@ -1,14 +1,35 @@
...
 
-	title #{Jobs.monitoringprocessor.Title}
-    
-	start /w .\Ng.exe monitoringprocessor -source #{Jobs.common.v3.Source} -index #{Jobs.common.v3.index} -gallery #{Jobs.common.v3.f2c.Gallery} -endpointsToTest "#{Jobs.endpointmonitoring.EndpointsToTest}" -statusFolder #{Jobs.endpointmonitoring.StatusFolder} -storageType azure -storageAccountName #{Jobs.common.v3.c2r.StorageAccountName} -storageKeyValue #{Jobs.common.v3.c2r.StorageAccountKey} -storageContainer #{Jobs.endpointmonitoring.MonitoringContainer} -storageQueueName #{Jobs.endpointmonitoring.PackageValidatorQueue} -storageTypeAuditing azure -storageAccountNameAuditing #{Jobs.feed2catalogv3.AuditingStorageAccountName} -storageKeyValueAuditing #{Jobs.feed2catalogv3.AuditingStorageAccountKey} -storageContainerAuditing auditing -storagePathAuditing package -instrumentationkey #{Jobs.common.v3.Logging.InstrumentationKey} -vaultName #{Deployment.Azure.KeyVault.VaultName} -clientId #{Deployment.Azure.KeyVault.ClientId} -certificateThumbprint #{Deployment.Azure.KeyVault.CertificateThumbprint} -verbose true -interval #{Jobs.common.v3.Interval}
+start /w .\Ng.exe monitoringprocessor ^
+-source #{Jobs.common.v3.Source} ^
```
joelverhagen: Can we indent here?

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/ValidationContext.cs
```diff
@@ -43,20 +50,43 @@ public class ValidationContext
...
             Client = client;
             CancellationToken = token;
         }
+
+        public Task<bool> GetMemoizedBoolAsync(string key, Lazy<Task<bool>> lazyTask)
+        {
+            return _boolCache.GetOrAdd(key, lazyTask).Value;
+        }
+
+        public Task<PackageRegistrationIndexMetadata> GetMemoizedIndexAsync(
```
joelverhagen: I think calling it "Cached" instead of "Memoized" would be a bit clearer to the average reader.

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/ValidationContext.cs
```diff
@@ -43,20 +50,43 @@ public class ValidationContext
...
+            : this()
         {
             Package = package;
             Entries = entries;
             DeletionAuditEntries = deletionAuditEntries;
             Client = client;
             CancellationToken = token;
         }
+
+        public Task<bool> GetMemoizedBoolAsync(string key, Lazy<Task<bool>> lazyTask)
```
joelverhagen: All three of these could have the same method name. The parameter types are different.

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/Registration/RegistrationIndexValidator.cs
```diff
@@ -18,34 +18,54 @@ public abstract class RegistrationIndexValidator : RegistrationValidator
         {
         }
 
-        protected override async Task<bool> ShouldRun(ValidationContext data)
+        protected override async Task<bool> ShouldRunAsync(ValidationContext context)
         {
-            return 
-                await base.ShouldRun(data) &&
+            var shouldRunTask = context.GetCachedResultAsync(
+                Keys.ShouldRunAsync,
```
scottbommarito: I'm not a big fan of this syntax:
- it requires the validators to know the cache exists
- it requires the validators to specify a lambda for the operation they want to do
- it requires the validators to specify the exact and correct key for the operation they want to do

Ideally, the validators shouldn't be aware of the context's cache. The context should everything automatically.

What if:
- the context had `Lazy`s for `V2RegistrationIndex`, `V2RegistrationLeaf`, `V3RegistrationIndex`, `V3RegistrationLeaf`, etc (anything that requires a web request)
- validators were rewritten to only access these fields instead of making web requests themselves

Another option:
- move methods to access `V2RegistrationIndex`, `V2RegistrationLeaf`, `V3RegistrationIndex`, `V3RegistrationLeaf`, etc to the context
- cache the returns in the context

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/Registration/RegistrationIndexValidator.cs
```diff
@@ -18,34 +18,54 @@ public abstract class RegistrationIndexValidator : RegistrationValidator
         {
         }
 
-        protected override async Task<bool> ShouldRun(ValidationContext data)
+        protected override async Task<bool> ShouldRunAsync(ValidationContext context)
         {
-            return 
-                await base.ShouldRun(data) &&
+            var shouldRunTask = context.GetCachedResultAsync(
```
scottbommarito: If you're going to cache the result of the base `Validator.ShouldRunAsync`, you should do it in the base `Validator.ShouldRunAsync` method itself. Same applies to `RegistrationValidator.GetIndexAsync` and `RegistrationValidator.GetLeafAsync` and even `RegistrationIndexValidator.ShouldRunAsync` itself

PR #345, opened 2018-08-23
Add MicroBuild dependency and signing of output DLLs

joelverhagen:
Also, remove some dead code.

This can be safely merged with no functional impact on existing (unsigned) deployments.

Progress on https://github.com/NuGet/Engineering/issues/1644.


joelverhagen:
I've only added MicroBuild signing to search service and its project dependencies.

src/NuGet.ApplicationInsights.Owin/NuGet.ApplicationInsights.Owin.csproj
```diff
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\..\packages\MicroBuild.Core.0.3.0\build\MicroBuild.Core.props" Condition="Exists('..\..\packages\MicroBuild.Core.0.3.0\build\MicroBuild.Core.props')" />
```
loic-sharma: >MicroBuild.Core.0.3.0 [](start = 34, length = 21)

If we bump up the MicroBuild version, we'll have to hunt for all the places that reference the version. I'm worried this could be easy to mess up without realizing. Would it be possible to have a `MicroBuildVersion` property?

joelverhagen: This is done by NuGet client when you update the package.

joelverhagen: Yup, props and targets.

PR #350, opened 2018-09-11
Catalog2Registration:  parallelize commit processing

dtivel:
Progress on https://github.com/NuGet/NuGetGallery/issues/6363.

Reprocessing production catalog entries from 2018-08-23 00:00Z to 2018-08-30 00:00Z was ~60% faster with this change.

Trial | Before (seconds) | After (seconds)
-- | -- | --
1 | 21,759.79 | 8,389.36
2 | 22,345.73 | 8,587.67
3 | 21,774.72 | 8,132.25
Average | 21,960.08 | 8,369.76


src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -365,7 +365,14 @@ protected override async Task OnDeleteAsync(Uri resourceUri, CancellationToken c
...
         {
-            return new Uri(_directory.Uri, name);
+            var baseUri = _directory.Uri.AbsoluteUri;
+
+            if (baseUri.EndsWith("/"))
+            {
+                return new Uri($"{baseUri}{name}", UriKind.Absolute);
+            }
+
+            return new Uri($"{baseUri}/{name}", UriKind.Absolute);
```
dtivel: This is a minor bug fix.  The `Uri` constructor overload that takes a `Uri` object as its first argument wouldn't use the absolute URI.  The result was when saving registration cursor, this would be logged:

    SaveAsync https://<account>.blob.core.windows.net/cursor.json

With this change, it now logs the container:

    SaveAsync https://<account>.blob.core.windows.net/v3-registration3/cursor.json

src/Catalog/Registration/RegistrationCollector.cs
```diff
@@ -131,5 +281,155 @@ public static ShouldIncludeRegistrationPackage GetShouldIncludeRegistrationPacka
 
             return (k, u, g) => !NuGetVersionUtility.IsGraphSemVer2(k.Version, u, g);
         }
+
+        private static void DequeueBatches(Queue<BatchTask> batches, Func<BatchTask, bool> isMatch)
```
skofman1: Here you dequeue batches only while the front matches. Consider changing the name to DequeueBatchesWhileMatches or something else that explains that

src/Catalog/Registration/RegistrationCollector.cs
```diff
@@ -67,14 +83,148 @@ protected override Task<IEnumerable<CatalogItemBatch>> CreateBatchesAsync(IEnume
...
+
+                            await front.SaveAsync(cancellationToken);
+
+                            Trace.TraceInformation($"{nameof(RegistrationCollector)}.{nameof(FetchAsync)} {nameof(front)}.{nameof(front.Value)} saved since timestamp changed from previous: {{0}}", front);
+
+                            DequeueBatches(processingBatches, batch => batch.CommitTimeStamp == commitTimeStamp);
+
+                            commitBatchTasksMap.Remove(commitTimeStamp);
+                        }
+                        else
```
skofman1: when will we ever get here?

dtivel: Never.  I removed it.

PR #355, opened 2018-09-13
Add secondary indexes

loic-sharma:
Adds scripts to build a lucene index in a secondary blob container. These indexes will be created by a VM hosted in a secondary region.

### DIFF
--- a/src/Ng/Scripts/Catalog2LuceneV3-SouthAsia.cmd
+++ b/src/Ng/Scripts/Catalog2LuceneV3-SouthAsia.cmd
@@ -1,18 +1,18 @@
 @echo OFF
-	
+
 cd Ng
 
 :Top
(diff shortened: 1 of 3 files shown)


src/Ng/Scripts/Catalog2LuceneV3-SouthAsia.cmd
```diff
@@ -0,0 +1,25 @@
+@echo OFF
+
+cd Ng
+
+:Top
+echo "Starting job - #{Jobs.catalog2lucenev3reg2southasia.Title}"
```
joelverhagen: `southasia` seems a bit odd. Perhaps `southeastasia`? Let's adhere to precedent

loic-sharma: Fixed

PR #359, opened 2018-09-19
Migrate projects to use PackageReference instead of packages.config

joelverhagen:
This is to aid in future Azure Search changes. This also allows us to update our NuGetGallery.Core reference which is not 4.6.2.

The following other changes were made, very much like the NuGetGallery migration done by @chenriksson.

- Move pack to `msbuild` pack for the dependency packages
- Move to .NET 4.6.2
- Move nuspec metadata to csproj
- Auto-generate binding redirects
  - The web.config for BasicSearch has explicit binding redirects since the auto-generation was not working and there was a warning at build time.
- Move tool versions to 15.0
- Remove NuGet 2.x (NuGet.Core) dependency by copying one type in: `DisposableAction`
- Remove MennoCopy (unused tool) and very old, unused test project.


joelverhagen:
Address https://github.com/nuget/nugetgallery/issues/6432.

src/NuGet.Indexing/DisposableAction.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+using System;
```
chenriksson: nit: newline after copyright

src/V3PerPackage/App.config
```diff
@@ -1,6 +1,3 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
-    <startup> 
-        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.1"/>
-    </startup>
-</configuration>
+<startup><supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.2"/></startup></configuration>
```
chenriksson: nit: would keep more readable formatting

src/NuGet.Indexing/DisposableAction.cs
chenriksson: This was pulled in via dependency before?

joelverhagen: Yes, NuGet.Core.

src/V3PerPackage/App.config
joelverhagen: This was generated by the tool. I'll try to format these.

src/NuGet.Indexing/DisposableAction.cs
joelverhagen: Added newline.

tests/NgTests/NgTests.csproj
```diff
@@ -32,233 +32,26 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
+  <PropertyGroup>
```
loic-sharma: >PropertyGroup [](start = 3, length = 13)

Can we merge the `PropertyGroup`s with no `Condition` to the main `PropertyGroup`?

loic-sharma: It looks like most `csproj`s have similar issues. Not a big deal, feel free to ignore.

joelverhagen: We took this approach in gallery. It's just a way to call out the properties use for pack.

PR #367, opened 2018-10-09
Add VersionListDataClient to read and write the version list JSON

joelverhagen:

Note that this is going into the `azuresearch` feature branch.

Example JSON:

```json
{
  "VersionProperties": {
    "1.0.0-beta.2": {
      "Listed": true,
      "SemVer2": true
    },
    "1.0.0-beta.10": {
      "Listed": true,
      "SemVer2": true
    },
    "2.0.0": {
      "Listed": true
    },
    "10.0.0": {}
  }
}
```

### DIFF
--- a/src/NuGet.Services.AzureSearch/VersionList/VersionListData.cs
+++ b/src/NuGet.Services.AzureSearch/VersionList/VersionListData.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace NuGet.Services.AzureSearch
+{
+    public class VersionListData
+    {
+        [JsonConstructor]
+        public VersionListData(Dictionary<string, VersionPropertiesData> versionProperties)
+        {
+            VersionProperties = versionProperties ?? throw new ArgumentNullException(nameof(versionProperties));
+        }
+
+        /// <summary>
+        /// A dictionary of all versions currently available for this package. This includes listed and unlisted
+        /// versions. The key is the full version string (can include build metadata).
+        /// </summary>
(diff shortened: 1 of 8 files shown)


src/NuGet.Services.AzureSearch/VersionList/VersionListData.cs
```diff
@@ -0,0 +1,28 @@
...
+        public VersionListData(Dictionary<string, VersionPropertiesData> versionProperties)
+        {
+            VersionProperties = new Dictionary<string, VersionPropertiesData>(
+                versionProperties,
+                StringComparer.OrdinalIgnoreCase);
+        }
+
+        /// <summary>
+        /// A dictionary of all versions currently available for this package. This includes listed and unlisted
+        /// versions. The key is the full version string (can include build metadata). The look-up is case-insensitive,
```
loic-sharma: >The look-up is case-insensitive [](start = 87, length = 31)

Should we have a unit test on this?

joelverhagen: Oh, I meant to remove this behavior and only use `VersionListData` for serialization concerns.

I'll remove this.

src/NuGet.Services.AzureSearch/VersionList/VersionListDataClient.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        public async Task ReplaceAsync(string id, VersionListData data, IAccessCondition accessCondition)
+        {
+            using (var stream = new MemoryStream())
+            {
+                using (var streamWriter = new StreamWriter(
+                    stream,
+                    encoding: new UTF8Encoding(encoderShouldEmitUTF8Identifier: false, throwOnInvalidBytes: true),
+                    bufferSize: 1024,
```
loic-sharma: Out of curiosity, what's the benefit of these encoding settings?

joelverhagen: This is the default for `StreamWriter`, but I want to keep the stream open.

PR #368, opened 2018-10-09
Add FilteredVersionList to determine SearchIndexChangeType

joelverhagen:

`SearchIndexChangeType` will be used by the caller (`catalog2azuresearch`) to determine what changes he has to make to each of 4 documents per ID.

### DIFF
--- a/src/NuGet.Services.AzureSearch/VersionList/FilteredVersionList.cs
+++ b/src/NuGet.Services.AzureSearch/VersionList/FilteredVersionList.cs
@@ -1,0 +1,203 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using NuGet.Versioning;
+
+namespace NuGet.Services.AzureSearch
+{
+    /// <summary>
+    /// This type tracks a version list for a specific <see cref="SearchFilters"/>. In other words, this implementation
+    /// assumes that a non-applicable version is never added to the list. In practice, <see cref="VersionLists"/> does
+    /// this filtering before adding a version to this list.
+    /// </summary>
+    internal class FilteredVersionList
+    {
+        internal NuGetVersion _latestOrNull;
+        internal readonly SortedList<NuGetVersion, FilteredVersionProperties> _versions;
+
+        public FilteredVersionList(IEnumerable<FilteredVersionProperties> versions)
+        {
+            if (versions == null)
+            {
+                throw new ArgumentNullException(nameof(versions));
+            }
+
+            _versions = new SortedList<NuGetVersion, FilteredVersionProperties>();
+
+            foreach (var version in versions)
+            {
+                _versions[version.ParsedVersion] = version;
+            }
+
+            _latestOrNull = CalculateLatest();
+        }
+
+        public string LatestOrNull
+        {
+            get
+            {
+                if (_latestOrNull == null)
+                {
+                    return null;
+                }
+
+                return _versions[_latestOrNull].FullVersion;
+            }
+        }
+
+        public IReadOnlyList<string> FullVersions => _versions
+            .Where(x => x.Value.Listed)
+            .Select(x => x.Value.FullVersion)
+            .ToList();
+
+        public SearchIndexChangeType Delete(NuGetVersion deleted)
+        {
+            var ctx = UpdateVersionList(
+                (v, p) => _versions.Remove(v),
+                deleted,
+                newProperties: null);
+
+            return DeleteFromSearchIndex(ctx);
+        }
+
+        public SearchIndexChangeType Upsert(FilteredVersionProperties addedProperties)
+        {
+            var ctx = UpdateVersionList(
+                (v, p) => _versions[v] = p,
+                addedProperties.ParsedVersion,
+                addedProperties);
+
+            return UpsertToSearchIndex(ctx);
+        }
+
+        private static SearchIndexChangeType DeleteFromSearchIndex(Context ctx)
+        {
+            if (ctx.NewLatest == null)
+            {
+                // If there is no longer a latest version, the search document should be deleted.
+                return SearchIndexChangeType.Delete;
+            }
+
+            if (ctx.NewLatest != ctx.OldLatest)
+            {
+                // If the latest version changes due to deletion, this can only happen if the existing latest version
+                // was the one that was deleted.
+                Assert(ctx.OldProperties != null, "This version should have existed before.");
+                Assert(ctx.OldProperties.Listed, "The existing version should have been listed.");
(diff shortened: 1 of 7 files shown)


joelverhagen:
Note that this implementation favors making more changes to the search index documents than would be strictly necessary in the happy path. Said another, I didn't even implement a `SearchIndexChangeType.NoChange` value. I do this so that reflow is as powerful as possible. If we see performance issues, we can introduce some no-oping logic.

src/NuGet.Services.AzureSearch/VersionList/FilteredVersionList.cs
```diff
@@ -0,0 +1,164 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using NuGet.Versioning;
+
+namespace NuGet.Services.AzureSearch
+{
+    /// <summary>
+    /// This type tracks a version list for a specific <see cref="SearchFilters"/>. In other words, this implementation
+    /// assumes that a non-applicable version is never added to the list. In practice, <see cref="VersionLists"/> does
```
joelverhagen: `VersionLists` will be added in a subsequent PR.

src/NuGet.Services.AzureSearch/VersionList/SearchIndexChangeType.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// <summary>
+        /// The latest version has been deleted or unlisted, so we need to replace the latest version metadata with an
+        /// older version's metadata.
+        /// </summary>
+        DowngradeLatest,
+
+        /// <summary>
+        /// Update the existing latest version's metadata and version list.
+        /// </summary>
+        UpdateExisting,
```
loic-sharma: >UpdateExisting [](start = 8, length = 14)

What do you think of `UpdateLatest`?

joelverhagen: Yeah, I was torn between `UpdateLatest` and `UpdateExisting`. Whatever name we choose, this enum value is what we use for both the "reflow latest" and "change latest to later version" cases.

If this one makes more sense to you, I'll change it.

src/NuGet.Services.AzureSearch/VersionList/FilteredVersionList.cs
```diff
@@ -0,0 +1,203 @@
...
+                oldProperties,
+                newProperties,
+                oldLatest,
+                _latest);
+        }
+
+        private NuGetVersion CalculateLatest()
+        {
+            for (var i = _versions.Count - 1; i >= 0; i--)
+            {
```
loic-sharma: Could you use LINQ here? Maybe something like:

```
_versions.Reverse()
  .Where(v =>v.Value.Listed)
  .Select(v => v.Key)
  .FirstOrDefault();
```

src/NuGet.Services.AzureSearch/VersionList/FilteredVersionList.cs
```diff
@@ -0,0 +1,203 @@
...
+
+            foreach (var version in versions)
+            {
+                _versions[version.ParsedVersion] = version;
+            }
+
+            _latest = CalculateLatest();
+        }
+
+        public string Latest
```
loic-sharma: >Latest [](start = 22, length = 6)

Thoughts on the name `LatestOrNull`?

joelverhagen: Sure, I like that better.

src/NuGet.Services.AzureSearch/VersionList/FilteredVersionList.cs
```diff
@@ -0,0 +1,203 @@
...
+    /// This type tracks a version list for a specific <see cref="SearchFilters"/>. In other words, this implementation
+    /// assumes that a non-applicable version is never added to the list. In practice, <see cref="VersionLists"/> does
+    /// this filtering before adding a version to this list.
+    /// </summary>
+    internal class FilteredVersionList
+    {
+        internal NuGetVersion _latest;
+        internal readonly SortedList<NuGetVersion, FilteredVersionProperties> _versions;
+
+        public FilteredVersionList(IEnumerable<FilteredVersionProperties> versions)
```
loic-sharma: >versions [](start = 74, length = 8)

This type is internal, but maybe we should add a null check on `versions`?

src/NuGet.Services.AzureSearch/VersionList/FilteredVersionList.cs
joelverhagen: I don't see a big clarity win here and the performance is not obviously (to the reader) equivalent -- although it should be similar.

I don't care much so I'll change it.

src/NuGet.Services.AzureSearch/VersionList/FilteredVersionList.cs
joelverhagen: sure

src/NuGet.Services.AzureSearch/VersionList/FilteredVersionList.cs
loic-sharma: I agree it's not a huge difference, but I do think that `.Reverse()` is slightly easier to read than a manual for loop.

---
In reply to: [224177873](https://github.com/NuGet/NuGet.Services.Metadata/pull/368#discussion_r224177873) [](ancestors = 224177873)

src/NuGet.Services.AzureSearch/VersionList/FilteredVersionList.cs
```diff
@@ -0,0 +1,203 @@
...
+            var ctx = UpdateVersionList(
+                (v, p) => _versions.Remove(v),
+                deleted,
+                newProperties: null);
+
+            return DeleteVersionFromSearchIndex(ctx);
+        }
+
+
+        public SearchIndexChangeType UpsertVersion(FilteredVersionProperties addedProperties)
```
loic-sharma: >SearchIndexChangeType [](start = 15, length = 21)

There's two empty lines above this guy

src/NuGet.Services.AzureSearch/VersionList/FilteredVersionList.cs
```diff
@@ -0,0 +1,203 @@
...
+                return SearchIndexChangeType.UpdateExisting;
+            }
+
+            if (ctx.NewLatest < ctx.OldLatest)
+            {
+                // If the new latest is a lower version than the old latest, this is a special case where we need to
+                // look up the old new latest's metadata.
+                Assert(ctx.NewLatest != ctx.ChangedVersion, "This case should already have been handled.");
+                Assert(ctx.OldLatest == ctx.ChangedVersion, "This case should already have been handled.");
+                Assert(!ctx.NewProperties.Listed, "A downgrade from an upserted version can only happen from an unlist.");
```
loic-sharma: Nice asserts 👍

PR #369, opened 2018-10-10
Add VersionLists to handle the 4 different views of the latest version

joelverhagen:

I also introduced a `Remove` method to `FilteredVersionList`. This has the same behavior as `Delete` for the search service, but will have a different behavior when I introduce the changes related to the service index.

When a non-applicable version is encountered, the search index should make sure it doesn't have that version at all (much like a `Delete`). For the hijack index, the non-applicable version should not be deleted but the `Latest` booleans should still be updated, for reflow. These specifics will come in a subsequent PR, but that is the motivation for treating `Delete` and `Remove` as separate gestures.

### DIFF
--- a/src/NuGet.Services.AzureSearch/VersionList/VersionLists.cs
+++ b/src/NuGet.Services.AzureSearch/VersionList/VersionLists.cs
@@ -1,0 +1,126 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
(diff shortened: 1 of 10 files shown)


src/NuGet.Services.AzureSearch/VersionList/VersionLists.cs
```diff
@@ -0,0 +1,126 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using NuGet.Versioning;
+using System;
+using System.Collections.Generic;
```
loic-sharma: `using`s are out of order here

PR #370, opened 2018-10-10
[Monitoring] Verify packages are repository signed

loic-sharma:
This adds a new validation to the monitoring job: `PackageIsRepositorySignedValidator`. This validation checks that the package on the flat container has a repository signature.

Part of https://github.com/NuGet/Engineering/issues/1627


src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/FlatContainer/FlatContainerValidator.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        protected Uri GetV3PackageUri(ValidationContext context)
+        {
+            // Based off DownloadResourceV3
+            // See: https://github.com/NuGet/NuGet.Client/blob/3803820961f4d61c06d07b179dab1d0439ec0d91/src/NuGet.Core/NuGet.Protocol/Resources/DownloadResourceV3.cs#L78
+            var id = context.Package.Id.ToLowerInvariant();
+            var version = context.Package.Version.ToNormalizedString().ToLowerInvariant();
+
+            return new Uri($"{_v3PackageBaseAddress.Value}/{id}/{version}/{id}.{version}.nupkg");
```
joelverhagen: This goes through CDN right? Two potential problems:

1. This download will increment the download count (not ideal).
1. This download might have a 2 minute cache, so you might not get the latest (repo-signed) version.

loic-sharma: Good catch. I've updated this PR so that the monitoring job downloads nupkgs from a configurable endpoint. We will use this to circumvent the CDN and download directly from the flat container.

src/NuGet.Services.Metadata.Catalog.Monitoring/Resources/PackageBaseAddressResource.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using NuGet.Protocol.Core.Types;
+
+namespace NuGet.Services.Metadata.Catalog.Monitoring
+{
+    public class PackageBaseAddressResource : INuGetResource
```
loic-sharma: You're probably wondering - why did I create a `INuGetResource` just so I could pass the `packageBaseAddress` config to the `FlatContainerValidator`? `Validator`s are created through reflection by the `ValidatorFactory`, and this factory only supports one kind of constructor for `Validator`s. Ideally, `Validator`s would be created through dependency injection, but this would be a large refactoring.

agr:
Need documentation for new classes and methods.

PR #373, opened 2018-10-17
 Merge hijack index changes by package version

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/6445.

Map the previously added `(NuGetVersion, HijackIndexChangeType)` pairs to a `MutableHijackIndexDocument` per version. The main goal of this work is to ensure that when a batch of catalog leafs comes in, we only issue a single update per Azure Search document.

This aggregation will be used twice:

1. To group together changes to the same version after a single version is upserted/deleted (this PR).
1. To group together changes to the same version after a batch of versions is upserted/deleted (future PR).

This adds a lot of guards to embed more intent into the implementation.

Also, added `IEquatable` on `MutableHijackIndexDocument` to ease unit testing.

### DIFF
--- a/src/NuGet.Services.AzureSearch/VersionList/Guard.cs
+++ b/src/NuGet.Services.AzureSearch/VersionList/Guard.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace NuGet.Services.AzureSearch
+{
+    internal static class Guard
+    {
+        /// <remarks>
+        /// We could use <see cref="System.Diagnostics.Debug.Assert(bool, string)"/> here, but it's preferable in this
+        /// case to even fail on a non-Debug build. The goal of this method is to allow the implementor to embed more
+        /// intent into the implementation so that future code changes are less likely to introduce bugs and so that
+        /// the implementation makes more sense to future readers. This is, of course, in addition to unit test
+        /// coverage.
+        /// </remarks>
+        public static void Assert(bool condition, string message)
+        {
+            if (!condition)
+            {
+                throw new InvalidOperationException(message);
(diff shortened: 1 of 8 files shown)


src/NuGet.Services.AzureSearch/VersionList/Guard.cs
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace NuGet.Services.AzureSearch
+{
+    internal static class Guard
+    {
+        /// <summary>
+        /// We could use <see cref="System.Diagnostics.Debug.Assert(bool, string)"/> here, but it's preferable
```
agr: This comment is worthy of &lt;remarks> section, not &lt;summary>. Maybe add couple of sentences about the purpose of the method?

src/NuGet.Services.AzureSearch/VersionList/MutableHijackIndexDocument.cs
```diff
@@ -0,0 +1,162 @@
...
+        public bool? LatestSemVer2 { get; private set; }
+
+        public void ApplyChange(SearchFilters searchFilters, HijackIndexChangeType changeType)
+        {
+            bool latest;
+            switch (changeType)
+            {
+                case HijackIndexChangeType.Delete:
+                    Guard.Assert(
+                        !Delete.HasValue || Delete.Value,
```
agr: `Delete != false` might be a bit more readable.

joelverhagen: Oh, nice point.

src/NuGet.Services.AzureSearch/VersionList/Guard.cs
joelverhagen: Done.

PR #374, opened 2018-10-18
Enable multiple version upserts or deletes at once

joelverhagen:

This PR allows a batch of version upserts or deletes to be applied to the version list at once. Note that the version list is an ID-specific thing so all of these changes are for a single package ID. The result new `ApplyChangesInternal` method of this will be a set of Azure Search document changes that the caller needs to push to the search and hijack Azure Search indexes.

The basic idea of the method is to apply the changes in descending SemVer 2.0.0 order. Descending order minimizes changes to the latest version. Each individual `VersionListChange` produces in a set of index changes (`MutableIndexChanges`). These index changes are merged into the index changes gotten from the applying the previous change to the version list.

Some existing method signatures changed to require the caller to call `NuGetVersion.Parse` before calling the method (instead of accepting a string and parsing in the method implementation). This simplifies the interface as it means the caller doesn't have to think about whether a string version parameter means full vs normalized vs original version.

For the hijack index, this is basically keeping track of all version document's latest booleans.

For the search index, this is managing some nuanced state transitions.

### DIFF
--- a/src/NuGet.Services.AzureSearch/VersionList/MutableHijackIndexDocument.cs
+++ b/src/NuGet.Services.AzureSearch/VersionList/MutableHijackIndexDocument.cs
@@ -57,6 +57,10 @@ namespace NuGet.Services.AzureSearch
                         UpdateMetadata != true,
                         "The hijack document has already been set to update metadata.");
                     Delete = true;
+                    foreach (var eachSearchFilters in MutableIndexChanges.AllSearchFilters)
+                    {
+                        SetLatest(eachSearchFilters, latest: null);
+                    }
                     return;
                 case HijackIndexChangeType.UpdateMetadata:
                     Guard.Assert(
@@ -74,16 +78,12 @@ namespace NuGet.Services.AzureSearch
                     latest = true;
                     break;
                 default:
-                    throw new NotImplementedException($"The change type '{changeType}' is not supported.");
+                    throw new NotImplementedException($"The hijack index change type '{changeType}' is not supported.");
             }
 
             Guard.Assert(
                 Delete != true,
                 "The hijack document has already been set to delete so the latest value can't be updated.");
-            var oldValue = GetLatest(searchFilters);
-            Guard.Assert(
-                !oldValue.HasValue || oldValue.Value == latest,
-                $"The latest value for search filters '{searchFilters}' cannot change from {oldValue} to {latest}.");
             SetLatest(searchFilters, latest);
         }
 
@@ -104,7 +104,7 @@ namespace NuGet.Services.AzureSearch
             }
         }
 
-        private void SetLatest(SearchFilters searchFilters, bool latest)
+        private void SetLatest(SearchFilters searchFilters, bool? latest)
         {
             switch (searchFilters)
             {
--- a/src/NuGet.Services.AzureSearch/VersionList/MutableIndexChanges.cs
+++ b/src/NuGet.Services.AzureSearch/VersionList/MutableIndexChanges.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using NuGet.Versioning;
+using SICT = NuGet.Services.AzureSearch.SearchIndexChangeType;
 
 namespace NuGet.Services.AzureSearch
 {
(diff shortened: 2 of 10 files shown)


joelverhagen:
Oh, and note that although catalog commits today either have all `PackageDetails` (i.e. upsert) or `PackageDelete`, the caller may batch multiple catalog commits together so it's possible for a sequence of `VersionListChange` instances to have both upserts and deletes.

Also, the caller is required to pass one `VersionListChange` per package version. If the caller has catalog leaves, this means taking the latest catalog commit timestamp per package version.

src/NuGet.Services.AzureSearch/VersionList/MutableIndexChanges.cs
```diff
@@ -13,22 +14,86 @@ namespace NuGet.Services.AzureSearch
...
+        /// are processed in descending version order so some state transition should not be possible.
+        /// </summary>
+        private static readonly IReadOnlyDictionary<KeyValuePair<SICT, SICT>, SICT> AcceptableTransitions
+            = new Dictionary<KeyValuePair<SICT, SICT>, SICT>
+            {
+                // Example: add an initial, listed version then add a lower, listed version
+                { KeyValuePair.Create(SICT.AddFirst, SICT.UpdateVersionList), SICT.AddFirst },
+
+                // Example: add an initial, unlisted version then add a lower, listed version
+                { KeyValuePair.Create(SICT.Delete, SICT.AddFirst), SICT.AddFirst },
```
loic-sharma: >        [](start = 83, length = 7)

There's empty whitespace after lines 35 and 38

src/NuGet.Services.AzureSearch/VersionList/MutableHijackIndexDocument.cs
```diff
@@ -57,6 +57,10 @@ public void ApplyChange(SearchFilters searchFilters, HijackIndexChangeType chang
                         UpdateMetadata != true,
                         "The hijack document has already been set to update metadata.");
                     Delete = true;
+                    foreach (var eachSearchFilters in MutableIndexChanges.AllSearchFilters)
+                    {
+                        SetLatest(eachSearchFilters, null);
```
loic-sharma: >SetLatest [](start = 24, length = 9)

`null` -> `latest: null`

src/NuGet.Services.AzureSearch/VersionList/MutableIndexChanges.cs
```diff
@@ -13,22 +14,86 @@ namespace NuGet.Services.AzureSearch
...
+            .Cast<SearchFilters>()
+            .ToList();
+
+        /// <summary>
+        /// This is a dictionary where the key is the state transition. The key is itself a
+        /// <see cref="KeyValuePair{TKey, TValue}"/> where the key is the existing state and the value is the added
+        /// state. The value of the dictionary is the resulting state from that transition. Remember that versions
+        /// are processed in descending version order so some state transition should not be possible.
+        /// </summary>
+        private static readonly IReadOnlyDictionary<KeyValuePair<SICT, SICT>, SICT> AcceptableTransitions
```
agr: Can we have domain-specific alternative to `KeyValuePair`? Its property names does not provide any context...

src/NuGet.Services.AzureSearch/VersionList/MutableIndexChanges.cs
```diff
@@ -37,40 +102,95 @@ public MutableIndexChanges()
...
+
+            var transition = KeyValuePair.Create(existingType, addedType);
+            if (AcceptableTransitions.TryGetValue(transition, out var result))
+            {
+                Search[searchFilters] = result;
+                return;
+            }
+
+            Guard.Assert(
+                false,
```
agr: Should we add `Guard.Fail`?

src/NuGet.Services.AzureSearch/VersionList/MutableIndexChanges.cs
joelverhagen: Yeah, I think I was lazy at that moment. I'll see what I can do.

src/NuGet.Services.AzureSearch/VersionList/MutableIndexChanges.cs
joelverhagen: Nice idea.

PR #377, opened 2018-10-22
Add ApplyChanges which returns an immutable set of change directives

joelverhagen:

The `Solidify` operation means taking a mutable, internal data structure to an immutable, public data structure. This gives guarantees to the callers that returned data structures won't be changing under his feet. Also, I take this time to map nullable booleans to `false` which is expected and simplifies caller experience.

Also, add some unit tests.


src/NuGet.Services.AzureSearch/VersionList/MutableIndexChanges.cs
```diff
@@ -190,6 +190,24 @@ private void MergeSearchIndexChanges(SearchFilters searchFilters, SICT addedType
...
+        public IndexChanges Solidify()
+        {
+            // Verify that the running list of hijack changes is the same as the pre-computed hijack document.
+            Guard.Assert(HijackDocuments.Count == HijackDocuments.Count, "The hijack document state has diverged.");
+            foreach (var pair in HijackChanges)
+            {
+                var expected = InitializeHijackIndexDocument(pair.Value);
+                var actual = HijackDocuments[pair.Key];
+                Guard.Assert(
+                    expected == actual,
```
joelverhagen: This is why I implemented `==`. It handles the null case and is more readable.

joelverhagen:
I renamed some things to differentiate between these models and the ones used to interact directly with Azure Search SDK.

loic-sharma:
>         public Dictionary<SearchFilters, SICT> Search { get; }

Would `SearchChanges` be a better name for this property?

---
Refers to: src/NuGet.Services.AzureSearch/VersionList/MutableIndexChanges.cs:88 in 43b8985. [](commit_id = 43b898533d462ce49d921bef70265aae30be11ff, deletion_comment = False)

joelverhagen:
> Would `SearchChanges` be a better name for this property?

It's part of the `IndexChanges` type so I was kind of thinking "changes" is implied.

loic-sharma:
> It's part of the `IndexChanges` type so I was kind of thinking "changes" is implied.

The hijack equivalent is called `HijackChanges`. I personally like the "changes" suffix :)

joelverhagen:
> The hijack equivalent is called `HijackChanges`. I personally like the "changes" suffix :)

Sure, sounds fine.

src/NuGet.Services.AzureSearch/VersionList/MutableIndexChanges.cs
```diff
@@ -190,6 +190,24 @@ private void MergeSearchIndexChanges(SearchFilters searchFilters, SICT addedType
             }
         }
 
+        public IndexChanges Solidify()
+        {
+            // Verify that the running list of hijack changes is the same as the pre-computed hijack document.
+            Guard.Assert(HijackDocuments.Count == HijackDocuments.Count, "The hijack document state has diverged.");
```
loic-sharma: >HijackDocuments.Count [](start = 50, length = 21)

Shoudl this be looking at `HijackChanges`?

joelverhagen: Nice catch!

PR #379, opened 2018-10-23
[Licenses] Update feed2catalog to ingest new license

zhhyu:
Fix issue: https://github.com/NuGet/NuGetGallery/issues/6520
Update the "nuspec.xslt" file to parse the new license type and content.


tests/CatalogTests/Helpers/UtilsTests.cs
```diff
@@ -109,5 +110,57 @@ private static MemoryStream GetPackageStream()
...
+        {
+            // Arrange
+            var stream = TestHelper.GetStream("Newtonsoft.Json.9.0.2-beta1.nupkg");
+            var metadata = Utils.GetNupkgMetadata(stream, packageHash: null);
+            var baseUrl = "http://example/";
+
+            // Actq
+            var graph = Utils.CreateNuspecGraph(metadata.Nuspec, baseUrl, normalizeXml: true);
+            var licenseFileTriples = graph.GetTriplesWithSubjectPredicate(
+                graph.CreateUriNode(new Uri(String.Concat(baseUrl, "newtonsoft.json.9.0.2-beta1.json"))),
```
joelverhagen: >new Uri(String.Concat(baseUrl, "newtonsoft.json.9.0.2-beta1.json") [](start = 36, length = 66)

Nit: make `new Uri(String.Concat(baseUrl, "newtonsoft.json.9.0.2-beta1.json")` into a local variable so it's not duplicated in 3 places

tests/CatalogTests/Helpers/UtilsTests.cs
```diff
@@ -109,5 +110,57 @@ private static MemoryStream GetPackageStream()
...
+            var graph = Utils.CreateNuspecGraph(metadata.Nuspec, baseUrl, normalizeXml: true);
+            var licenseTriples = graph.GetTriplesWithSubjectPredicate(
+                graph.CreateUriNode(new Uri(String.Concat(baseUrl, "testpackage.license.0.1.0.json"))),
+                graph.CreateUriNode(new Uri(String.Concat(Schema.Prefixes.NuGet, licenseType)))
+                );
+            var result = (LiteralNode)licenseTriples.First().Object;
+            
+            // Assert
+            Assert.Equal(1, licenseTriples.Count());
+            Assert.Equal(licenseContent, result.Value);
```
joelverhagen: Can we have assertions related to `licenseUrl` too? Basically, I think this should assert that no `licenseUrl` is mixed in here if the package has a `<license>` element

tests/CatalogTests/CatalogTests.csproj
```diff
@@ -175,6 +175,12 @@
     <None Include="TestData\OneValidDependencyOneEmptyIdWithGroups.0.1.0.nupkg">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Include="TestData\TestPackage.LicenseExpression.0.1.0.nupkg">
+	 <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
loic-sharma: >	 [](start = 0, length = 1)

Please use spaces instead of tabs. Same comment for line 182

tests/CatalogTests/Helpers/UtilsTests.cs
```diff
@@ -109,5 +110,57 @@ private static MemoryStream GetPackageStream()
...
+
+        [Fact]
+        public void GetNupkgMetadataWithLicenseUrl_ReturnsLicenseUrl()
+        {
+            // Arrange
+            var stream = TestHelper.GetStream("Newtonsoft.Json.9.0.2-beta1.nupkg");
+            var metadata = Utils.GetNupkgMetadata(stream, packageHash: null);
+            var baseUrl = "http://example/";
+
+            // Actq
```
loic-sharma: >Actq [](start = 15, length = 4)

`Actq` -> `Act`

tests/CatalogTests/Helpers/UtilsTests.cs
zhhyu: Changed to local variable.

tests/CatalogTests/Helpers/UtilsTests.cs
zhhyu: Assert licenseUrl now.

tests/CatalogTests/CatalogTests.csproj
zhhyu: Done.

tests/CatalogTests/Helpers/UtilsTests.cs
zhhyu: Sorry for the typo.

tests/CatalogTests/Helpers/UtilsTests.cs
agr: `licenseUrl` is orthogonal to the license information. In these test cases it is absent because the test packages have no license url specified in nuspec, not because a `license` tag is present. And this is correct behavior. License URL should be "mangled" downstream: catalog2registration and catalog2lucene has to check if there is `licenseExpression` or `licenseFile` present and ignore `licenseUrl` in those cases.

To reflect that in the tests, we probably need to include `licenseUrl` to test packages and check that it is picked up properly and has no effect on presence/value of `licenseFile` and `licenseExpression` properties.

zhhyu: Add other test cases now.

PR #380, opened 2018-10-25
Add initial documents for search and hijack index and improve latest API

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/6439.
Progress on https://github.com/NuGet/NuGetGallery/issues/6440.

This PR does two things.

**Add initial Azure Search document models.** There are multiple subsets of the fields, to support the different update patterns. For example, when updating the search index, you only want to update the `Versions` property (e.g. a non-latest version was added).

**Improve the API on `VersionLists` to get latest version information.** Previously you had to call multiple properties/methods to determine the latest version and the version list. Now all the data you need is returned as `LatestVersionInfo`. This will be used in the search index updater code reacting to change types like `AddFirst`. This also made some test assertions cleaner.

### DIFF
--- a/src/NuGet.Services.AzureSearch/Models/BaseMetadataDocument.cs
+++ b/src/NuGet.Services.AzureSearch/Models/BaseMetadataDocument.cs
@@ -1,0 +1,41 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Azure.Search;
+
+namespace NuGet.Services.AzureSearch
+{
+    /// <summary>
+    /// Implements most of <see cref="IBaseMetadataDocument"/>. Some fields are not defined here since they have
+    /// different Azure Search attributes.
+    /// </summary>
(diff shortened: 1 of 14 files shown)


joelverhagen:
I tried to only do a minimal subset of the Azure Search attributes (e.g. `[IsFilterable]`) so support our official API surface area and _not_ query parsing. For example, we will need to make more changes to support `owners:microsoft`. That is out of scope for this PR.

src/NuGet.Services.AzureSearch/Models/BaseMetadataDocument.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Azure.Search;
+
+namespace NuGet.Services.AzureSearch
+{
+    public abstract class BaseMetadataDocument : KeyedDocument
```
agr: Some documentation for the new classes?..

joelverhagen: Yeah, good idea.

PR #382, opened 2018-10-26
Add a job to fix packages' MD5 hash blob property

loic-sharma:
This creates a job that looks for blobs in the packages container that are missing the Content MD5 hash property. It downloads these blobs, hashes them, and updates their blob metadata.

⚠️ This creates  a new Octopus nupkg, `Ng.Operations`. This package contains Ng.exe without the Octopus PreDeploy.ps1/PostDeploy.ps1 scripts. This is necessary as the `Package2PackagesFixup` job shouldn't be installed as a service.

Part of https://github.com/NuGet/NuGetGallery/issues/6515


src/Catalog/FixPackageHashHandler.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            // Skip the package if it has a Content MD5 hash. Don't log to minimize noise.
+            if (blob.ContentMD5 != null)
+            {
+                return;
+            }
+
+            string hash;
+            using (var hashAlgorithm = MD5.Create())
+            using (var packageStream = await _httpClient.GetStreamAsync(blob.Uri))
```
loic-sharma: `FixPackageHashHandler` and `ValidatePackageHashHandler` both use `HttpClient` to download their packages as Azure Blob Storage's SDK hangs on package downloads. See: https://github.com/Azure/azure-storage-net/issues/470

src/Catalog/FixPackageHashHandler.cs
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        public async Task ProcessPackageAsync(CatalogIndexEntry packageEntry, ICloudBlockBlob blob)
+        {
+            await blob.FetchAttributesAsync(CancellationToken.None);
+
+            // Skip the package if it has a Content MD5 hash. Don't log to minimize noise.
+            if (blob.ContentMD5 != null)
+            {
+                return;
```
loic-sharma: I can't use `LogDebug` has Ng forces Debug logs to be enabled 😢

src/Catalog/FixPackageHashHandler.cs
```diff
@@ -0,0 +1,71 @@
...
+            using (var packageStream = await _httpClient.GetStreamAsync(blob.Uri))
+            {
+                var hashBytes = hashAlgorithm.ComputeHash(packageStream);
+
+                hash = Convert.ToBase64String(hashBytes);
+            }
+
+            blob.ContentMD5 = hash;
+
+            var condition = AccessCondition.GenerateIfMatchCondition(blob.ETag);
```
joelverhagen: :+1:

src/Catalog/PackagesContainerCatalogProcessor.cs
```diff
@@ -0,0 +1,157 @@
...
+                    e,
+                    "Could not process package {PackageId} {PackageVersion}",
+                    catalogEntry.Id,
+                    catalogEntry.Version);
+            }
+        }
+
+        private string BuildPackageFileName(CatalogIndexEntry packageEntry)
+        {
+            return $"{packageEntry.Id.ToLower()}.{packageEntry.Version.ToNormalizedString().ToLower()}.nupkg";
```
joelverhagen: >ToLower [](start = 38, length = 7)

`ToLowerInvariant`

src/Ng/Jobs/Catalog2PackageFixupJob.cs
```diff
@@ -0,0 +1,160 @@
...
+                handler = new ValidatePackageHashHandler(
+                    httpClient,
+                    TelemetryService,
+                    LoggerFactory.CreateLogger<ValidatePackageHashHandler>());
+            }
+            else
+            {
+                Logger.LogInformation("Ensuring all packages have a Content MD5 hash...");
+
+                handler = new FixPackageHashHandler(
```
joelverhagen: >handler [](start = 16, length = 7)

So if we want to clean up cache header or anything else, what would we do? enhance `FixPackageHashHandler` or add a new handler?

src/Ng/Jobs/Catalog2PackageFixupJob.cs
```diff
@@ -0,0 +1,160 @@
...
+
+            httpClient.Timeout = TimeSpan.FromMinutes(10);
+
+            // Prepare the handler that will run on each catalog entry.
+            IPackagesContainerHandler handler;
+            if (verify)
+            {
+                Logger.LogInformation("Validating that all packages have the proper Content MD5 hash...");
+
+                handler = new ValidatePackageHashHandler(
```
joelverhagen: Shouldn't DI be newing these things up?

src/Ng/Jobs/Catalog2PackageFixupJob.cs
```diff
@@ -0,0 +1,160 @@
...
+            // Prepare the Azure Blob Storage client.
+            var credentials = new StorageCredentials(storageAccount, storageKey);
+            var account = new CloudStorageAccount(credentials, useHttps: true);
+
+            var container = account
+                .CreateCloudBlobClient()
+                .GetContainerReference(storageContainer);
+
+            // Prepare the HTTP Client
+            var httpClient = new HttpClient(new WebRequestHandler
```
joelverhagen: >HttpClient [](start = 33, length = 10)

Can we have a user agent please?

src/Ng/Jobs/Catalog2PackageFixupJob.cs
loic-sharma: Either approach would work. I’d prefer a new handler to keep each handler simple.

src/Catalog/FixPackageHashHandler.cs
```diff
@@ -0,0 +1,71 @@
...
+            if (blob.ContentMD5 != null)
+            {
+                return;
+            }
+
+            string hash;
+            using (var hashAlgorithm = MD5.Create())
+            using (var packageStream = await _httpClient.GetStreamAsync(blob.Uri))
+            {
+                var hashBytes = hashAlgorithm.ComputeHash(packageStream);
```
joelverhagen: So AFAIK `HttpClient.GetStream` does not buffer. `ComputeHash` does not either right?

In other words, we don't ever have the whole package in memory, right?

loic-sharma: Yup, this doesn't buffer entire packages into memory. 

.NET docs does indeed state that `HttpClient.GetStreamAsync` does not buffer the response body (see [this doc](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.getstreamasync?view=netframework-4.7.2#remarks)). I couldn't find such a guarantee for `ComputeHash`, but the corefx implementation doesn't buffer either (see [this code](https://github.com/dotnet/corefx/blob/master/src/System.Security.Cryptography.Primitives/src/System/Security/Cryptography/HashAlgorithm.cs#L106)).

src/Catalog/PackagesContainerCatalogProcessor.cs
loic-sharma: Good catch. Fixed

src/Ng/Jobs/Catalog2PackageFixupJob.cs
loic-sharma: Added dependency injection.

src/Ng/Jobs/Catalog2PackageFixupJob.cs
loic-sharma: Added. These downloads are going straight to Azure Blob Storage, so these shouldn't affect download counts.

src/Ng/Jobs/Catalog2PackageFixupJob.cs
```diff
@@ -0,0 +1,183 @@
...
+                var assembly = Assembly.GetEntryAssembly();
+                var assemblyName = assembly.GetName().Name;
+                var assemblyVersion = assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? "0.0.0";
+
+                var httpClient = new HttpClient(new WebRequestHandler
+                {
+                    AllowPipelining = true
+                });
+
+                httpClient.DefaultRequestHeaders.Add("User-Agent", $"{assemblyName}/{assemblyVersion}");
```
scottbommarito: This is really cool. IMO we may want to use this elsewhere as well.

You may want to add a comment in here describing what we're doing, e.g. using the assembly information to set the User Agent.

src/Ng/Jobs/Catalog2PackageFixupJob.cs
```diff
@@ -0,0 +1,183 @@
...
+            // Prepare the catalog reader.
+            services.AddSingleton(p =>
+            {
+                var telemetryService = p.GetRequiredService<ITelemetryService>();
+                var httpMessageHandlerFactory = CommandHelpers.GetHttpMessageHandlerFactory(telemetryService, verbose);
+
+                return new CollectorHttpClient(httpMessageHandlerFactory());
+            });
+
+            services.AddTransient(p =>
```
scottbommarito: In Autofac you could simplify this by explicitly specifying the `new Uri(source)` parameter...can you do this with `ServiceCollection`?

src/Ng/Jobs/Catalog2PackageFixupJob.cs
```diff
@@ -0,0 +1,183 @@
...
+
+            await Task.WhenAll(tasks);
+
+            Logger.LogInformation(
+                "Processed {ProcessedCount} packages in {ProcessDuration}",
+                totalEntries,
+                stopwatch.Elapsed);
+        }
+
+        private async Task LogProgress(ConcurrentBag<CatalogIndexEntry> packageEntries)
```
scottbommarito: I love this--it's a very clever way to emit progress information.

src/Ng/Jobs/Catalog2PackageFixupJob.cs
loic-sharma: Ah, this is a good point. Usually in ASP.NET Core world you have a type that represents your config/options that you inject into your service. So my options are:

1. Bring in Autofac to configure this service. That seems a little overkill
2. I create a new option type to configure the `CatalogIndexReader`. This also seems a little overkill

Do you feel strongly about this? If so, I'll go with option 2. Otherwise I'll leave it as is :)

src/Catalog/FixPackageHashHandler.cs
agr: .net [reference source](https://referencesource.microsoft.com/#mscorlib/system/security/cryptography/hashalgorithm.cs,5b02a2a217146fcf) contains very similar code.

src/Ng/Jobs/Catalog2PackageFixupJob.cs
scottbommarito: Oh it was definitely a nitpick. This is fine. Was just curious about ASP.NET Core.

src/Ng/Jobs/Catalog2PackageFixupJob.cs
loic-sharma: We already use this code! I stole this from NuGet.Jobs 😄

loic-sharma: Added a comment describing what's happening.

src/Catalog/PackagesContainerCatalogProcessor.cs
```diff
@@ -0,0 +1,160 @@
...
+
+        private string BuildPackageFileName(CatalogIndexEntry packageEntry)
+        {
+            var packageId = packageEntry.Id.ToLowerInvariant();
+            var packageVersion = packageEntry.Version.ToNormalizedString().ToLowerInvariant();
+
+            return $"{packageId}.{packageVersion}.nupkg";
+        }
+
+        private bool IsRetryableException(Exception exception)
```
agr: `static`?

agr: (same for other private methods around)

src/Catalog/ValidatePackageHashHandler.cs
```diff
@@ -0,0 +1,70 @@
...
+            _httpClient = httpClient ?? throw new ArgumentNullException(nameof(httpClient));
+            _telemetryService = telemetryService ?? throw new ArgumentNullException(nameof(telemetryService));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public async Task ProcessPackageAsync(CatalogIndexEntry packageEntry, ICloudBlockBlob blob)
+        {
+            await blob.FetchAttributesAsync(CancellationToken.None);
+
+            if (blob.ContentMD5 == null)
```
agr: Can `ContentMD5` be an empty string? Wrong length? Invalid content?
Are we OK with it if it can?

src/Catalog/Persistence/AzureCloudBlockBlob.cs
```diff
@@ -15,6 +16,12 @@ public sealed class AzureCloudBlockBlob : ICloudBlockBlob
     {
         private readonly CloudBlockBlob _blob;
 
+        public string ContentMD5
```
agr: Since for some reason it is string (I understand that that's what library provides), we need some documentation about what is exactly in the string.
Usually, string representation of a hash is a hex string, but the code below uses Base64 and the Azure .NET API [documentation](https://docs.microsoft.com/en-us/dotnet/api/microsoft.windowsazure.storage.blob.blobproperties.contentmd5?view=azure-dotnet#Microsoft_WindowsAzure_Storage_Blob_BlobProperties_ContentMD5) does not cover that.

src/Catalog/ValidatePackageHashHandler.cs
loic-sharma: That's covered by line 59, and will be tracked by `TrackPackageHasIncorrectHash`

src/Catalog/PackagesContainerCatalogProcessor.cs
loic-sharma: Fixed

src/Catalog/Persistence/AzureCloudBlockBlob.cs
loic-sharma: Added comment

src/Catalog/FixPackageHashHandler.cs
loic-sharma: Good idea. Added!

src/Ng/Jobs/Catalog2PackageFixupJob.cs
```diff
@@ -0,0 +1,184 @@
...
+        private const int MaximumPackageProcessingAttempts = 5;
+        private static readonly TimeSpan MaximumPackageProcessingTime = TimeSpan.FromMinutes(10);
+        private static readonly TimeSpan DonePollingInterval = TimeSpan.FromMinutes(1);
+
+        private IServiceProvider _serviceProvider;
+
+        public Catalog2PackageFixupJob(ITelemetryService telemetryService, ILoggerFactory loggerFactory)
+            : base(telemetryService, loggerFactory)
+        {
+            ThreadPool.SetMinThreads(MaxDegreeOfParallelism, 4);
```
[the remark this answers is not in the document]
loic-sharma: I pulled this from Azure Blob Storage's tutorial on how to optimize performance for downloading large amounts of random data ([this article](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blob-scalable-app-download-files)). I didn't measure performance without the call to `ThreadPool.SetMinThreads`, but considering that tasks are doing CPU intensive operations (computing hashes), it seems likely that a large number of threads is necessary.

src/Catalog/PackagesContainerCatalogProcessor.cs
```diff
@@ -0,0 +1,160 @@
...
+                var blob = new AzureCloudBlockBlob(rawBlob);
+
+                for (int i = 0; i < MaximumPackageProcessingAttempts; i++)
+                {
+                    try
+                    {
+                        await _handler.ProcessPackageAsync(catalogEntry, blob);
+                        return;
+                    }
+                    catch (Exception e) when (IsRetryableException(e))
```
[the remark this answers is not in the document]
loic-sharma: For the most part, our v3 jobs call `HttpClient.GetAsync` and existing retry handlers like `RetryWithExponentialBackoff` reflect this. This is the first retry handler built for `HttpClient.GetStreamAsync`. I think we can make this generic once/if other jobs use `GetStreamAsync` as well.

PR #386, opened 2018-10-30
Add tests for CatalogIndexEntry

dtivel:
Progress on https://github.com/NuGet/NuGetGallery/issues/6565.

Before adding a new member to `CatalogIndexEntry`, I'm making the type immutable and adding unit tests.


src/Catalog/CatalogIndexEntry.cs
```diff
@@ -11,134 +11,111 @@
...
-        {
-            return x.CommitTimeStamp.CompareTo(y.CommitTimeStamp);
-        }
-    }
 
+            var uri = new Uri(token["@id"].ToString());
+            var type = token["@type"].ToString();
+            var commitId = token["commitId"].ToString();
+            var commitTimeStamp = DateTime.ParseExact(
+                token["commitTimeStamp"].ToString(),
```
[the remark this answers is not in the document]
dtivel: The reason for this change is that tests failed on my dev machine because my machine's current time zone is _not_ UTC.  All our build and job machines run in UTC, so this change is about ensuring consistency wherever the code is executed.  We've had similar time zone bugs before ([example 1](https://github.com/NuGet/NuGetGallery/issues/6187) and [example 2](https://github.com/NuGet/NuGetGallery/issues/6189#issuecomment-406408036)).

This issue wasn't caught until now because there simply weren't tests for `CatalogIndexEntry`.

It's one thing that API documentation doesn't specify the exact datetime format.  It's another thing to ensure that the implementation does not change in ways that are not well understood.  Feed2Catalog tests ([here](https://github.com/NuGet/NuGet.Services.Metadata/blob/c9f960202abeffe6be1977686d4024f31e42be91/tests/NgTests/Feed2CatalogTests.cs#L1106) and [here](https://github.com/NuGet/NuGet.Services.Metadata/blob/c9f960202abeffe6be1977686d4024f31e42be91/tests/NgTests/Feed2CatalogTests.cs#L1245)) verify that commit timestamps follow a specific format, and I expect those tests to fail if the format changes.

src/NuGet.Services.Metadata.Catalog.Monitoring/ValidationCollector.cs
```diff
@@ -43,7 +43,7 @@ protected override async Task ProcessSortedBatchAsync(CollectorHttpClient client
 
             _logger.LogInformation("Processing catalog entries for {PackageId} {PackageVersion}.", packageId, packageVersion);
 
-            var catalogEntries = sortedBatch.Value.Select(c => new CatalogIndexEntry(c));
+            var catalogEntries = sortedBatch.Value.Select(token => CatalogIndexEntry.Create(token));
```
agr: >token => CatalogIndexEntry.Create(token) [](start = 58, length = 40)

You can save on delegate creation and pass `CatalogIndexEntry.Create` directly as a delegate.

src/Catalog/CatalogIndexEntry.cs
```diff
@@ -11,134 +11,111 @@
...
-            Version = version;
-        }
+            Uri = uri ?? throw new ArgumentNullException(nameof(uri));
 
-        [JsonConstructor]
-        public CatalogIndexEntry()
-        {
-        }
+            if (string.IsNullOrEmpty(type))
+            {
```
agr: Is whitespace allowed as a value for `type`?

PR #387, opened 2018-10-30
Set MonitoringProcessor user agent correctly

scottbommarito:
https://github.com/NuGet/Engineering/issues/1813

This was slightly more complex than I expected because the lack of a `public` `set`.


src/Ng/Jobs/MonitoringProcessorJob.cs
```diff
@@ -71,6 +71,21 @@ protected override void Init(IDictionary<string, string> arguments, Cancellation
...
+
+        /// <remarks>
+        /// Unfortunately, we have to use reflection to set the user agent string as we'd like to.
+        /// This can be greatly simplified if <see cref="UserAgent"/> adds a public way to set <see cref="UserAgent.UserAgentString"/> by string.
+        /// </remarks>
+        private static void SetUserAgentString()
+        {
+            var userAgentString = UserAgentUtility.GetUserAgent();
+
+            typeof(UserAgent)
```
dtivel: How about using `UserAgent.SetUserAgentString(…)`.  Update `UserAgentStringBuilder` as necessary.

scottbommarito: Although I agree that improving the client APIs is the best end goal, this was intended to be a quick fix and I would rather not to do any additional work to update the client APIs because it will slow this work down substantially.

I think it would be best to file an issue on Home, refer to it in this code, and then update it when the issue is resolved. I don't think this code gains much from using `UserAgentStringBuilder` rather than reflection given how simple it is.

Another alternative is to use the existing `UserAgentStringBuilder` and [change the client name](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Protocol/UserAgentStringBuilder.cs#L27) but ignore the version that it adds to the string (which is probably `4.8.0`).

PR #391, opened 2018-11-06
Add a timeout to catalog2lucene's index commits

loic-sharma:
Fixes https://github.com/NuGet/Engineering/issues/1838

Verify on DEV with:

```
traces
| where timestamp >= todatetime('2018-11-06T21:04:07.1397859Z') 
| where customDimensions.JobName == "catalog2lucene"
| order by timestamp asc
```


src/Ng/SearchIndexFromCatalogCollector.cs
```diff
@@ -101,13 +105,48 @@ public void EnsureCommitted()
...
+                {
+                    commitTask = commitTask.TimeoutAfter(_commitTimeout.Value);
+                }
+
+                await commitTask;
+            }
+            catch (OperationCanceledException)
+            {
+                _telemetryService.TrackIndexCommitTimeout();
+                _logger.LogError("TIMEOUT Commit timed out!");
```
chenriksson: Is there any other info we want to log here, like `NumDocs()`?

tests/NgTests/SearchIndexFromCatalogCollectorTests.cs
```diff
@@ -102,6 +111,63 @@ public async Task DoesNotSkipPackagesWhenExceptionOccurs(string catalogUri, stri
...
+            await mockServer.AddStorageAsync(catalogStorage);
+
+            ReadWriteCursor front = new DurableCursor(
+                storage.ResolveUri("cursor.json"),
+                storage,
+                MemoryCursor.MinValue);
+            ReadCursor back = MemoryCursor.CreateMax();
+
+            var commitTimeout = TimeSpan.FromSeconds(1);
+            var stuckDuration = TimeSpan.FromMinutes(1);
```
chenriksson: Adds a minute to tests, or are they run in parallel? Could this be shorter?

loic-sharma: The timeout is set to a second, so this test completes after a second.

src/Ng/SearchIndexFromCatalogCollector.cs
loic-sharma: Good idea. Added

PR #394, opened 2018-11-08
Add IndexActionBuilder to construct IndexActions per package registration

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/6440
Address https://github.com/NuGet/NuGetGallery/issues/5435

### DIFF
--- a/src/Catalog/Helpers/Utils.cs
+++ b/src/Catalog/Helpers/Utils.cs
@@ -35,6 +35,19 @@ namespace NuGet.Services.Metadata.Catalog
 
         private static readonly ConcurrentDictionary<string, string> _resourceStringCache = new ConcurrentDictionary<string, string>();
 
+        private static readonly char[] TagTrimChars = { ',', ' ', '\t', '|', ';' };
+
+        public static string[] SplitTags(string original)
+        {
+            var fields = original
+                .Split(TagTrimChars)
+                .Select(w => w.Trim(TagTrimChars))
+                .Where(w => w.Length > 0)
+                .ToArray();
+
+            return fields;
+        }
+
         public static Stream GetResourceStream(string resourceName)
         {
             if (string.IsNullOrEmpty(resourceName))
--- a/src/NuGet.Services.AzureSearch/IndexActionBuilder.cs
+++ b/src/NuGet.Services.AzureSearch/IndexActionBuilder.cs
@@ -1,0 +1,215 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Web;
+using Microsoft.Azure.Search.Models;
+using NuGet.Services.AzureSearch.Db2AzureSearch;
+using NuGet.Services.Metadata.Catalog;
+using NuGet.Versioning;
+using NuGetGallery;
+
+namespace NuGet.Services.AzureSearch
+{
+    public class IndexActionBuilder : IIndexActionBuilder
+    {
+        private static readonly DateTimeOffset UnlistedPublished = new DateTimeOffset(1900, 1, 1, 0, 0, 0, TimeSpan.Zero);
+
+        public SearchAndHijackIndexActions AddNewPackageRegistration(NewPackageRegistration packageRegistration)
+        {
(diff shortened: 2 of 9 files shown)


src/NuGet.Services.AzureSearch/IndexActionBuilder.cs
```diff
@@ -0,0 +1,213 @@
...
+            var key = GetSearchDocumentKey(packageRegistration.PackageId, searchFilters);
+
+            if (changeType == SearchIndexChangeType.Delete)
+            {
+                return IndexAction.Delete(new KeyedDocument { Key = key });
+            }
+
+            if (changeType != SearchIndexChangeType.AddFirst)
+            {
+                throw new NotImplementedException();
```
joelverhagen: This will be expanded later for `catalog2azuresearch`.

src/NuGet.Services.AzureSearch/IndexActionBuilder.cs
```diff
@@ -0,0 +1,213 @@
...
+            var latest = versionLists.GetLatestVersionInfoOrNull(searchFilters);
+            var package = versionToPackage[latest.ParsedVersion];
+
+            var metadata = new SearchDocument.Full();
+            ApplyMetadata(metadata, packageRegistration.PackageId, package);
+
+            metadata.Key = key;
+            metadata.FullVersion = latest.FullVersion;
+            metadata.TotalDownloadCount = packageRegistration.TotalDownloadCount;
+            metadata.Owners = packageRegistration
```
joelverhagen: Progress on https://github.com/NuGet/NuGetGallery/issues/6258.

src/Catalog/Helpers/Utils.cs
```diff
@@ -35,6 +35,19 @@ public static class Utils
 
         private static readonly ConcurrentDictionary<string, string> _resourceStringCache = new ConcurrentDictionary<string, string>();
 
+        private static readonly char[] TagTrimChars = { ',', ' ', '\t', '|', ';' };
+
+        public static string[] SplitTags(string original)
```
joelverhagen: This is so feed2catalog and db2azuresearch use the same tag splitting and normalizing logic.

src/NuGet.Services.AzureSearch/IndexActionBuilder.cs
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        private static DateTimeOffset AssumeUtc(DateTime dateTime)
+        {
+            return new DateTimeOffset(
+                DateTime.SpecifyKind(dateTime, DateTimeKind.Utc),
+                TimeSpan.Zero);
+        }
+
+        private static string GetSearchDocumentKey(string packageId, SearchFilters searchFilters)
```
joelverhagen: I made these keys a little longer than what is minimally required so that they are a bit readable. Take a look at the test cases below for examples.

In general, I combine a "readable" string, which is not necessarily unique due to character restrictions, and a "unique" string, which is unique via base64 encoding of a unique string.

tests/NuGet.Services.AzureSearch.Tests/IndexActionBuilderFacts.cs
```diff
@@ -0,0 +1,455 @@
...
+                new SearchCredentials("unit-test-api-key"),
+                testHandler))
+            {
+                var indexClient = serviceClient.Indexes.GetClient("unit-test-index");
+                await indexClient.Documents.IndexAsync(IndexBatch.Upload(new[] { obj }));
+                return testHandler.LastRequestBody;
+            }
+        }
+
+        private class TestHttpClientHandler : HttpClientHandler
```
joelverhagen: The encoding APIs used by Azure Search to take a document and serialize it as JSON before sending it to the REST APIs is private so I made this test HTTP client handler to capture the POSTed JSON.

src/Catalog/Helpers/Utils.cs
```diff
@@ -35,6 +35,19 @@ public static class Utils
...
+        {
+            var fields = original
+                .Split(TagTrimChars)
+                .Select(w => w.Trim(TagTrimChars))
+                .Where(w => w.Length > 0)
```
agr: Wouldn't `original.Split(TagTrimChars, StringSplitOptions.RemoveEmptyEntries)` do the same as the three lines above?

joelverhagen: No, because the `Trim` happens after the split.

PR #398, opened 2018-11-09
[HOTFIX] Force process termination on index commit timeouts

loic-sharma:
Previously, a `OperationCanceledException` was thrown if a commit to the index timed out. While this correctly caused the process to end, it would wait until all `using`s were properly `Dispose()`d. Unfortunately, `IndexWriter.Dispose()` hangs and prevented the shutdown from completing.

This hotfix forces the process to end without allowing `Dispose()` methods to run.


joelverhagen:
"untestable" 😄

PR #400, opened 2018-11-10
 Write the version lists to blob storage in db2azuresearch

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/6440

### DIFF
--- a/src/NuGet.Services.AzureSearch/VersionList/IVersionListDataClient.cs
+++ b/src/NuGet.Services.AzureSearch/VersionList/IVersionListDataClient.cs
@@ -1,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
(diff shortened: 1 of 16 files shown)


src/NuGet.Services.AzureSearch/VersionList/IVersionListDataClient.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using NuGetGallery;
+
+namespace NuGet.Services.AzureSearch
+{
+    public interface IVersionListDataClient
```
loic-sharma: Do we have any documentation on the version list resource anywhere? Considering this is an "internal" resource with no public documentation, it'd be nice to have a quick explanation somewhere.

joelverhagen: Good point. It's in the OneNote but I'll add code comments

PR #401, opened 2018-11-13
V3:  extract parallelized commit processing into common utility class

dtivel:
Progress on https://github.com/NuGet/Engineering/issues/1843.

At a high level this change aims to do 3 things:

* extract existing parallelized commit processing functionality into a common utility class so it can be leveraged by Catalog2Dnx and Joel's work on search
* improve the implementation
* add tests for both new and existing code

Here is the general approach for parallelizing catalog commits in jobs like Catalog2Registration and Catalog2Dnx:

Where `P` represents a package ID and `T` a commit timestamp, suppose we have the following catalog commits for fixed timestamp range `T₀-T₃`:
```
          P₀   P₁   P₂   P₃
   ^      ------------------
   |      T₃   T₃
   |                     T₂
   |      T₁        T₁
  time    T₀   T₀
```

Each column contains all catalog commits for its corresponding package ID (not identity).

Each column is represented by a `CatalogCommitBatchTask` instance.  Each group of columns sharing the same minimum commit timestamp is represented by a `CatalogCommitBatchTasks` instance.  In the above example, there will be 3 `CatalogCommitBatchTasks` instances:

1.  an instance for `P₀` and `P₁`, since they share the same minimum commit timestamp `T₀`
2.  an instance for `P₂`
3.  an instance for `P₃`

`RegistrationCollector` (and presumably `DnxCatalogCollector` in the future) processes `maxConcurrentBatches` columns in parallel.  The default is 10.

Only when all processing for a `CatalogCommitBatchTasks` instance completes successfully will a `CommitCollector` update its cursor.  In the above example, the cursor will be updated as follows:

1.  When both `P₀-T₀` and `P₁-T₀` succeed, update the cursor to `T₀`.
2.  When `P₂-T₁` succeeds, update the cursor to `T₁`.  Successful completion of `P₀-T₁` is implied by completion of the previous step.
3.  When `P₃-T₂` succeeds, update the cursor to `T₃`.  Successful completion of `P₀-T₁` and `P₂-T₁` is implied by completion of the previous step.  Note that since this is the last work item for the given time range, the cursor is not updated to `T₂` but instead the maximum commit timestamp for the entire commit timestamp range.

This PR adds 2 improvements over the existing implementation:

1.  Because updating the cursor is costly, updating the cursor to the latest successfully committed timestamp instead of the current commit timestamp.  The above example already demonstrates this in step 3 by updating the cursor to `T₃` instead of first `T₂` and then `T₃`.
2.  Only the most recent catalog commit per package identity needs to be processed; older commits are skipped.  In the above example, `P₀-T₀`, `P₀-T₁`, and `P₁-T₀` would be skipped.


src/Catalog/CatalogCommitBatchTask.cs
```diff
@@ -0,0 +1,54 @@
...
+using System;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.Metadata.Catalog
+{
+    /// <summary>
+    /// Represents an asynchrononous task associated with catalog changes for a specific commit item key
+    /// and potentially spanning multiple commits.
+    /// </summary>
+    public sealed class CatalogCommitBatchTask
```
joelverhagen: Nit: implement IEquatable<CatalogCommitBatchTask>

src/Catalog/CatalogCommitUtilities.cs
```diff
@@ -0,0 +1,457 @@
...
+                processingBatches.Enqueue(batchTask);
+            }
+        }
+
+        internal static async Task<bool> FetchAsync(
+            CollectorHttpClient client,
+            ReadWriteCursor front,
+            ReadCursor back,
+            FetchCatalogCommitsAsync fetchCatalogCommitsAsync,
+            GetCatalogCommitItemKey getCatalogCommitItemKey,
```
joelverhagen: This seems to be unused

src/Catalog/CatalogCommitUtilities.cs
```diff
@@ -0,0 +1,457 @@
...
+            {
+                throw new ArgumentNullException(nameof(getCatalogCommitItemKey));
+            }
+
+            var catalogItemsGroups = catalogItems
+                .GroupBy(catalogItem => getCatalogCommitItemKey(catalogItem));
+
+            foreach (var catalogItemsGroup in catalogItemsGroups)
+            {
+                var key = catalogItemsGroup.Key;
```
joelverhagen: >catalogItemsGroup [](start = 26, length = 17)

nit: seems like a pointless local

src/Catalog/CatalogCommitUtilities.cs
```diff
@@ -0,0 +1,457 @@
...
+
+                // Before filtering out all but the latest commit for each package identity, determine the earliest
+                // commit timestamp for all items in this batch.  This timestamp is important for processing commits
```
joelverhagen: >timestamp [](start = 71, length = 9)

I don't understand why the min *before filtering* is important. Let's talk offline if it's complex.

src/Catalog/CatalogCommitUtilities.cs
```diff
@@ -0,0 +1,457 @@
...
+                        processingBatches,
+                        lastBatch,
+                        maxConcurrentBatches,
+                        processCommitItemBatchAsync,
+                        cancellationToken);
+                }
+
+                if (hasAnyBatchFailed)
+                {
+                    var innerException = exceptions.Count == 1 ? exceptions.Single() : new AggregateException(exceptions);
```
joelverhagen: >innerException [](start = 24, length = 14)

Will Application Insights have record off ALL of these exceptions?

src/Catalog/ProcessCommitItemBatchAsync.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+using Newtonsoft.Json.Linq;
+
+namespace NuGet.Services.Metadata.Catalog
+{
+    public delegate Task ProcessCommitItemBatchAsync(
```
joelverhagen: >ProcessCommitItemBatchAsync [](start = 25, length = 27)

nit: perhaps these tiny delegate files can be rolled into CatalogCommitUtilities at the top

src/Catalog/CatalogCommitUtilities.cs
dtivel: In my V3 performance work I have striven to preserve existing assumptions and behaviors as much as possible, no matter how small or obscure.  Before my commit parallelization, Catalog2Registration would update its cursor only when all commits up to the present commit were fully and successfully processed.  I aimed to preserve that behavior even with commit parallelization.  The minimum commit timestamp before filtering enables this.

I'm open to changing the behavior to completely ignore older commits and their timestamps, but I think it should be separate from this PR.

src/Catalog/ProcessCommitItemBatchAsync.cs
dtivel: I kept one type per file.  `CatalogCommitUtilities` is already fairly large.

src/Catalog/CatalogCommit.cs
```diff
@@ -0,0 +1,41 @@
...
+            CommitTimeStamp = commitTimeStamp;
+            Uri = uri;
+        }
+
+        public DateTime CommitTimeStamp { get; }
+        public Uri Uri { get; }
+
+        public int CompareTo(object obj)
+        {
+            return CommitTimeStamp.CompareTo(((CatalogCommit)obj).CommitTimeStamp);
```
agr: Is it OK to have the possibility of throwing `InvalidCastException` here?

src/Catalog/CatalogCommitBatchTask.cs
```diff
@@ -0,0 +1,54 @@
...
+        public override bool Equals(object obj)
+        {
+            var other = obj as CatalogCommitBatchTask;
+
+            if (ReferenceEquals(other, null))
+            {
+                return false;
+            }
+
+            return GetHashCode() == other.GetHashCode();
```
agr: Shouldn't we compare keys when hashes are the same?

src/Catalog/CatalogCommitItem.cs
```diff
@@ -0,0 +1,99 @@
...
+        public Uri Uri { get; }
+        public DateTime CommitTimeStamp { get; }
+        public string CommitId { get; }
+        public PackageIdentity PackageIdentity { get; }
+        public IReadOnlyList<string> Types { get; }
+        public IReadOnlyList<Uri> TypeUris { get; }
+
+        public int CompareTo(object obj)
+        {
+            return CommitTimeStamp.CompareTo(((CatalogCommitItem)obj).CommitTimeStamp);
```
agr: Same comment as above. Public methods should validate their arguments before attempting such casts.

src/Catalog/CatalogCommitItemBatch.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// <exception cref="ArgumentException">Thrown if <paramref name="items" /> is either <c>null</c> or empty.</exception>
+        public CatalogCommitItemBatch(DateTime commitTimeStamp, string key, IEnumerable<CatalogCommitItem> items)
+        {
+            if (items == null || !items.Any())
+            {
+                throw new ArgumentException(Strings.ArgumentMustNotBeNullOrEmpty, nameof(items));
+            }
+
+            CommitTimeStamp = commitTimeStamp;
+            Key = key;
```
agr: Null check for `key`?

dtivel: No, see [documentation](https://github.com/NuGet/NuGet.Services.Metadata/pull/401/files#diff-6bb356028c5b5df8e93a83cb6fe26c68R22).

src/Catalog/CatalogCommitBatchTask.cs
dtivel: Good catch!

src/Catalog/CatalogCommitUtilities.cs
dtivel: No, but I addressed this with a subsequent [commit](https://github.com/NuGet/NuGet.Services.Metadata/pull/401/commits/141ba285117dd9b043592d26eebcb6a97faad60c).

PR #402, opened 2018-11-13
Introduce some minor refactoring to prepare for catalog2azuresearch

joelverhagen:
Pull in more test infra, rename things to be more generic, etc. The goal here is to do "noisy" changes all at once and leave the meat of the work to future PRs.


src/NuGet.Services.AzureSearch/ICommand.cs
```diff
@@ -3,9 +3,9 @@
 
 using System.Threading.Tasks;
 
-namespace NuGet.Services.AzureSearch.Db2AzureSearch
+namespace NuGet.Services.AzureSearch
 {
-    public interface IDb2AzureSearchCommand
+    public interface ICommand
```
skofman1: this looks job infra related. Is there a reason to use a different model here then we use for other jobs?

joelverhagen: This is called by a `Job`. It's so the whole thing can be activated by DI.

joelverhagen: Spoke offline. Will remove this unnecessary interface.

PR #403, opened 2018-11-13
Extract batch pushing to its own class to share with catalog2azuresearch

joelverhagen:

This PR introduces `BatchPusher`. This is a **stateful** class that handles pushing changes to the two Azure Search indexes and updating the version list for a package ID that is fully pushed to Azure Search.

This class has three methods:

1. `EnqueueIndexActions` - does no hard work, just puts the search and hijack index actions in their respective queues and keeps track of the version list to push.
1. `PushFullBatchesAsync` - pushes full batches to Azure search (1000 index actions). It updates hijack index first then updates the search index. If a package ID has all of its index changes pushed to Azure Search, then the version list should be updated too.
1. `FinishAsync` - same as `PushFullBatchesAsync` but at the end allows the last partial batch to be pushed.

The trick of the implementation is a reference count pattern on package ID. As index actions are enqueued for the package ID, the reference count goes up. As the index actions are pushed to Azure Search, the reference count goes down. When the reference count reaches zero, the version list for that package ID is updated.

### DIFF
--- a/src/NuGet.Services.AzureSearch/BatchPusher.cs
+++ b/src/NuGet.Services.AzureSearch/BatchPusher.cs
@@ -1,0 +1,257 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Azure.Search.Models;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using NuGet.Services.AzureSearch.Wrappers;
+
+namespace NuGet.Services.AzureSearch
+{
+    public class BatchPusher : IBatchPusher
+    {
+        private readonly ISearchIndexClientWrapper _searchIndexClient;
+        private readonly ISearchIndexClientWrapper _hijackIndexClient;
+        private readonly IVersionListDataClient _versionListDataClient;
+        private readonly IOptionsSnapshot<AzureSearchConfiguration> _options;
+        private readonly ILogger<BatchPusher> _logger;
+        internal readonly Dictionary<string, int> _idReferenceCount;
+        internal readonly Queue<IdAndValue<IndexAction<KeyedDocument>>> _searchActions;
+        internal readonly Queue<IdAndValue<IndexAction<KeyedDocument>>> _hijackActions;
+        internal readonly Dictionary<string, ResultAndAccessCondition<VersionListData>> _versionListDataResults;
+
+        public BatchPusher(
+            ISearchIndexClientWrapper searchIndexClient,
+            ISearchIndexClientWrapper hijackIndexClient,
+            IVersionListDataClient versionListDataClient,
+            IOptionsSnapshot<AzureSearchConfiguration> options,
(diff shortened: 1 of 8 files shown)


src/NuGet.Services.AzureSearch/Db2AzureSearch/Db2AzureSearchCommand.cs
```diff
@@ -20,28 +19,22 @@ public class Db2AzureSearchCommand
...
         private readonly ILogger<Db2AzureSearchCommand> _logger;
 
         public Db2AzureSearchCommand(
             INewPackageRegistrationProducer producer,
             IIndexActionBuilder indexActionBuilder,
             ISearchServiceClientWrapper serviceClient,
-            ISearchIndexClientWrapper searchIndexClient,
-            ISearchIndexClientWrapper hijackIndexClient,
-            IVersionListDataClient versionListDataClient,
+            Func<IBatchPusher> batchPusherFactory,
```
loic-sharma: >Func<IBatchPusher> batchPusherFactory [](start = 12, length = 37)

This seems to be missing from the DI container

joelverhagen: `Func<T>` comes for free when you register `T`.

joelverhagen: ... which means you have to register `T`. 😆

loic-sharma: Whaaaaat?! I wish GitHub had a "mind blown" emoji

src/NuGet.Services.AzureSearch/BatchPusher.cs
```diff
@@ -0,0 +1,257 @@
...
+                    {
+                        _logger.LogDebug("Updating version list for package ID {PackageId}.", finished.Id);
+                        await _versionListDataClient.ReplaceAsync(
+                            finished.Id,
+                            finished.Value.Result,
+                            finished.Value.AccessCondition);
+                    }
+
+                    stopwatch.Stop();
+                    _logger.LogInformation(
```
loic-sharma: Could we make this a metric?

src/NuGet.Services.AzureSearch/BatchPusher.cs
```diff
@@ -0,0 +1,257 @@
...
+            {
+                return;
+            }
+
+            if (batch.Count > _options.Value.AzureSearchBatchSize)
+            {
+                throw new ArgumentException("The provided batch is too large.");
+            }
+
+            _logger.LogInformation(
```
loic-sharma: Could we add a metric that tracks the duration of the indexing operation, which index was updated, and whether the operaiton was successful? This would be useful for monitoring

src/NuGet.Services.AzureSearch/BatchPusher.cs
joelverhagen: Tracked here: https://github.com/NuGet/NuGetGallery/issues/6664

src/NuGet.Services.AzureSearch/BatchPusher.cs
joelverhagen: Tracked here: https://github.com/NuGet/NuGetGallery/issues/6664

PR #405, opened 2018-11-14
Copy NuGet.Protocol.Catalog from Samples repository

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/6666.

This is just a simple copy from [Samples repository](https://github.com/NuGet/Samples/tree/master/CatalogReaderExample). I had to move from SDK-based project to legacy .csproj which means it no longer targets .NET Standard 1.3 😢. I did not want to introduce SDK-based projects are part of this effort.

This code is already used by package lag monitor.


src/NuGet.Protocol.Catalog/NuGet.Protocol.Catalog.csproj
```diff
@@ -0,0 +1,98 @@
...
+  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+  <PropertyGroup>
+    <SignPath>..\..\build</SignPath>
+    <SignPath Condition="'$(BUILD_SOURCESDIRECTORY)' != ''">$(BUILD_SOURCESDIRECTORY)\build</SignPath>
+    <SignPath Condition="'$(NuGetBuildPath)' != ''">$(NuGetBuildPath)</SignPath>
+    <SignType Condition="'$(SignType)' == ''">none</SignType>
+  </PropertyGroup>
+  <Import Project="$(SignPath)\sign.targets" Condition="Exists('$(SignPath)\sign.targets')" />
+  <Import Project="$(SignPath)\sign.microbuild.targets" Condition="Exists('$(SignPath)\sign.microbuild.targets')" />
+  <Import Project="..\..\sign.thirdparty.targets" />
```
scottbommarito: I would double check that all of the dependencies of this project are in the third-party list. May want to kick off a release build on this feature branch just to make sure that `Extract and Verify Artifacts` succeeds.

joelverhagen: Good idea.

joelverhagen: I forgot MicroBuild. Should be good now.

PR #410, opened 2018-11-16
V3 Monitoring - Improve logging when queue message processing fails

scottbommarito:
Added this when trying to debug why https://github.com/NuGet/NuGetGallery/issues/6584 failed to deploy

If a validation fails to run, and its status fails to be saved, we don't properly log the exceptions. This change improves the logging so that when the job fails to process a queue message, we log everything to AI.


src/NuGet.Services.Metadata.Catalog.Monitoring/Utility/LogEvents.cs
```diff
@@ -7,6 +7,8 @@ namespace NuGet.Services.Metadata.Catalog.Monitoring
 {
     public static class LogEvents
     {
+        public static EventId QueueMessageFatalFailure = new EventId(905, "Failed to process queue message");
```
dtivel: Order this line below by event ID?  Then it's easier to see what the next unused ID is.

PR #412, opened 2018-11-20
Catalog2Dnx:  parallelize commit processing

dtivel:
Progress on https://github.com/NuGet/Engineering/issues/1843.

In the DEV environment with a China storage destination I observed a processing rate of 7,401 packages/hour for 1,998 packages with unique package ID's and 9,164 packages/hour for 2,000 packages with the same package ID.  With a U.S. storage destination, Catalog2Dnx has processing rates well over 30,000 packages/hour.

Each test run consisted of packages following production package size and package entry count distributions to the nearest percentile.

This PR also adds package sizes to customMetrics telemetry.  It's interesting data but otherwise hard to correlate across different data sources.

@loic-sharma, @skofman1, @joelverhagen


src/Catalog/Dnx/DnxCatalogCollector.cs
```diff
@@ -255,6 +298,8 @@ private async Task<bool> AreRequiredPropertiesPresentAsync(Storage destinationSt
                 // If the ETag's differ, we'll fall back to using a single HTTP GET request.
                 var token1 = await _sourceStorage.GetOptimisticConcurrencyControlTokenAsync(sourceUri, cancellationToken);
 
+                telemetryProperties[TelemetryConstants.SizeInBytes] = sourceBlob.Length.ToString();
+
```
loic-sharma: Should `ProcessPackageDetailsViaHttpAsync` also add this property?

dtivel: Yes, somehow I removed it.  Adding it back.

src/Ng/Jobs/NgJob.cs
```diff
@@ -17,7 +17,7 @@ public abstract class NgJob
         protected ILoggerFactory LoggerFactory;
         protected ILogger Logger;
 
-        protected int MaxDegreeOfParallelism { get; }
+        protected int MaxDegreeOfParallelism { get; set; } = 8;
```
loic-sharma: >MaxDegreeOfParallelism [](start = 22, length = 22)

This seems like a risky change. Does this affect feed2catalog/catalog2registration/catalog2lucene?

Please update catalog2packagefixup. See: [https://github.com/NuGet/NuGet.Services.Metadata/blob/master/src/Ng/Jobs/Catalog2PackageFixupJob.cs#L144](https://github.com/NuGet/NuGet.Services.Metadata/blob/master/src/Ng/Jobs/Catalog2PackageFixupJob.cs#L144)

dtivel: Good catch.  This was unintentional.

PR #414, opened 2018-11-22
Refactoring: moved FeedHelpers.GetCatalogPropertiesAsync

xavierdecoster:
This is now `CatalogProperties.ReadAsync` (as it has nothing to do with a feed).

Noticed this when exploring `db2catalog`. This refactoring will help keep upcoming PRs smaller and more focused.


tests/CatalogTests/Helpers/CatalogPropertiesTests.cs
```diff
@@ -31,5 +36,196 @@ public void Constructor_InitializesPropertiesWithNonNullValues()
...
+            var catalogProperties = await CatalogProperties.ReadAsync(
+                storage.Object,
+                Mock.Of<ITelemetryService>(),
+                CancellationToken.None);
+
+            Assert.NotNull(catalogProperties);
+            Assert.NotNull(catalogProperties.LastCreated);
+            Assert.NotNull(catalogProperties.LastDeleted);
+            Assert.NotNull(catalogProperties.LastEdited);
+            Assert.Equal(datetimeOffset, catalogProperties.LastCreated.Value);
```
joelverhagen: different timestamp values to be used to verify that they aren't getting mixed up

joelverhagen: Well, considering these are moved, feel free to ignore. But now's a good as time as any IMO.

xavierdecoster: Assuming this is what you meant, I think I handled it as part of https://github.com/NuGet/NuGet.Services.Metadata/pull/414/commits/c9eea547a693d526e0bea2053a9240a503d5b10e

PR #415, opened 2018-11-22
Removed unused ng.exe CLI options from feed2catalog and catalog2dnx jobs

xavierdecoster:
Spotted these unused commandline arguments. 
No point in displaying them in command help if they're not used anywhere :)

So either we remove them, or we use them. If we want the latter, please explain their purpose @dtivel @joelverhagen :)


loic-sharma:
These are used by @dtivel’s performance work. The V3 Jobs prefer blog storage to do tricks like server side copy, or, downloading just a nuspec from a nupkg blob.

loic-sharma:
I don’t see how hiding arguments from `GetUsage` is an improvement.

xavierdecoster:
@loic-sharma because these arguments are not used anywhere in code. Then what's the point in having them? Why displaying arguments if they're not used by the job?

scottbommarito:
@xavierdecoster - These arguments ARE used. This PR should be closed.

For example: https://github.com/NuGet/NuGet.Services.Metadata/blob/master/src/Ng/Jobs/Feed2CatalogJob.cs#L86

Although `storageAccountNamePreferredPackageSourceStorage` isn't referenced explicitly in code, it is constructed automatically using the hardcoded string at https://github.com/NuGet/NuGet.Services.Metadata/blob/master/src/Ng/CommandHelpers.cs#L204.

The pattern is definitely confusing, but it will become a lot less confusing when we move `Ng` to JSON configuration, so we shouldn't take any action here right now.

xavierdecoster:
Ok, closing, but this is really confusing. I'd prefer explicit over implicit code references to avoid magic. When building `db2catalog`, we'll be touching this code base and I'd like to avoid duplicating magic or making mistakes. Preferably, this also happens before introducing `db2catalog` so those PRs are not polluted by such changes.

scottbommarito:
@xavierdecoster - I hate the "magic" too, but I think it should be done as part of a larger infrastructure change (and not as part of `db2catalog`) because moving these jobs to a proper configuration system is no small undertaking. I think any solution to the configuration problem besides moving to JSON is just a temporary hack.

xavierdecoster:
@scottbommarito I think @chenriksson was working on moving some jobs to JSON config. 

Even though this particular job can be migrated over independently of `db2catalog`, it would be good to have this one done first imho. No additional magic/hacks please! 😄

PR #417, opened 2018-11-26
 Enhance NuGet.Protocol.Catalog to support other client implementations

joelverhagen:

`ISimpleHttpClient` will be used by a future `RegistrationClient` to read registration blobs.

Also, add `IsSemVer2` extension method to package details catalog leaves.

### DIFF
--- a/src/NuGet.Protocol.Catalog/Models/ModelExtensions.cs
+++ b/src/NuGet.Protocol.Catalog/Models/ModelExtensions.cs
@@ -151,7 +151,7 @@ namespace NuGet.Protocol.Catalog
         }
 
         /// <summary>
-        /// Determines if the provided package details list represents a listed package.
+        /// Determines if the provided package details leaf represents a listed package.
         /// </summary>
         /// <param name="leaf">The catalog leaf.</param>
         /// <returns>True if the package is listed.</returns>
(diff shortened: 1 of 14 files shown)


src/NuGet.Protocol.Catalog/SimpleHttpClient.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.Net.Http;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+namespace NuGet.Protocol.Catalog
+{
+    public class SimpleHttpClient : ISimpleHttpClient
+    {
+        private static readonly JsonSerializer JsonSerializer = CatalogJsonSerialization.Serializer;
```
loic-sharma: >JsonSerializer [](start = 47, length = 14)

If it wasn't for this `JsonSerializer`, `SimpleHttpClient` wouldn't be catalog specific. Instead of having a static property, could we instead accept a `JsonSerializer` in the constructor? What are the tradeoffs?

src/NuGet.Protocol.Catalog/SimpleHttpClient.cs
```diff
@@ -0,0 +1,80 @@
...
+        private readonly HttpClient _httpClient;
+        private readonly ILogger _logger;
+
+        public SimpleHttpClient(HttpClient httpClient, ILogger<SimpleHttpClient> logger)
+            : this(httpClient, (ILogger)logger)
+        {
+        }
+
+        internal SimpleHttpClient(HttpClient httpClient, ILogger logger)
```
loic-sharma: >SimpleHttpClient [](start = 17, length = 16)

What's this constructor for?

joelverhagen: Catalog client's existing ctor (to avoid API breaking changes).

src/NuGet.Protocol.Catalog/SimpleHttpClient.cs
joelverhagen: I was favoring a simple client that should work on all V3 JSON. I think the name of the static class `CatalogJsonSerialization` is what is confusing.

Perhaps renaming it to `NuGetJsonSerialization` would clear this up?

src/NuGet.Protocol.Catalog/SimpleHttpClient.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.IO;
+using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+namespace NuGet.Protocol.Catalog
+{
+    public class SimpleHttpClient : ISimpleHttpClient
```
chenriksson: nit: why not `JsonHttpClient` or similar to reflect the JSON deserialization that the class does?

src/NuGet.Protocol.Catalog/Models/ModelExtensions.cs
```diff
@@ -166,5 +166,47 @@ public static bool IsListed(this PackageDetailsCatalogLeaf leaf)
...
             return leaf.Published.Year != 1900;
         }
+
+        /// <summary>
+        /// Determines if the provied package details leaf represents a SemVer 2.0.0 package. A package is considered
+        /// SemVer 2.0.0 if it's version is SemVer 2.0.0 or one of its dependency version ranges is SemVer 2.0.0.
+        /// </summary>
+        /// <param name="leaf">The catalog leaf.</param>
+        /// <returns>True if the package is SemVer 2.0.0.</returns>
+        public static bool IsSemVer2(this PackageDetailsCatalogLeaf leaf)
```
chenriksson: Assume this will be used in a separate PR, for bug fix?

joelverhagen: Separate PR.

src/NuGet.Protocol.Catalog/SimpleHttpClient.cs
joelverhagen: My hope is that other convenience methods could be added here as well. For example, fetching the .nuspec as `XDocument` or a .nupkg as a local temp file. `Json` is just the first convenience needed.

PR #420, opened 2018-11-28
Monitoring:  move caching into ValidationContext

dtivel:
Progress on https://github.com/NuGet/NuGetGallery/issues/6565.

Apply late feedback from https://github.com/NuGet/NuGet.Services.Metadata/pull/342#discussion_r211786011.


src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/ValidationContext.cs
```diff
@@ -63,34 +71,47 @@ public class ValidationContext
...
-            Lazy<Task<PackageRegistrationLeafMetadata>> lazyTask)
+        public Task<PackageRegistrationIndexMetadata> GetIndexV2Async() => _v2Index.Value;
+        public Task<PackageRegistrationIndexMetadata> GetIndexV3Async() => _v3Index.Value;
+        public Task<PackageRegistrationLeafMetadata> GetLeafV2Async() => _v2Leaf.Value;
+        public Task<PackageRegistrationLeafMetadata> GetLeafV3Async() => _v3Leaf.Value;
+
+        public Task<PackageTimestampMetadata> GetTimestampMetadataAsync()
         {
-            return _leafCache.GetOrAdd(key, lazyTask).Value;
+            return _timestampMetadataResourceV2.GetAsync(this);
```
scottbommarito: 1 - This is specifically the `PackageTimestampMetadata` for V2. The V3 `PackageTimestampMetadata` is constructed from the catalog entries in memory. You should change the naming so that this is clear.

2 - Unlike the other metadata fields, this isn't cached right now. You could easily rewrite `PackageTimestampMetadataResourceV2.GetAsync` to take in `Package` and `Client` directly and then align this with the way `GetIndexV2Async` and the other cached metadata are declared (with `Lazy`) so that it's cached..

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/ValidatorFactoryFactory.cs
```diff
@@ -22,35 +17,9 @@ public ValidatorFactoryFactory(ValidatorConfiguration validatorConfig, ILoggerFa
             _loggerFactory = loggerFactory ?? throw new ArgumentNullException(nameof(loggerFactory));
         }
 
-        public ValidatorFactory Create(string galleryUrl, string indexUrl)
+        public ValidatorFactory Create()
```
scottbommarito: Because the resource provider logic is no longer here, this class is pretty useless. I'd get rid of it and then call `ValidatorFactory` constructor in `PackageValidatorFactory.Create` directly.

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/ValidationContext.cs
dtivel: 1.  Done.
2.  Done, except I didn't change the `GetAsync(...)` method signature.  You'd also have to pass `DeletionAuditEntries`, but then why not just pass the `ValidationContext` instance as we already do?

PR #423, opened 2018-11-29
Add CatalogLeafFetcher for getting catalog leaves via registration

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/6446

We keep track of available/unavailable discoveries so that the caller can react to any information now available in the registration. The list of lists provided to the `GetLatestLeavesAsync` are created from the listed versions per `SearchFilters`. Basically we want to find a new latest version for each of the downgrades. This might be more than one version depending on the search filters.

Also fix up some broken config.

### DIFF
--- a/src/NuGet.Services.AzureSearch/Catalog2AzureSearch/CatalogLeafFetcher.cs
+++ b/src/NuGet.Services.AzureSearch/Catalog2AzureSearch/CatalogLeafFetcher.cs
@@ -1,0 +1,255 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using NuGet.Protocol.Catalog;
+using NuGet.Protocol.Registration;
+using NuGet.Versioning;
+
+namespace NuGet.Services.AzureSearch.Catalog2AzureSearch
+{
+    public class CatalogLeafFetcher : ICatalogLeafFetcher
+    {
+        private readonly IRegistrationClient _registrationClient;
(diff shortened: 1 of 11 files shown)


src/NuGet.Services.AzureSearch/Catalog2AzureSearch/CatalogLeafFetcher.cs
```diff
@@ -0,0 +1,253 @@
...
+                Page = page;
+                Range = new VersionRange(
+                    minVersion: NuGetVersion.Parse(page.Lower),
+                    includeMinVersion: true,
+                    maxVersion: NuGetVersion.Parse(page.Upper),
+                    includeMaxVersion: true);
+                RangeString = Range.ToNormalizedString();
+
+                if (page.Items != null)
+                {
```
loic-sharma: Is this null check necessary? Are there registration pages with null items? The docs say `items` is a required property: [https://docs.microsoft.com/en-us/nuget/api/registration-base-url-resource#registration-page](https://docs.microsoft.com/en-us/nuget/api/registration-base-url-resource#registration-page)

I'm only asking this out of curiosity, it's definitely better to be defensive around null values :)

src/NuGet.Services.AzureSearch/Catalog2AzureSearch/CatalogLeafFetcher.cs
```diff
@@ -0,0 +1,253 @@
...
+                        leaf.Url);
+                    return;
+                }
+                else
+                {
+                    _logger.LogInformation(
+                        "{PackageId} {Version} was found to be unlisted. This will not be used as a latest version.",
+                        packageId,
+                        version,
+                        leaf.Url);
```
loic-sharma: >leaf.Url [](start = 24, length = 8)

This is missing a field in the log message.

joelverhagen: Nice catch. I guess Microsoft.Extensions.Logging is okay with this (no throw).

PR #424, opened 2018-11-29
[Search 1] Add document builder logic

joelverhagen:

This is just a bunch of mapping logic. I used JSON serialization to assert the mapping because this is exactly what is sent to Azure Search. It also covers when new fields are added.

### DIFF
--- a/src/NuGet.Services.AzureSearch/DocumentUtilities.cs
+++ b/src/NuGet.Services.AzureSearch/DocumentUtilities.cs
@@ -1,0 +1,241 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Web;
+using NuGet.Frameworks;
+using NuGet.Protocol.Catalog;
+using NuGet.Services.Entities;
+using NuGet.Services.Metadata.Catalog;
+using NuGet.Versioning;
+using PackageDependency = NuGet.Protocol.Catalog.PackageDependency;
+
+namespace NuGet.Services.AzureSearch
+{
+    internal static class DocumentUtilities
+    {
+        private static readonly VersionRangeFormatter VersionRangeFormatter = new VersionRangeFormatter();
+        private static readonly DateTimeOffset UnlistedPublished = new DateTimeOffset(Metadata.Catalog.Constants.UnpublishedDate);
+
+        private static readonly HashSet<NuGetFramework> SpecialFrameworks = new HashSet<NuGetFramework>
+        {
+            NuGetFramework.AnyFramework,
+            NuGetFramework.AgnosticFramework,
+            NuGetFramework.UnsupportedFramework
+        };
+
+        public static void PopulateMetadata(
+            IBaseMetadataDocument document,
+            string packageId,
+            Package package)
+        {
+            document.Authors = package.FlattenedAuthors;
+            document.Copyright = package.Copyright;
+            document.Created = AssumeUtc(package.Created);
+            document.Description = package.Description;
+            document.FileSize = package.PackageFileSize;
+            document.FlattenedDependencies = package.FlattenedDependencies;
+            document.Hash = package.Hash;
+            document.HashAlgorithm = package.HashAlgorithm;
+            document.IconUrl = package.IconUrl;
+            document.Language = package.Language;
+            document.LastEdited = AssumeUtc(package.LastEdited);
+            document.LicenseUrl = package.LicenseUrl;
+            document.MinClientVersion = package.MinClientVersion;
+            document.NormalizedVersion = package.NormalizedVersion;
+            document.OriginalVersion = package.Version;
+            document.PackageId = packageId;
+            document.Prerelease = package.IsPrerelease;
+            document.ProjectUrl = package.ProjectUrl;
+            document.Published = package.Listed ? AssumeUtc(package.Published) : UnlistedPublished;
+            document.ReleaseNotes = package.ReleaseNotes;
+            document.RequiresLicenseAcceptance = package.RequiresLicenseAcceptance;
+            document.SemVerLevel = package.SemVerLevelKey;
+            document.Summary = package.Summary;
+            document.Tags = Utils.SplitTags(package.Tags ?? string.Empty);
+            document.Title = package.Title;
+        }
+
+        public static void PopulateMetadata(
+            IBaseMetadataDocument document,
+            string normalizedVersion,
+            PackageDetailsCatalogLeaf leaf)
+        {
+            document.Authors = leaf.Authors;
+            document.Copyright = leaf.Copyright;
+            document.Created = leaf.Created;
+            document.Description = leaf.Description;
+            document.FileSize = leaf.PackageSize;
+            document.FlattenedDependencies = GetFlattenedDependencies(leaf);
+            document.Hash = leaf.PackageHash;
+            document.HashAlgorithm = leaf.PackageHashAlgorithm;
+            document.IconUrl = leaf.IconUrl;
+            document.Language = leaf.Language;
+            document.LastEdited = leaf.LastEdited;
+            document.LicenseUrl = leaf.LicenseUrl;
+            document.MinClientVersion = leaf.MinClientVersion;
+            document.NormalizedVersion = normalizedVersion;
+            document.OriginalVersion = leaf.VerbatimVersion ?? leaf.PackageVersion;
+            document.PackageId = leaf.PackageId;
+            document.Prerelease = leaf.IsPrerelease;
+            document.ProjectUrl = leaf.ProjectUrl;
+            document.Published = leaf.Published;
+            document.ReleaseNotes = leaf.ReleaseNotes;
+            document.RequiresLicenseAcceptance = leaf.RequireLicenseAgreement;
+            document.SemVerLevel = leaf.IsSemVer2() ? 2 : (int?)null;
+            document.Summary = leaf.Summary;
+            document.Tags = leaf.Tags == null ? null : leaf.Tags.ToArray();
+            document.Title = leaf.Title;
+        }
+
+        public static string GetSearchDocumentKey(string packageId, SearchFilters searchFilters)
+        {
+            var lowerId = packageId.ToLowerInvariant();
+            var encodedId = EncodeKey(lowerId);
+            return $"{encodedId}-{searchFilters}";
+        }
+
+        public static string GetHijackDocumentKey(string packageId, string normalizedVersion)
+        {
+            var lowerId = packageId.ToLowerInvariant();
+            var lowerVersion = normalizedVersion.ToLowerInvariant();
+            return EncodeKey($"{lowerId}/{lowerVersion}");
+        }
+
+        private static string EncodeKey(string rawKey)
+        {
+            // First, encode the raw value for uniqueness.
+            var bytes = Encoding.UTF8.GetBytes(rawKey);
+            var unique = HttpServerUtility.UrlTokenEncode(bytes);
+
+            // Then, prepend a string as close to the raw key as possible, for readability.
+            var readable = ReplaceUnsafeKeyCharacters(rawKey).TrimStart('_');
+
+            return readable.Length > 0 ? $"{readable}-{unique}" : unique;
+        }
+
+        private static string ReplaceUnsafeKeyCharacters(string input)
+        {
+            return Regex.Replace(
+                input,
+                "[^A-Za-z0-9-_]", // Remove equal sign as well, since it's ugly.
+                "_",
+                RegexOptions.None,
+                TimeSpan.FromSeconds(30));
+        }
+
+        private static DateTimeOffset? AssumeUtc(DateTime? dateTime)
+        {
+            if (!dateTime.HasValue)
+            {
+                return null;
+            }
+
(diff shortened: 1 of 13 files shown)


joelverhagen:
🔔

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
```diff
@@ -0,0 +1,216 @@
...
+
+        private static DateTimeOffset? AssumeUtc(DateTime? dateTime)
+        {
+            if (!dateTime.HasValue)
+            {
+                return null;
+            }
+
+            return new DateTimeOffset(
+                DateTime.SpecifyKind(dateTime.Value, DateTimeKind.Utc),
```
dtivel: The call to `SpecifyKind(…)` is superfluous.  All it does is create a new `DateTime` object with the same `Ticks` value and a new `DateTimeKind` that you specify.  But the time zone is specified again below.

Just use [`DateTimeOffset(long, TimeSpan)`](https://docs.microsoft.com/en-us/dotnet/api/system.datetimeoffset.-ctor?view=netframework-4.7.2#System_DateTimeOffset__ctor_System_Int64_System_TimeSpan_):

    return new DateTimeOffset(dateTime.Ticks, TimeSpan.Zero);

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
```diff
@@ -0,0 +1,216 @@
...
+
+                    builder.Append(":");
+                    AddFlattenedFrameworkDependency(targetFramework, builder);
+                }
+            }
+
+            return builder.Length > 0 ? builder.ToString() : null;
+        }
+
+        private static void AddFlattennedPackageDependency(
```
dtivel: There's 1 "n" in "Flattened".

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        public static void PopulateMetadata(
+            IBaseMetadataDocument document,
+            string normalizedVersion,
+            PackageDetailsCatalogLeaf leaf)
+        {
+            document.Authors = leaf.Authors;
+            document.Copyright = leaf.Copyright;
+            document.Created = leaf.Created;
```
dtivel: Why not `AssumeUtc(…)` here, [here](https://github.com/NuGet/NuGet.Services.Metadata/pull/424/files#diff-e16587255d293a02ecfd8c0d95e68d3fR77), and [here](https://github.com/NuGet/NuGet.Services.Metadata/pull/424/files#diff-e16587255d293a02ecfd8c0d95e68d3fR85)?

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
```diff
@@ -0,0 +1,216 @@
...
+            return EncodeKey($"{lowerId}/{lowerVersion}");
+        }
+
+        private static string EncodeKey(string rawKey)
+        {
+            // First, encode the raw value for uniqueness.
+            var bytes = Encoding.UTF8.GetBytes(rawKey);
+            var unique = HttpServerUtility.UrlTokenEncode(bytes);
+
+            // Then, prepend a string as close to the raw key as possible, for readibility.
```
dtivel: readability

loic-sharma: >readibility [](start = 79, length = 11)

readability

src/NuGet.Services.AzureSearch/HijackDocumentBuilder.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        private static void PopulateFull(
+            HijackDocument.Full document,
+            string packageId,
+            string normalizedVersion,
+            HijackDocumentChanges changes,
+            string title)
+        {
+            PopulateLatest(document, packageId, normalizedVersion, changes);
+            document.SortableTitle = title ?? packageId;
```
dtivel: Why fall back to only package ID and not say, package identity?  Is this just preserving existing behavior?

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
loic-sharma: >AddFlattennedPackageDependency [](start = 28, length = 30)

`Flattenned` -> `Flattened` :)

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
```diff
@@ -0,0 +1,216 @@
...
+using NuGet.Services.Entities;
+using NuGet.Services.Metadata.Catalog;
+using NuGet.Versioning;
+using PackageDependency = NuGet.Protocol.Catalog.PackageDependency;
+
+namespace NuGet.Services.AzureSearch
+{
+    internal static class DocumentUtilities
+    {
+        private static readonly DateTimeOffset UnlistedPublished = new DateTimeOffset(1900, 1, 1, 0, 0, 0, TimeSpan.Zero);
```
dtivel: Can you reference `NuGet.Services.Metadata.Catalog.Constants.UnpublishedDate` instead of redefining this constant?

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
```diff
@@ -0,0 +1,216 @@
...
+                    }
+                }
+                else
+                {
+                    if (builder.Length > 0)
+                    {
+                        builder.Append("|");
+                    }
+
+                    builder.Append(":");
```
dtivel: It would be good to add documentation --- internal is fine --- about this custom serialization format.

src/NuGet.Services.AzureSearch/HijackDocumentBuilder.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        public HijackDocument.Full Full(
+            string normalizedVersion,
+            HijackDocumentChanges changes,
+            PackageDetailsCatalogLeaf leaf)
+        {
+            var document = new HijackDocument.Full();
+
+            PopulateFull(document, leaf.PackageId, normalizedVersion, changes, leaf.Title);
```
loic-sharma: >PopulateFull [](start = 12, length = 12)

Why doesn't `PopulateFull` call `PopulateMetadata`? It seems like "full" should include the metadata

tests/NuGet.Services.AzureSearch.Tests/DocumentUtilitiesFacts.cs
```diff
@@ -0,0 +1,63 @@
...
+using Xunit;
+
+namespace NuGet.Services.AzureSearch
+{
+    public class DocumentUtilitiesFacts
+    {
+        public class GetHijackDocumentKey
+        {
+            [Theory]
+            [InlineData("NuGet.Versioning", "1.0.0", "nuget_versioning_1_0_0-bnVnZXQudmVyc2lvbmluZy8xLjAuMA2")]
```
loic-sharma: Nice work making the keys human readable 👍

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
joelverhagen: Great point!

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
joelverhagen: They are already `DateTimeOffset` which have specific timezone information.

src/NuGet.Services.AzureSearch/HijackDocumentBuilder.cs
joelverhagen: Yes, existing behavior.

src/NuGet.Services.AzureSearch/HijackDocumentBuilder.cs
joelverhagen: The parameters of `PopulateMetadata` vary for catalog2azuresearch vs. db2azuresearch. This is an internal method that is meant to sent the fields that have the same parameters between these two cases.

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
joelverhagen: That's a `DateTime`, but I'll re-use it for another static.

src/NuGet.Services.AzureSearch/HijackDocumentBuilder.cs
loic-sharma: Ah right. Ignore my comment

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
```diff
@@ -0,0 +1,239 @@
...
+            return new DateTimeOffset(dateTime.Value.Ticks, TimeSpan.Zero);
+        }
+
+        /// <summary>
+        /// This method produces output for official client. The implementation on the client side, at one point of time
+        /// was:
+        /// https://github.com/NuGet/NuGet.Client/blob/b404acf6eb88c2b6086a9cbb5106104534de2428/src/NuGet.Core/NuGet.Protocol/LegacyFeed/V2FeedPackageInfo.cs#L228-L308
+        /// 
+        /// The output is a string where each dependency is seperated by a "|" character and information about each
+        /// dependency is seperated by a ":" character. Per dependency, the colon-seperated data is a pair or triple.
```
dtivel: separated

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
```diff
@@ -0,0 +1,239 @@
...
+        /// The first field is the dependency package ID. If it is an empty string, then the version range must also be
```
dtivel: "[I]t" in the second sentence is ambiguous.  How about:

    The three fields in order are:  dependency package ID, version range, and target framework.  If the first field is an empty string, ...

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
```diff
@@ -0,0 +1,239 @@
...
+        /// and empty strings but the framework is included, this means that the package supports the specified
+        /// framework but has no dependencies specific to that framework.
+        /// 
+        /// Output:
+        ///     [DEPENDENCY[|DEPENDENCY][|...][|DEPENDENCY]]
+        /// Each dependency:
+        ///     [PACKAGE_ID]:[VERSION_RANGE][:TARGET_FRAMEWORK]
+        ///   
+        /// Example A (no target frameworks):
+        ///     Microsoft.Data.OData:5.0.2|Microsoft.WindowsAzure.ConfigurationManager:1.8.0
```
dtivel: Per your description above, this is also valid:

    Microsoft.Data.OData:5.0.2:|Microsoft.WindowsAzure.ConfigurationManager:1.8.0:

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
dtivel: Flattenned is still misspelled.

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
```diff
@@ -0,0 +1,239 @@
...
+            {
+                try
+                {
+                    builder.Append(":");
+                    builder.Append(targetFramework?.GetShortFolderName());
+                }
+                catch (FrameworkException)
+                {
+                    // ignoring FrameworkException on purpose - we don't want the job crashing
+                    // whenever someone uploads an unsupported framework
```
dtivel: This could result in the following output string:

    ::

Is that valid or even useful?

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
```diff
@@ -0,0 +1,239 @@
...
+            }
+
+            builder.Append(packageDependency.Id);
+            builder.Append(":");
+
+            var versionRange = packageDependency.ParseRange();
+
+            if (!VersionRange.All.Equals(versionRange))
+            {
+                builder.Append(versionRange?.ToString("S", new VersionRangeFormatter()));
```
dtivel: Can the `VersionRangeFormatter` instance be `static readonly`?

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
joelverhagen: Oh, my fault, I confused this instance for the one Loic mentioned.

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
joelverhagen: Yes, that's right. I'll add this to the example to emphasize that.

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
joelverhagen: Sure, good idea.

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
joelverhagen: Sure!

src/NuGet.Services.AzureSearch/DocumentUtilities.cs
joelverhagen: Valid? Yes. This is existing behavior. Useful? Well, this is essentially the same assertion as a `<group />` element in the .nuspec... which today is noise (all other dependency groups are more applicable than the "any" TFM. But again this is existing behavior so it's not worth the risk to change it.

PR #426, opened 2018-12-03
[Search 2] Allow parallel version list replacement in BatchPusher

joelverhagen:
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/424.

Polish HijackDocumentChanges and VersionListDataClient

### DIFF
--- a/src/NuGet.Services.AzureSearch/BatchPusher.cs
+++ b/src/NuGet.Services.AzureSearch/BatchPusher.cs
@@ -2,10 +2,12 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
+using Microsoft.Azure.Search;
 using Microsoft.Azure.Search.Models;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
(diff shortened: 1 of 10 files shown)


src/NuGet.Services.AzureSearch/VersionList/HijackDocumentChanges.cs
```diff
@@ -15,6 +15,16 @@ public class HijackDocumentChanges
             bool latestStableSemVer2,
             bool latestSemVer2)
         {
+            if (delete && updateMetadata)
+            {
+                throw new ArgumentException("A deleting and updating a hijack document are mutually exclusive.");
```
loic-sharma: >A deleting [](start = 45, length = 10)

`A deleting...` -> `Deleting and...`

src/NuGet.Services.AzureSearch/BatchPusher.cs
```diff
@@ -117,22 +126,31 @@ private async Task PushBatchesAsync(bool onlyFull)
...
+                        {
+                            await Task.Yield();
+                            while (work.TryTake(out var finished))
+                            {
+                                await _versionListDataClient.ReplaceAsync(
+                                    finished.Id,
+                                    finished.Value.Result,
+                                    finished.Value.AccessCondition);
+                            }
+                        });
```
loic-sharma: >}); [](start = 24, length = 3)

I would `ToList` this to be safe. Someone could accidentally double the number of workers by adding a log of `tasks.Count`

joelverhagen: Sounds good. I normally do this for that very reason, but I guess I forgot here.

PR #444, opened 2018-12-11
Determine IsLatest and IsLatestStable for the search documents

joelverhagen:
V2 search has `IsLatest` and `IsLatestStable`. We already have these values in the hijack index but we need them in the search service as well for `/api/v2/Search()`.

When the search filter includes prerelease, we can determine the `IsLatestStable` by using the inverse of the latest version's `Prerelease` flag.

When the search filter does not include the prerelease, we can determine the `IsLatest` by checking the latest version on the corresponding search filter that includes prerelease.

### DIFF
--- a/src/NuGet.Services.AzureSearch/Catalog2AzureSearch/CatalogIndexActionBuilder.cs
+++ b/src/NuGet.Services.AzureSearch/Catalog2AzureSearch/CatalogIndexActionBuilder.cs
@@ -191,7 +191,10 @@ namespace NuGet.Services.AzureSearch.Catalog2AzureSearch
             SearchFilters searchFilters,
             SearchIndexChangeType changeType)
         {
-            var latest = context.VersionLists.GetLatestVersionInfoOrNull(searchFilters);
+            var latestFlags = _search.LatestFlagsOrNull(context.VersionLists, searchFilters);
+            Guard.Assert(
+                changeType == SearchIndexChangeType.Delete || latestFlags != null,
+                "Either the search document is being or there is a latest version.");
 
             switch (changeType)
             {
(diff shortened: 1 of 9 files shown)


src/NuGet.Services.AzureSearch/Models/SearchDocument.cs
```diff
@@ -59,6 +63,25 @@ public class UpdateVersionList : KeyedDocument, IVersions
...
         {
             string[] Versions { get; set; }
+            bool? IsLatestStable { get; set; }
+            bool? IsLatest { get; set; }
+        }
+
+        /// <summary>
+        /// The data required to populate <see cref="IVersions"/> and other <see cref="SearchDocument"/> classes.
+        /// </summary>
+        public class LatestFlags
```
loic-sharma: From my understanding, these are the `LatestFlags` for a specific search filter. If so, please mention that in the comment to add a little more context

src/NuGet.Services.AzureSearch/SearchDocumentBuilder.cs
```diff
@@ -8,6 +8,74 @@ namespace NuGet.Services.AzureSearch
...
     {
+        public SearchDocument.LatestFlags LatestFlagsOrNull(VersionLists versionLists, SearchFilters searchFilters)
+        {
+            var latest = versionLists.GetLatestVersionInfoOrNull(searchFilters);
+            if (latest == null)
+            {
+                return null;
+            }
+
+            // The latest version, given the "include prerelease" bit of the search filter, may or may not be the
```
loic-sharma: Fantastic comment 👍

src/NuGet.Services.AzureSearch/SearchDocumentBuilder.cs
```diff
@@ -8,6 +8,74 @@ namespace NuGet.Services.AzureSearch
 {
     public class SearchDocumentBuilder : ISearchDocumentBuilder
     {
+        public SearchDocument.LatestFlags LatestFlagsOrNull(VersionLists versionLists, SearchFilters searchFilters)
+        {
+            var latest = versionLists.GetLatestVersionInfoOrNull(searchFilters);
+            if (latest == null)
```
loic-sharma: I seem to have forgotten - in what cases would `latest` be null again? Is it if you have an incomplete `versionLists`? When does that happen? I ask as on line 65 you call `GetLatestVersionInfoOrNull` again, but that time you don't do a null check

src/NuGet.Services.AzureSearch/Models/SearchDocument.cs
joelverhagen: All `SearchEntities` are per search filter. Sure, i'll add a comment.

src/NuGet.Services.AzureSearch/SearchDocumentBuilder.cs
```diff
@@ -8,6 +8,74 @@ namespace NuGet.Services.AzureSearch
...
+            // IsLatest is true or false without looking at the version list that includes prerelease versions. For
+            // cases #3 and #4, we know IsLatest will always be true and we can determine IsLatestStable by looking
+            // at whether the latest version is prerelease or not.
+            bool isLatestStable;
+            bool isLatest;
+            if ((searchFilters & SearchFilters.IncludePrerelease) == 0)
+            {
+                // This is the case where prerelease versions are excluded.
+                var latestIncludePrerelease = versionLists
+                    .GetLatestVersionInfoOrNull(searchFilters | SearchFilters.IncludePrerelease);
```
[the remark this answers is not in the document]
joelverhagen: If a search filter excluding prerelease has a latest version, then a more broad filter (one that includes prerelease) will also have a latest version. Said another way, `searchFilters | SearchFilters.IncludePrerelease` represents a superset of the packages matched by `searchFilters`.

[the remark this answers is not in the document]
joelverhagen: I'll add a `Guard` to clarify this.

src/NuGet.Services.AzureSearch/Catalog2AzureSearch/CatalogIndexActionBuilder.cs
```diff
@@ -191,7 +191,7 @@ private async Task<IndexChanges> GetIndexChangesAsync(Context context)
             SearchFilters searchFilters,
             SearchIndexChangeType changeType)
         {
-            var latest = context.VersionLists.GetLatestVersionInfoOrNull(searchFilters);
+            var latestFlags = _search.LatestFlagsOrNull(context.VersionLists, searchFilters);
```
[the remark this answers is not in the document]
joelverhagen: I'll add a Guard. I find this better than an assertion that is just in English but not in code.

PR #449, opened 2018-12-13
[Search 1] Add debugging properties to documents submitted to Azure Search

joelverhagen:
See `CommittedDocument` for the specific properties.

Since we have document batching, I made the `LastUpdatedDocument` property a little clever to capture the current timestamp as late as possible. I thought it was worth the trick to get a timestamp that is close to when the actual submission to Azure Search is (chronological). If I don't do this, the timestamp can be off by as much time as it takes to build up a full batch to Azure Search (1000 documents).

### DIFF
--- a/src/NuGet.Services.AzureSearch/Models/CurrentTimestamp.cs
+++ b/src/NuGet.Services.AzureSearch/Models/CurrentTimestamp.cs
@@ -1,0 +1,60 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+
+namespace NuGet.Services.AzureSearch
+{
+    /// <summary>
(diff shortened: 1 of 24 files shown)


src/NuGet.Services.AzureSearch/Models/CurrentTimestamp.cs
```diff
@@ -0,0 +1,59 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+
+namespace NuGet.Services.AzureSearch
+{
+    /// <summary>
+    /// A "last updated" timestamp set on an Azure Search is approximate since there is some undefied time between when
```
loic-sharma: > Azure Search  [](start = 44, length = 14)

`Azure Search` -> `Azure Search document`?

`undefied` -> `undefined`

joelverhagen: Yes.

Yes.

PR #454, opened 2018-12-17
[Search 1] Add RegistrationUrlBuilder to consolidate URL building logic

joelverhagen:
Update NuGet.Services.Validation.Common.Job dependency (brings newer gallery and client APIs)


joelverhagen:
This will be used by the V3 search result since it has registration URLs in the response body.

PR #455, opened 2018-12-18
[Search 2] Add auxiliary file loading infrastructure

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/6449

1. Load the downloads and verified packages auxiliary files using existing parsers
1. No-op loading the auxiliary files using etags
1. Introduce a simple, synchronous interface `IAuxiliaryData` which can be used in a non-async context. This is ideal for code-paths that should be fast-ish, like building response bodies.

`AuxiliaryFileReloader` is the guy that will be launched on app start to reload the auxiliary files in the background.

`AuxiliaryDataCache.InitializeAsync` must be called in an async context before `AuxiliaryDataCache.Get` is called.

### DIFF
--- a/src/NuGet.Services.AzureSearch/SearchService/AuxiliaryFileClient.cs
+++ b/src/NuGet.Services.AzureSearch/SearchService/AuxiliaryFileClient.cs
@@ -1,0 +1,149 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Net;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Microsoft.WindowsAzure.Storage;
+using Newtonsoft.Json;
+using NuGet.Indexing;
+using NuGetGallery;
+
+namespace NuGet.Services.AzureSearch.SearchService
+{
+    public class AuxiliaryFileClient : IAuxiliaryFileClient
+    {
+        private readonly ICloudBlobClient _cloudBlobClient;
+        private readonly IOptionsSnapshot<SearchServiceConfiguration> _options;
+        private readonly ILogger<AuxiliaryFileClient> _logger;
+        private readonly Lazy<ICloudBlobContainer> _lazyContainer;
+
+        public AuxiliaryFileClient(
+            ICloudBlobClient cloudBlobClient,
+            IOptionsSnapshot<SearchServiceConfiguration> options,
+            ILogger<AuxiliaryFileClient> logger)
+        {
+            _cloudBlobClient = cloudBlobClient ?? throw new ArgumentNullException(nameof(cloudBlobClient));
+            _options = options ?? throw new ArgumentNullException(nameof(options));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+            _lazyContainer = new Lazy<ICloudBlobContainer>(
+                () => _cloudBlobClient.GetContainerReference(_options.Value.AuxiliaryDataStorageContainer));
+        }
+
+        private ICloudBlobContainer Container => _lazyContainer.Value;
+
+        public async Task<AuxiliaryFileResult<Downloads>> LoadDownloadsAsync(string etag)
+        {
+            return await LoadAuxiliaryFileAsync(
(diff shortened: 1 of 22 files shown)


joelverhagen:
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/454.

src/NuGet.Services.AzureSearch/SearchService/AuxiliaryFileClient.cs
```diff
@@ -0,0 +1,149 @@
...
+        public AuxiliaryFileClient(
+            ICloudBlobClient cloudBlobClient,
+            IOptionsSnapshot<SearchServiceConfiguration> options,
+            ILogger<AuxiliaryFileClient> logger)
+        {
+            _cloudBlobClient = cloudBlobClient ?? throw new ArgumentNullException(nameof(cloudBlobClient));
+            _options = options ?? throw new ArgumentNullException(nameof(options));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+            _lazyContainer = new Lazy<ICloudBlobContainer>(
```
loic-sharma: >_lazyContainer [](start = 12, length = 14)

Does this need to be lazy? AFAIK, this just creates some objects (see [this](https://www.fuget.org/packages/WindowsAzure.Storage/9.3.3/lib/netstandard1.3/Microsoft.WindowsAzure.Storage.dll/Microsoft.WindowsAzure.Storage.Blob/CloudBlobClient?code=true#M%3AMicrosoft.WindowsAzure.Storage.Blob.CloudBlobClient.GetContainerReference%28System.String%29)).

src/NuGet.Services.AzureSearch/SearchService/AuxiliaryFileClient.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        private async Task<AuxiliaryFileResult<T>> LoadAuxiliaryFileAsync<T>(
+            string blobName,
+            string etag,
+            Func<ILoader, T> loadData) where T : class
+        {
+            _logger.LogInformation("Attempted to load blob {BlobName} with etag {ETag}.", blobName, etag);
+
+            var stopwatch = Stopwatch.StartNew();
+            var blob = Container.GetBlobReference(blobName);
```
loic-sharma: Can we use the `TelemetryClient` here? It'd be useful to record these durations as metrics too.

src/NuGet.Services.AzureSearch/SearchService/Models/AuxiliaryFilesMetadata.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        [JsonConstructor]
+        public AuxiliaryFilesMetadata(AuxiliaryFileMetadata downloads, AuxiliaryFileMetadata verifiedPackages)
+        {
+            Downloads = downloads ?? throw new ArgumentNullException(nameof(downloads));
+            VerifiedPackages = verifiedPackages ?? throw new ArgumentNullException(nameof(verifiedPackages));
+        }
+
+        public AuxiliaryFileMetadata Downloads { get; }
+        public AuxiliaryFileMetadata VerifiedPackages { get; }
```
loic-sharma: >         [](start = 62, length = 8)

There's some extra space at the end of this line.

src/NuGet.Services.AzureSearch/SearchService/AuxiliaryFileClient.cs
joelverhagen: Yeah, I've investigated the same thing before. I chose this pattern since it was undocumented at the time of my investigation about what kind of validation or "work" is done here. In general, I prefer the debugging experience of failures in method bodies (ASP.NET "actions") instead of in activation (in the ASP.NET pipeline before you enter the constructor).

If you feel strongly I can change this.

src/NuGet.Services.AzureSearch/SearchService/AuxiliaryFileClient.cs
joelverhagen: Tracked https://github.com/NuGet/NuGetGallery/issues/6664.

src/NuGet.Services.AzureSearch/SearchService/AuxiliaryFileClient.cs
joelverhagen: I pushed this PR but if you want follow-up I can take care of that in a subsequent PR. Just let me know.

PR #458, opened 2019-01-02
 Add AzureSearchService to orchestrate querying the index and building the response

joelverhagen:
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/457

Progress on https://github.com/NuGet/NuGetGallery/issues/6450
Progress on https://github.com/NuGet/NuGetGallery/issues/6453

`AzureSearchService` is the top level unit of business logic for performing search queries. It is meant to be one step below the "view concern" like an ASP.NET controller.


src/NuGet.Services.AzureSearch/SearchService/AzureSearchService.cs
```diff
@@ -0,0 +1,84 @@
...
+                return await UseSearchIndexAsync(request);
+            }
+        }
+
+        public async Task<V3SearchResponse> V3SearchAsync(V3SearchRequest request)
+        {
+            var text = _parametersBuilder.GetSearchTextForV3Search(request);
+            var parameters = _parametersBuilder.GetSearchParametersForV3Search(request);
+
+            var stopwatch = Stopwatch.StartNew();
```
agr: What do you think about introducing a wrapper, so this `Stopwatch.StartNew()`, `<do your thing>`, `stopwatch.Stop()` pattern is not repeated over and over?
Something like 
```C#
static TimeSpan Measure(Func<Task> action)
{
  var sw = Stopwatch.StartNew();
  action();
  sw.Stop();
  return sw.Elapsed;
}
```
Then in the code:
```C#
var sw = Something.Measure(async () => result = await _searchIndex.Documents.SearchAsync(...));
```

joelverhagen: Sure!

PR #461, opened 2019-01-03
Clean up dependency injection and continue work on search service entry point

joelverhagen:
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/459

### DIFF
--- a/src/NuGet.Services.SearchService/App_Start/WebApiConfig.cs
+++ b/src/NuGet.Services.SearchService/App_Start/WebApiConfig.cs
@@ -3,6 +3,8 @@
 
 using System;
 using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
 using System.Web.Hosting;
 using System.Web.Http;
 using System.Web.Http.Dependencies;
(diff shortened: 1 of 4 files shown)


joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/6450
Progress on https://github.com/NuGet/NuGetGallery/issues/6453

src/NuGet.Services.SearchService/App_Start/WebApiConfig.cs
```diff
@@ -40,11 +47,31 @@ public static void Register(HttpConfiguration config)
...
+
+            HostingEnvironment.QueueBackgroundWorkItem(token => ReloadAuxiliaryFilesAsync(dependencyResolver, token));
+        }
+
+        private static async Task ReloadAuxiliaryFilesAsync(IDependencyResolver dependencyResolver, CancellationToken token)
+        {
+            using (var scope = dependencyResolver.BeginScope())
+            {
+                var loader = (IAuxiliaryFileReloader)scope.GetService(typeof(IAuxiliaryFileReloader));
+                await loader.ReloadContinuouslyAsync(token);
```
loic-sharma: We've had issues where a database context was used in a scope that was never recycled. I know this doesn't use a database context today, but I'm a little concerned about this scope's lifetime. Do we even need a scope here? It seems the loader doesn't depend on any scoped services.

joelverhagen: Sure, I think I was hoping to isolate the background worker as much as possible from the root scope, but it's probably better to keep this simple until we have a problem.

PR #467, opened 2019-01-09
Use environment variables for configuration in Search Service

joelverhagen:
I will replace the package reference with a stable version when https://github.com/NuGet/ServerCommon/pull/272 is merged and published.

I left the JSON file so that it is easier to debug locally.

### DIFF
--- a/src/NuGet.Services.SearchService/App_Start/WebApiConfig.cs
+++ b/src/NuGet.Services.SearchService/App_Start/WebApiConfig.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.IO;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


src/NuGet.Services.SearchService/App_Start/WebApiConfig.cs
```diff
@@ -94,25 +94,30 @@ private static IDependencyResolver GetDependencyResolver(HttpConfiguration confi
...
 
-            var builder = new ConfigurationBuilder()
-                .SetBasePath(basePath)
-                .AddJsonFile(configurationFilename, optional: false, reloadOnChange: false);
-
-            var uninjectedConfiguration = builder.Build();
+            // Load the configuration without injection. This allows us to read KeyVault configuration.
+            var uninjectedBuilder = new ConfigurationBuilder()
+                .AddJsonFile(jsonFile) // The JSON file is useful for local development.
+                .AddEnvironmentVariables(prefix); // Environment variables take precedence.
```
scottbommarito: To clarify--does the last thing added take precedence?

joelverhagen: Yes.

PR #468, opened 2019-01-10
Add /search/diag endpoint to Azure Search Service

joelverhagen:
I changed the response body to have more helpful information. 

Address https://github.com/nuget/nugetgallery/issues/6454

Output looks like this:

```json
{
    "Success": true,
    "Server": {
        "MachineName": "RD00155D941F25",
        "ProcessId": 12376,
        "ProcessStartTime": "2019-01-10T03:25:46.7568469+00:00",
        "ProcessDuration": "00:02:35.3677385",
        "DeploymentLabel": "DEV-USNC-4.0.0-jver-diag-2320652",
        "AssemblyCommitId": "796a27a",
        "AssemblyInformationalVersion": "4.0.0-jver-azsdep-2320652",
        "AssemblyBuildDateUtc": "01/10/2019 03:13:07 +00:00",
        "InstanceId": "8da5b5cd140c25c0d571bac7c5d4593bcc2f07fa87f6215e576e76ca1fc30873"
    },
    "SearchIndex": {
        "Name": "search-dev-20181214-00",
        "DocumentCount": 503741,
        "WarmQueryDuration": "00:00:00.1491502"
    },
    "HijackIndex": {
        "Name": "hijack-dev-20181214-00",
        "DocumentCount": 1798018,
        "WarmQueryDuration": "00:00:00.0913182"
    },
    "AuxiliaryFiles": {
        "Downloads": {
            "LastModified": "2018-12-17T19:57:48+00:00",
            "Loaded": "2019-01-10T03:27:43.6653936+00:00",
            "LoadDuration": "00:00:19.0527679",
            "FileSize": 42507754,
            "ETag": "\"0x8D66459EB34E3FB\""
        },
        "VerifiedPackages": {
            "LastModified": "2018-12-17T19:57:49+00:00",
            "Loaded": "2019-01-10T03:27:25.6290707+00:00",
            "LoadDuration": "00:00:00.6896924",
            "FileSize": 370219,
            "ETag": "\"0x8D66459EBD0F1C5\""
        }
    }
}
```

### DIFF
--- a/src/NuGet.Services.AzureSearch/Measure.cs
+++ b/src/NuGet.Services.AzureSearch/Measure.cs
@@ -1,0 +1,30 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.AzureSearch
+{
+    public static class Measure
+    {
+        public static async Task<DurationMeasurement<T>> DurationWithValueAsync<T>(Func<Task<T>> actAsync)
+        {
+            var stopwatch = Stopwatch.StartNew();
+            var value = await actAsync();
+            stopwatch.Stop();
+            return new DurationMeasurement<T>(value, stopwatch.Elapsed);
+        }
+
+        public static async Task<TimeSpan> DurationAsync(Func<Task> actAsync)
+        {
+            var result = await DurationWithValueAsync(async () =>
+            {
+                await actAsync();
+                return 0;
+            });
+            return result.Duration;
+        }
+    }
+}
--- a/src/NuGet.Services.SearchService/Settings/local.json
+++ b/src/NuGet.Services.SearchService/Settings/local.json
@@ -11,7 +11,8 @@
     "AuxiliaryDataStorageDownloadsPath": "downloads.v1.json",
     "AuxiliaryDataStorageVerifiedPackagesPath": "verifiedPackages.json",
     "AuxiliaryDataReloadFrequency": "01:00:00",
-    "AuxiliaryDataReloadFailureRetryFrequency": "00:00:30"
+    "AuxiliaryDataReloadFailureRetryFrequency": "00:00:30",
+    "DeploymentLabel": "1.0.0-zlocal"
   },
 
   "KeyVault_VaultName": "",
--- a/src/NuGet.Services.AzureSearch/DependencyInjectionExtensions.cs
+++ b/src/NuGet.Services.AzureSearch/DependencyInjectionExtensions.cs
@@ -76,6 +76,14 @@ namespace NuGet.Services.AzureSearch
                     c.ResolveKeyed<ISearchIndexClientWrapper>(searchIndexKey),
                     c.ResolveKeyed<ISearchIndexClientWrapper>(hijackIndexKey),
                     c.Resolve<ISearchResponseBuilder>()));
+
+            containerBuilder
+                .Register<ISearchStatusService>(c => new SearchStatusService(
+                    c.ResolveKeyed<ISearchIndexClientWrapper>(searchIndexKey),
+                    c.ResolveKeyed<ISearchIndexClientWrapper>(hijackIndexKey),
+                    c.Resolve<IAuxiliaryDataCache>(),
+                    c.Resolve<IOptionsSnapshot<SearchServiceConfiguration>>(),
+                    c.Resolve<ILogger<SearchStatusService>>()));
         }
 
         private static void RegisterAzureSearchJobStorageServices(ContainerBuilder containerBuilder, string key)
--- a/src/NuGet.Services.AzureSearch/DurationMeasurement.cs
+++ b/src/NuGet.Services.AzureSearch/DurationMeasurement.cs
@@ -1,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace NuGet.Services.AzureSearch
+{
+    public class DurationMeasurement<T>
+    {
+        public DurationMeasurement(T result, TimeSpan duration)
+        {
+            Value = result;
+            Duration = duration;
+        }
+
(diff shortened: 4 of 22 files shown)


src/NuGet.Services.AzureSearch/Measure.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using System;
+using System.Diagnostics;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.AzureSearch
+{
+    public static class Measure
+    {
+        public static async Task<DurationMeasurement<T>> DurationWithValueAsync<T>(Func<Task<T>> getValueAsync)
```
cristinamanum: Maybe use the same argument name as below :"actAsync" ?

src/NuGet.Services.AzureSearch/SearchService/Models/StatusResponse.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Services.AzureSearch.SearchService
+{
+    public class StatusResponse
+    {
+        public bool Success { get; set; }
```
cristinamanum: What does the Success mean? Is it correlated with the values below?

joelverhagen: I'll leave a code comment. Success means that all of the search service dependencies could be interacted with successfully. This essentially means that the search and hijack indexes can be queried and that the auxiliary data was loaded at least once (stale data is okay).

src/NuGet.Services.SearchService/Settings/local.json
```diff
@@ -11,7 +11,8 @@
     "AuxiliaryDataStorageDownloadsPath": "downloads.v1.json",
     "AuxiliaryDataStorageVerifiedPackagesPath": "verifiedPackages.json",
     "AuxiliaryDataReloadFrequency": "01:00:00",
-    "AuxiliaryDataReloadFailureRetryFrequency": "00:00:30"
+    "AuxiliaryDataReloadFailureRetryFrequency": "00:00:30",
+    "DeploymentLabel": "1.0.0-zlocal"
```
skofman1: As far as I know deployment labels are generated by EV2. Is the intention here for EV2 to place the deployment label in the config file?

src/NuGet.Services.AzureSearch/Measure.cs
joelverhagen: Sounds good.

src/NuGet.Services.SearchService/Settings/local.json
joelverhagen: This config file is only observed locally (thus the `local.json` file name). It is used for debugging on your dev box. In the startup of the app, we define that environment variables (i.e. app settings in Azure Web Apps) take precedence over this file. Ev2 will place all app settings, including its deployment label as environment variables, so this `1.0.0-zlocal` value will never be seen in production.

src/NuGet.Services.AzureSearch/SearchService/StatusService.cs
```diff
@@ -0,0 +1,145 @@
...
+            ILogger<StatusService> logger)
+        {
+            _searchIndex = searchIndex ?? throw new ArgumentNullException(nameof(searchIndex));
+            _hijackIndex = hijackIndex ?? throw new ArgumentNullException(nameof(hijackIndex));
+            _auxiliaryDataCache = auxiliaryDataCache ?? throw new ArgumentNullException(nameof(auxiliaryDataCache));
+            _options = options ?? throw new ArgumentNullException(nameof(options));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public async Task<StatusResponse> GetStatusAsync(Assembly assemblyForMetadata)
```
skofman1: >assemblyForMetadata [](start = 66, length = 19)

why is this input. Will it no always be 
typeof(SearchController).Assembly

joelverhagen: Search controller is in another (non-dependent) assembly. 

Also, this supports unit testing where I can set arbitrary assembly metadata.

src/NuGet.Services.AzureSearch/SearchService/StatusService.cs
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        private async Task TryAsync<T>(
+            Func<Task<T>> getAsync,
+            StatusResponse response,
+            string operation)
+        {
+            try
+            {
+                await Task.Yield();
```
cristinamanum: Why Task.Yield() is needed?

joelverhagen: So that all four of the status sub-operations run in parallel -- even leading, synchronous steps.

src/NuGet.Services.AzureSearch/SearchService/StatusService.cs
```diff
@@ -0,0 +1,145 @@
...
+                .GetCustomAttributes<AssemblyMetadataAttribute>()
+                .Where(x => x.Key == name)
+                .Select(x => x.Value)
+                .FirstOrDefault();
+        }
+
+        private static async Task<IndexStatus> GetIndexStatusAsync(ISearchIndexClientWrapper index)
+        {
+            var documentCount = await index.Documents.CountAsync();
+            var warmQueryDuration = await Measure.DurationAsync(() => index.Documents.SearchAsync("*", new SearchParameters()));
```
cristinamanum: Will this query get more expensive as more documents are added? Will this method executed often, could be any server impact?

joelverhagen: I found no indication from documentation that it will become _significantly_ slower. The index I tested it on had full production data in it and was running on the minimum replica/partition count that we would go live with.

If this query is slow we will bigger problems. This is essentially the default query that is done when someone opens the [default packages list on gallery](https://www.nuget.org/packages) or on client.

joelverhagen: By the way, this is mitigated by https://github.com/nuget/nugetgallery/issues/6463.

PR #470, opened 2019-01-15
Catalog2lucene - License Url update for license expression/file

zhhyu:
For catalog2lucene pipeline, when the package contains license expression/file, the license Url should be updated as expected: "~/packages/id/version/license". 

Fix: https://github.com/NuGet/NuGetGallery/issues/6526


src/NuGet.Indexing/Extraction/CatalogPackageMetadataExtraction.cs
```diff
@@ -55,18 +57,52 @@ private class Extractor
...
 
                 return _metadata;
             }
 
+            private void AddLicenseUrl()
+            {
+                if (_galleryBaseAddress != null && _catalog[MetadataConstants.LicenseExpressionPropertyName] != null)
+                {
+                    _metadata[MetadataConstants.LicenseUrlPropertyName] = GetOverWriteLicenseUrl();
+                } else if (_galleryBaseAddress != null && _catalog[MetadataConstants.LicenseFilePropertyName] != null)
```
joelverhagen: nit: `else if` should be on a different line that the curlies

src/NuGet.Indexing/Extraction/CatalogPackageMetadataExtraction.cs
```diff
@@ -55,18 +57,52 @@ private class Extractor
...
+                AddLicenseUrl();
                 AddFlattenedDependencies();
                 AddSupportedFrameworks();
 
                 return _metadata;
             }
 
+            private void AddLicenseUrl()
+            {
+                if (_galleryBaseAddress != null && _catalog[MetadataConstants.LicenseExpressionPropertyName] != null)
```
skofman1: This if statement can be simplified:
  if (_galleryBaseAddress != null && 
(_catalog[MetadataConstants.LicenseExpressionPropertyName] != null || _catalog[MetadataConstants.LicenseFilePropertyName] != null))
{
_metadata[MetadataConstants.LicenseUrlPropertyName] = GetOverWriteLicenseUrl();
}
else
{
AddString(MetadataConstants.LicenseUrlPropertyName);
}

src/NuGet.Indexing/Extraction/CatalogPackageMetadataExtraction.cs
```diff
@@ -55,18 +57,52 @@ private class Extractor
...
+                } else if (_galleryBaseAddress != null && _catalog[MetadataConstants.LicenseFilePropertyName] != null)
+                {
+                    _metadata[MetadataConstants.LicenseUrlPropertyName] = GetOverWriteLicenseUrl();
+                } else
+                {
+                    AddString(MetadataConstants.LicenseUrlPropertyName);
+                }
+            }
+
+            private string GetOverWriteLicenseUrl()
```
skofman1: We should have a similar method in Ctalog2Registration. Can we share the code?

src/V3PerPackage/PerBatchProcessor.cs
```diff
@@ -282,6 +282,7 @@ private async Task ExecuteCatalog2LuceneAsync(PerBatchContext context, Uri catal
                     commitEachBatch,
                     commitTimeout,
                     baseAddress,
+                    context.Global.GalleryBaseAddress.AbsoluteUri,
```
agr: >GalleryBaseAddress [](start = 35, length = 18)

Can `GalleryBaseAddress` be null here?

src/Ng/SearchIndexFromCatalogCollector.cs
```diff
@@ -38,6 +39,7 @@ public class SearchIndexFromCatalogCollector : CommitCollector
             bool commitEachBatch,
             TimeSpan? commitTimeout,
             string baseAddress,
+            string galleryBaseAddress,
             ITelemetryService telemetryService,
```
agr: This should probably by a `Uri` object instead of a `string`.

src/NuGet.Indexing/Extraction/CatalogPackageMetadataExtraction.cs
zhhyu: We have a  sharing license helper to get license Url now.

src/V3PerPackage/PerBatchProcessor.cs
zhhyu: Use Uri rather than string here.

src/Ng/SearchIndexFromCatalogCollector.cs
zhhyu: Yes, Uri is better than string.

src/NuGet.Indexing/Extraction/CatalogPackageMetadataExtraction.cs
```diff
@@ -55,18 +57,52 @@ private class Extractor
...
+            {
+                var uriBuilder = new UriBuilder(_galleryBaseAddress);
+                var packageId = JTokenToString(_catalog[MetadataConstants.IdPropertyName]);
+                var packageVersion = JTokenToString(_catalog[MetadataConstants.NormalizedVersionPropertyName]);
+
+                if (packageId == null)
```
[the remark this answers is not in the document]
zhhyu: Keep consistency for different cases here.

src/Catalog/Registration/RegistrationMakerCatalogItem.cs
```diff
@@ -161,22 +161,19 @@ private string GetLicenseUrl()
...
+            string packageId = _catalogItem.GetTriplesWithSubjectPredicate(subject, _catalogItem.CreateUriNode(Schema.Predicates.Id)).FirstOrDefault().Object.ToString();
+            string packageVersion = NuGetVersionUtility.NormalizeVersion(_catalogItem.GetTriplesWithSubjectPredicate(subject, _catalogItem.CreateUriNode(Schema.Predicates.Version)).FirstOrDefault().Object.ToString());
             Triple licenseExpression = _catalogItem.GetTriplesWithSubjectPredicate(subject, _catalogItem.CreateUriNode(Schema.Predicates.LicenseExpression)).FirstOrDefault();
             Triple licenseFile = _catalogItem.GetTriplesWithSubjectPredicate(subject, _catalogItem.CreateUriNode(Schema.Predicates.LicenseFile)).FirstOrDefault();
             Triple licenseUrl = _catalogItem.GetTriplesWithSubjectPredicate(subject, _catalogItem.CreateUriNode(Schema.Predicates.LicenseUrl)).FirstOrDefault();
 
-            if (_galleryBaseAddress != null)
+            if (_galleryBaseAddress != null &&
+                packageId != null &&
+                packageVersion != null &&
```
shishirx34: for strings `string.IsNullOrEmpty` is more robust. I prefer using `string.IsNullOrWhitespace` which covers all string cases.

src/Catalog/Registration/RegistrationMakerCatalogItem.cs
```diff
@@ -161,22 +161,19 @@ private string GetLicenseUrl()
...
+                ((licenseExpression != null && !string.IsNullOrWhiteSpace(licenseExpression.Object.ToString())) ||
+                 (licenseFile != null && !string.IsNullOrWhiteSpace(licenseFile.Object.ToString()))))
```
shishirx34: you can merge this into single call like 
```cs
!string.IsNullOrWhiteSpace(licenseExpression?.Object.ToString()) || !string.IsNullOrWhiteSpace(licenseFile?.Object.ToString())
```

src/Catalog/Registration/RegistrationMakerCatalogItem.cs
shishirx34: do this for all strings above.

src/NuGet.Indexing/Extraction/CatalogPackageMetadataExtraction.cs
```diff
@@ -55,18 +58,37 @@ private class Extractor
...
             }
 
+            private void AddLicenseUrl()
+            {
+                var packageId = JTokenToString(_catalog[MetadataConstants.IdPropertyName]);
+                var packageVersion = JTokenToString(_catalog[MetadataConstants.NormalizedVersionPropertyName]);
+
+                if (_galleryBaseAddress != null &&
+                    packageId != null &&
+                    packageVersion != null &&
```
shishirx34: why not use `!string.IsNullOrEmpty`?

src/Catalog/Helpers/LicenseHelper.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        /// <summary>
+        /// Generate the license url given package info and gallery base url.
+        /// </summary>
+        /// <param name="packageId">package Id</param>
+        /// <param name="packageVersion">package version</param>
+        /// <param name="galleryBaseAddress">url of gallery base address</param>
+        /// <returns>The url of license in gallery</returns>
+        public static string GetGalleryLicenseUrl(string packageId, string packageVersion, Uri galleryBaseAddress)
+        {
```
agr: We should have `null` checks for public method arguments.

tests/CatalogTests/Helpers/LicenseHelperTests.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace CatalogTests.Helpers
+{
+    public class LicenseHelperTests
+    {
+        [Theory]
+        [InlineData("testPackage", "1.0.0", "https://testnuget", "https://testnuget/packages/testPackage/1.0.0/license")]
+        [InlineData("testPackage", "1.0.0", "https://testnuget/", "https://testnuget/packages/testPackage/1.0.0/license")]
+        [InlineData("testPackage", "1.0.0", "https://testnuget//", "https://testnuget/packages/testPackage/1.0.0/license")]
+        public void GivenPackageIdAndVersionAndGalleryBaseUrl_ReturnsLicenseUrl(string packageId, string packageVersion, string galleryBaseAddress, string expectedLicenseUrl)
```
agr: I'd add a test that emphasizes the fact that no argument validation is done by the method besides `null` check. For example, pass some garbage as a version and check the result. Additionally, we need a test that checks the proper handling of characters outside of ASCII range.

agr: Example of such package: [测试更新包](https://www.nuget.org/packages/测试更新包/)

---
In reply to: [248849565](https://github.com/NuGet/NuGet.Services.Metadata/pull/470#discussion_r248849565) [](ancestors = 248849565)

src/NuGet.Indexing/NuGet.Indexing.csproj
```diff
@@ -199,6 +199,15 @@
       <Version>8.2.1</Version>
     </PackageReference>
   </ItemGroup>
+  <ItemGroup>
+    <WCFMetadata Include="Connected Services\" />
+  </ItemGroup>
+  <ItemGroup>
```
agr: VS added it on its own, I presume?.. Should we keep it?

PR #474, opened 2019-01-18
Add Application Insights to Azure Search service

joelverhagen:
Address https://github.com/nuget/nugetgallery/issues/6794

### DIFF
--- a/src/NuGet.Services.SearchService/App_Start/WebApiConfig.cs
+++ b/src/NuGet.Services.SearchService/App_Start/WebApiConfig.cs
@@ -12,9 +12,11 @@ using System.Web.Http.Dependencies;
 using Autofac;
 using Autofac.Extensions.DependencyInjection;
 using Autofac.Integration.WebApi;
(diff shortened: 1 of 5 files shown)


src/NuGet.Services.SearchService/App_Start/WebApiConfig.cs
```diff
@@ -99,6 +107,11 @@ private static IDependencyResolver GetDependencyResolver(HttpConfiguration confi
             builder.Populate(services);
             builder.AddAzureSearch();
 
+            var loggerConfiguration = LoggingSetup.CreateDefaultLoggerConfiguration(withConsoleLogger: false);
```
scottbommarito: Why did we replace `services.AddLogging` with this?
Does `AddLogging` not pick up the Application Insights instrumentation?

joelverhagen: AddLogging doesn't use the Serilog enriched LoggerFactory. Gallery does not use AddLogging and I am basically aligning with that

PR #475, opened 2019-01-18
Make EnsureInitializedAsync no-op if it's already initialized

joelverhagen:
Previously on start-up the first request would also try to load the auxiliary data. This is not necessary and caused troubling logs.

### DIFF
--- a/src/NuGet.Services.AzureSearch/SearchService/AuxiliaryDataCache.cs
+++ b/src/NuGet.Services.AzureSearch/SearchService/AuxiliaryDataCache.cs
@@ -27,20 +27,20 @@ namespace NuGet.Services.AzureSearch.SearchService
 
         public bool Initialized => _data != null;
 
-        public async Task InitializeAsync()
+        public async Task EnsureInitializedAsync()
         {
             if (!Initialized)
             {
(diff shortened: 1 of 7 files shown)


src/NuGet.Services.AzureSearch/SearchService/AuxiliaryDataCache.cs
```diff
@@ -27,20 +27,20 @@ public class AuxiliaryDataCache : IAuxiliaryDataCache
...
         }
 
         public async Task TryLoadAsync(CancellationToken token)
         {
-            await LoadAsync(TimeSpan.Zero, token);
+            await LoadAsync(TimeSpan.Zero, allowNoOp: false, token: token);
         }
 
-        private async Task LoadAsync(TimeSpan timeout, CancellationToken token)
+        private async Task LoadAsync(TimeSpan timeout, bool allowNoOp, CancellationToken token)
```
scottbommarito: Maybe `shouldReload` or something similar?
`allowNoOp` isn't super descriptive as to what it's allowing to no-op.

joelverhagen: Yeah, that's better

PR #477, opened 2019-01-19
Extract search text building into its own class

joelverhagen:
The query building code will be pretty big so it makes sense to be on its own.

Progress on https://github.com/NuGet/NuGetGallery/issues/6456.

### DIFF
--- a/src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
+++ b/src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
@@ -1,0 +1,33 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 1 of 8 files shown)


src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    public class SearchTextBuilder : ISearchTextBuilder
+    {
+        public string V2Search(V2SearchRequest request)
+        {
+            var query = request.Query;
+
+            if (request.LuceneQuery)
+            {
+                // TODO: convert a leading "id:" to "packageid:"
```
agr: Do we have work items to address those TODOs? Can you put links/IDs into the comment itself?

joelverhagen: Great point! I'll add this.

PR #479, opened 2019-02-06
Add instance name option to ng

joelverhagen:
This allows us to differentiate between multiple instances of the job running in the same environment but with different configuration.

The default value is the job name. I chose a non-null default so it is easy to `summarize` by it. 

Progress on https://github.com/NuGet/Engineering/issues/2087

### DIFF
--- a/src/Ng/Telemetry/JobNameTelemetryInitializer.cs
+++ b/src/Ng/Telemetry/JobNameTelemetryInitializer.cs
@@ -10,17 +10,21 @@ namespace Ng
     public class JobNameTelemetryInitializer : ITelemetryInitializer
     {
         private const string JobNameKey = "JobName";
+        private const string InstanceNameKey = "InstanceName";
 
         private readonly string _jobName;
+        private readonly string _instanceName;
 
-        public JobNameTelemetryInitializer(string jobName)
+        public JobNameTelemetryInitializer(string jobName, string instanceName)
         {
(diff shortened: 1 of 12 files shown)


src/Ng/Telemetry/JobNameTelemetryInitializer.cs
```diff
@@ -10,17 +10,24 @@ namespace Ng
...
+        public JobNameTelemetryInitializer(string jobName, string instanceName)
         {
             _jobName = jobName ?? throw new ArgumentNullException(nameof(jobName));
+            _instanceName = instanceName;
         }
 
         public void Initialize(ITelemetry telemetry)
         {
             telemetry.Context.Properties[JobNameKey] = _jobName;
+            if (_instanceName != null)
```
dtivel: Given that the default is non-null, do you need this `null` check?

joelverhagen: I hope to make this initializer class more general purpose. For example, some jobs must be singletons across and environment (unlike, say catalog2lucene which has an instance per region per environment).

joelverhagen: I've thought about this more and I think you are right. I was also favoring allowing null because there will be 90 days of telemetry that don't have this value since it's the old version of the job so anyone querying would need to think about the case where the key is missing. However this is not great since it's making the code support more states for a temporary situation.

I will require both.

PR #482, opened 2019-02-26
[Azure Search] Add initial query parsing

loic-sharma:
Parses NuGet queries and generates an Azure Search query. You can find documentation for Azure Search's query syntax [here](https://docs.microsoft.com/en-us/azure/search/query-lucene-syntax). You can find documentation for NuGet's query syntax [here](https://microsoft.sharepoint.com/teams/NuGet/_layouts/OneNote.aspx?id=%2Fteams%2FNuGet%2FTeam%2FNugetServer%2FNugetServerTeamNote&wd=target%28Specs.one%7C910181C5-8695-4A1C-9792-A56DBB86F1C3%2FSearch%20Service%7C58FE20C1-2314-434F-85B7-A8FE900AE3D2%2F%29).

This is missing:

* Partial match for the `id` query field. Tracked by https://github.com/NuGet/NuGetGallery/issues/6922
* Exact match for the `packageId`, `version`, and `owners` query fields. Tracked by https://github.com/NuGet/NuGetGallery/issues/6920
* Unicode handling: https://github.com/NuGet/NuGetGallery/issues/6939

This change has been deployed to DEV. You can test this here: https://nuget-dev-usnc-azuresearch-staging.azurewebsites.net/query?q=packageId:Newtonsoft.Json

Addresses https://github.com/NuGet/NuGetGallery/issues/6456


src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
```diff
@@ -23,11 +53,81 @@ public string V3Search(V3SearchRequest request)
...
+        }
+
+        private static IEnumerable<string> ProcessFieldValues(QueryField field, IEnumerable<string> values)
+        {
+            switch (field)
+            {
+                // The "packageId" and "owners" fields are case insensitive exact matches
+                case QueryField.PackageId:
+                case QueryField.Owner:
+                    return values.Select(v => v.ToLowerInvariant());
```
joelverhagen: My understanding is that you should not need to lowercase a search term because the analyzer associated with the search field will be used. See `analyzer` vs. `searchAnalyzer` vs. `indexAnalyzer`:
https://docs.microsoft.com/en-us/azure/search/search-what-is-an-index#components-of-an-index

My understanding is that we are using `analyzer` which uses the same analyzer for both searching and indexing.

src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
```diff
@@ -23,11 +53,81 @@ public string V3Search(V3SearchRequest request)
...
+                    return values.SelectMany(Utils.SplitTags);
+
+                // The "version" field is a case insensitive exact match on the normalized value
+                // Invalid version strings won't be normalized.
+                case QueryField.Version:
+                    return values.Select(value =>
+                    {
+                        if (NuGetVersion.TryParse(value, out var version))
+                        {
+                            return version.ToNormalizedString().ToLowerInvariant();
```
joelverhagen: Again, `ToLowerInvariant` should not be necessary. The normalization is necessary 😄

src/NuGet.Services.AzureSearch/SearchService/AzureSearchQuery.cs
```diff
@@ -0,0 +1,161 @@
...
+namespace NuGet.Services.AzureSearch.SearchService
+{
+    /// <summary>
+    /// The type used to generate Azure Search Service queries. Used by <see cref="SearchTextBuilder"/>.
+    /// Given the query "fieldA:value1 value2":
+    /// 
+    ///   * "value1" is a field-scoped value 
+    ///   * "value2" is a non-field scoped value
+    /// </summary>
+    internal class AzureSearchQuery
```
joelverhagen: Consider respecting https://docs.microsoft.com/en-us/azure/search/query-lucene-syntax#bkmk_querysizelimits.

src/NuGet.Services.AzureSearch/SearchService/AzureSearchQuery.cs
```diff
@@ -0,0 +1,161 @@
...
+        public override string ToString()
+        {
+            var result = new StringBuilder();
+
+            foreach (var fieldScopedTerm in _fieldScopedValues)
+            {
+                // Separate groups of field-scoped terms are AND'd together.
+                if (result.Length > 0)
+                {
+                    result.Append(" AND ");
```
joelverhagen: I would recommend choosing between `+` and `AND`.
https://docs.microsoft.com/en-us/azure/search/query-lucene-syntax#and-operator-and--or-

`+` seems a bit more idiomatic.

src/NuGet.Services.AzureSearch/SearchService/AzureSearchQuery.cs
```diff
@@ -0,0 +1,161 @@
...
+    /// <summary>
+    /// The type used to generate Azure Search Service queries. Used by <see cref="SearchTextBuilder"/>.
+    /// Given the query "fieldA:value1 value2":
+    /// 
+    ///   * "value1" is a field-scoped value 
+    ///   * "value2" is a non-field scoped value
+    /// </summary>
+    internal class AzureSearchQuery
+    {
+        private static readonly HashSet<char> SpecialCharacters = new HashSet<char>
```
joelverhagen: Link docs for this list

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
```diff
@@ -0,0 +1,170 @@
...
+                // Map of inputs to expected output
+                var data = new Dictionary<string, string>
+                {
+                    { "", "*" },
+                    { " ", "*" },
+
+                    // TODO: id and packageId query fields should map to different fields with different analyzers
+                    // See: https://github.com/NuGet/NuGetGallery/issues/6920
+                    // See: https://github.com/NuGet/NuGetGallery/issues/6922
+                    { "id:test", "packageId:test" },
```
joelverhagen: IIRC field names like "packageId" are case insensitive. Test non-canonical casings.

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
```diff
@@ -0,0 +1,170 @@
...
+                var actual = _target.V2Search(new V2SearchRequest { Query = input });
+
+                Assert.Equal(expected, actual);
+            }
+
+            // TODO: id and packageId query fields should map to different fields with different analyzers
+            // See: https://github.com/NuGet/NuGetGallery/issues/6920
+            // See: https://github.com/NuGet/NuGetGallery/issues/6922
+            [Theory]
+            [InlineData(false, "packageId:hello")]
```
joelverhagen: This should only happen if `luceneQuery` is `true` right?

src/NuGet.Services.AzureSearch/SearchService/AzureSearchQuery.cs
joelverhagen: Maybe `AzureSearchQueryBuilder` to emphasize the statefulness.

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
```diff
@@ -0,0 +1,170 @@
...
+                    { "version:1.0.0", "normalizedVersion:1.0.0" },
+                    { "title:hello", "title:hello" },
+                    { "description:hello", "description:hello" },
+                    { "tag:hi", "tags:hi" },
+                    { "tags:foo", "tags:foo" },
+                    { "author:bob", "authors:bob" },
+                    { "authors:billy", "authors:billy" },
+                    { "summary:test", "summary:test" },
+                    { "owner:goat", "owners:goat" },
+                    { "owners:nugget", "owners:nugget" },
```
joelverhagen: Every field and non-scoped terms should have assertions that we don't change the casing. In general I think we should depend on the analyzer to do as much as possible

src/NuGet.Services.AzureSearch/SearchService/AzureSearchQuery.cs
```diff
@@ -0,0 +1,161 @@
...
+                {
+                    result.Append(' ');
+                }
+
+                AppendEscapedString(result, values[i]);
+            }
+        }
+
+        private static void AppendEscapedString(StringBuilder result, string input)
+        {
```
joelverhagen: What the legacy service's behavior on surrogate pairs? This is kind of a nasty rabbit hole be we should know what the story is.

src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
```diff
@@ -23,11 +53,81 @@ public string V3Search(V3SearchRequest request)
...
+             }
+
+            return result;
+        }
+
+        private static IEnumerable<string> ProcessFieldValues(QueryField field, IEnumerable<string> values)
+        {
+            switch (field)
+            {
+                // The "packageId" and "owners" fields are case insensitive exact matches
```
joelverhagen: As mentioned elsewhere I think the analyzers should define the case sensitives --- but really will there be any field at all that is case sensitive?

src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
```diff
@@ -23,11 +53,81 @@ public string V3Search(V3SearchRequest request)
...
+                case QueryField.PackageId:
+                case QueryField.Owner:
+                    return values.Select(v => v.ToLowerInvariant());
+
+                // Expand tags by their delimiters
+                case QueryField.Tag:
+                    return values.SelectMany(Utils.SplitTags);
```
joelverhagen: Where does the de-duping happen? Could this produce duplicates?

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
```diff
@@ -0,0 +1,170 @@
...
+                    { "id packageId version title description tag author summary owner owners", "+(id packageId version title description tag author summary owner owners)" },
+                    
+                    // Quotes allow adjacent terms to be searched
+                    { @"""foo bar""", @"+(""foo bar"")" },
+                    { @"""foo bar"" baz", @"+(""foo bar"" baz)" },
+                    { @"id:""foo bar""", @"packageId:""foo bar""" },
+                    { @"id:""a b"" c id:d packageId:e f", @"+packageId:(""a b"" d e) c f" },
+
+                    // Duplicate search terms on the same query field are folded
+                    { "a a", "+(a)" },
```
joelverhagen: My knowledge about non-scope Lucene queries is not strong enough. Do you really need the `+` when there is only one term?

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
```diff
@@ -0,0 +1,170 @@
...
+                    { "title:a title:a", "title:a" },
+                    { "tag:a tags:a", "tags:a" },
+
+                    // Single word query terms are unquoted.
+                    { @"""a""", "+(a)" },
```
joelverhagen: 👍

src/NuGet.Services.AzureSearch/SearchService/AzureSearchQuery.cs
joelverhagen: 😈 you can punt this till later but they shouldn't like cause an exception or give a backdoor

src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
loic-sharma: [The analyzer for exact matches](http://lucene.apache.org/core/7_3_1/analyzers-common/org/apache/lucene/analysis/core/KeywordAnalyzer.html) is case sensitive. To work around this, you can either:

1. Use [a filter to lowercase terms](https://lucene.apache.org/core/4_10_3/analyzers-common/org/apache/lucene/analysis/core/LowerCaseFilter.html)
2. Lowercase the field's content in the index and the search query.

The first approach would require building a custom Azure Search Analyzer that chains the lowercase filter and the keyword analyzer. Both approaches results in a new field that is no longer displayable as the content has lost its casing. The second approach is easier to implement, which is why used that here. Does that make sense?

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
loic-sharma: That's correct. I'm unit testing the current behavior, but this will be addressed by the TODO issues that are linked in the comment above.

src/NuGet.Services.AzureSearch/SearchService/AzureSearchQuery.cs
```diff
@@ -0,0 +1,161 @@
...
+using System.Linq;
+using System.Text;
+
+namespace NuGet.Services.AzureSearch.SearchService
+{
+    /// <summary>
+    /// The type used to generate Azure Search Service queries. Used by <see cref="SearchTextBuilder"/>.
+    /// Given the query "fieldA:value1 value2":
+    /// 
+    ///   * "value1" is a field-scoped value
```
agr: XML documentation has [markup](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/list) for lists.

src/NuGet.Services.AzureSearch/SearchService/AzureSearchQuery.cs
```diff
@@ -0,0 +1,161 @@
...
+
+                // If there are both non-field-scoped terms and field-scoped terms, only the field-scoped terms
+                // are required.
+                if (_nonFieldScopedValues.Any())
+                {
+                    result.Append('+');
+                }
+
+                result.Append(fieldScopedTerm.Key);
+                result.Append(':');
```
agr: `Append` returns `StringBuilder`, so you can chain calls, or just use `AppendFormat`.

src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
loic-sharma: Good catch, this is fixed now

src/NuGet.Services.AzureSearch/SearchService/AzureSearchQuery.cs
loic-sharma: Renamed

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
loic-sharma: Yes, the spec explicitly called this behavior out so I made it explicit in the generated queries as well. I agree it's a little ugly, so I removed the + when from non-field-scoped terms

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
loic-sharma: Added a test

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
loic-sharma: Added tests

src/NuGet.Services.AzureSearch/SearchService/AzureSearchQuery.cs
loic-sharma: Added with unit tests

src/NuGet.Services.AzureSearch/SearchService/AzureSearchQuery.cs
loic-sharma: Moved everything to the required operator `+`

src/NuGet.Services.AzureSearch/SearchService/AzureSearchQuery.cs
loic-sharma: Added

src/NuGet.Services.AzureSearch/SearchService/AzureSearchQuery.cs
loic-sharma: I'd rather avoid verbose XML syntax as we're not generating docs from these comments. Let me know if you feel strongly about this.

src/NuGet.Services.AzureSearch/SearchService/AzureSearchQuery.cs
loic-sharma: Oh cool, I didn't know this. I toyed around with chaining, but I preferred how this code looked without chaining

src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
loic-sharma: I've reverted my changes for case insensitive exact match semantics, this will re-added by https://github.com/NuGet/NuGetGallery/issues/6920

src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
loic-sharma: I've reverted my changes for case insensitive exact match semantics, this will re-added by https://github.com/NuGet/NuGetGallery/issues/6920

src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
loic-sharma: I've reverted my changes for case insensitive exact match semantics, this will re-added by https://github.com/NuGet/NuGetGallery/issues/6920 

(But no, there's no case sensitive fields)

src/NuGet.Services.AzureSearch/SearchService/AzureSearchQuery.cs
loic-sharma: Moved to separate issue: https://github.com/NuGet/NuGetGallery/issues/6939

loic-sharma:
🔔 All feedback should be addressed

loic-sharma:
Parses NuGet queries and generates an Azure Search query. Addresses https://github.com/NuGet/NuGetGallery/issues/6456

PR #483, opened 2019-02-26
[HotFix] Make concurrent batch count as tunable parameter

shishirx34:
The C2R job keeps on hanging. One of the hypothesis is that may be disable parallelization in catalog 2 registration job and only process single package ID at a time might give us more details and make progress on China job.


src/Ng/Jobs/Catalog2RegistrationJob.cs
```diff
@@ -118,6 +120,20 @@ protected override void Init(IDictionary<string, string> arguments, Cancellation
             TelemetryService.GlobalDimensions[TelemetryConstants.Destination] = _destination?.AbsoluteUri;
         }
 
+        private int MaxConcurrentBatches(int argBatches)
```
scottbommarito: IMO this logic should be in `RegistrationCollector`'s constructor instead so that no caller can pass in an invalid number of batches.

shishirx34: I agree.. but other places we do it when reading arguments, I would keep it consistent.

PR #484, opened 2019-02-26
[HotFix] Apply container blob request options to blobs

joelverhagen:
Address https://github.com/NuGet/Engineering/issues/2152

### DIFF
--- a/src/Catalog/Persistence/AzureStorage.cs
+++ b/src/Catalog/Persistence/AzureStorage.cs
@@ -103,7 +103,7 @@ namespace NuGet.Services.Metadata.Catalog.Persistence
             cancellationToken.ThrowIfCancellationRequested();
 
             string blobName = GetName(resourceUri);
-            CloudBlockBlob blob = _directory.GetBlockBlobReference(blobName);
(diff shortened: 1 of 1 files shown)


src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -410,5 +415,19 @@ public async Task<bool> HasPropertiesAsync(Uri blobUri, string contentType, stri
...
+        private CloudBlockBlob GetBlockBlobReference(string blobName)
+        {
+            var blob = _directory.GetBlockBlobReference(blobName);
+
+            ApplyBlobRequestOptions(blob, _directory);
+
+            return blob;
+        }
+
+        private static void ApplyBlobRequestOptions(CloudBlockBlob blob, CloudBlobDirectory directory)
```
dtivel: You could simplify this to:

    private void ApplyBlobRequestOptions(CloudBlockBlob blob)

...and just use `_directory` in the implementation, similar to the method above.

joelverhagen: Good point. Fixed.

PR #486, opened 2019-02-28
[HotFix] Introduce retry around storage LoadAsync, SaveAsync, and DeleteAsync

joelverhagen:
These were the methods that were failing intermittently on EA -> Mooncake storage.


src/Catalog/Retry.cs
```diff
@@ -0,0 +1,58 @@
...
+            for (int currentAttempt = 0; currentAttempt < maxAttempts; currentAttempt++)
+            {
+                try
+                {
+                    runLogic();
+                    return;
+                }
+                catch (Exception e)
+                {
+                    if (currentAttempt < maxAttempts && shouldRetry(e))
```
loic-sharma: These could be `when` statements on the `catch`

joelverhagen: Sure. I just mimicked the existing synchronous implementation but I'm okay with making this change to both.

loic-sharma: Ah I didn't realize this file was moved. It's fine either ways 😄

src/NuGet.Indexing/Retry.cs
```diff
@@ -1,34 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
```
shishirx34: Is it fair to assume this change won't break any other high level dependencies/not used at all? or we are going to treat this as a breaking change?

joelverhagen: `NuGet.Indexing` depends on `NuGet.Services.Metadata.Catalog` so the class is still available to the previous users.

src/Catalog/Retry.cs
```diff
@@ -0,0 +1,58 @@
...
+                    }
+                    else
+                    {
+                        throw;
+                    }
+                }
+            }
+        }
+
+        public static async Task IncrementalAsync(Func<Task> runLogicAsync, Func<Exception, bool> shouldRetry, int maxAttempts, TimeSpan waitIncrement)
```
dtivel: Is it worth logging a custom metric for attempts > 1?  That way we could easily understand how well this solution works and also how frequently retries are necessary over time.

src/Catalog/Retry.cs
joelverhagen: Done.

src/Catalog/Retry.cs
joelverhagen: Good idea. Done.

joelverhagen:
I am going to hold off on this hotfix until we see another problem running catalog2registration from EA.

joelverhagen:
Closing this PR. Will re-open if needed.

PR #489, opened 2019-03-07
[Azure Search] Add autocomplete

loic-sharma:
For the package id autocomplete, this is a breaking change from the existing search service. Instead of using ngrams, this implementation of autocomplete is a strict prefix search. The previous autocomplete implementation has some pretty strange characteristics, hence this breaking change. Feel free to reach out to me if you have any questions.

I've verified that the last 90 days' autocomplete data indicate that prefix search is acceptable. You can find this data [here](https://microsoft.sharepoint.com/teams/NuGet/Team/NugetServer/NugetServerTeamNote/Specs.one#Autocomplete&section-id={910181C5-8695-4A1C-9792-A56DBB86F1C3}&page-id={C24C80A7-4F9E-4560-A33F-F32140816202}&object-id={98BE5690-465E-0CDF-141D-9162382D7E9F}&1F).

Addresses https://github.com/nuget/nugetgallery/issues/6451


src/NuGet.Services.AzureSearch/SearchService/AzureSearchService.cs
```diff
@@ -58,6 +58,23 @@ public async Task<V3SearchResponse> V3SearchAsync(V3SearchRequest request)
                 result.Duration);
         }
 
+        public async Task<AutocompleteResponse> AutocompleteAsync(AutocompleteRequest request)
+        {
+            var text = _textBuilder.Autocomplete(request);
+            var parameters = _parametersBuilder.Autocomplete(request);
+
+            var result = await Measure.DurationWithValueAsync(() => _searchIndex.Documents.SearchAsync<SearchDocument.Full>(
```
agr: `result` -> `duration`?

src/NuGet.Services.AzureSearch/SearchService/SearchResponseBuilder.cs
```diff
@@ -110,6 +111,50 @@ public class SearchResponseBuilder : ISearchResponseBuilder
...
+
+            List<string> data;
+            switch (request.Type)
+            {
+                case AutocompleteRequestType.PackageIds:
+                    data = results.Select(x => x.Document.PackageId).ToList();
+                    break;
+
+                case AutocompleteRequestType.PackageVersions:
+                    data = results.SelectMany(x => x.Document.Versions).ToList();
```
joelverhagen: nit: I would throw if there is more than one result and then should do `[0]`. `SelectMany` across multiple IDs will never be the desired behavior AFAIK.

src/NuGet.Services.AzureSearch/SearchService/SearchResponseBuilder.cs
```diff
@@ -110,6 +111,50 @@ public class SearchResponseBuilder : ISearchResponseBuilder
...
+                TotalHits = result.Count.Value,
+                Data = data,
+                Debug = DebugInformation.CreateOrNull(
+                    request,
+                    _options.Value.SearchIndexName,
+                    parameters,
+                    text,
+                    result,
+                    duration,
+                    AuxiliaryData.Metadata),
```
joelverhagen: Hm... sending back aux metadata is a little weird in this case since it is unused in this code path. Maybe set it to null here.

src/NuGet.Services.AzureSearch/SearchService/AzureSearchQueryBuilder.cs
```diff
@@ -38,171 +38,175 @@ internal class AzureSearchQueryBuilder
...
+
+                if (prefixSearch)
+                {
+                    _result.Append('*');
+                }
             }
         }
 
-        private static void AppendEscapedString(StringBuilder result, string input)
+        private void AppendEscapedString(string input, bool quoteWhiteSpace = true)
```
joelverhagen: is it necessary have default parameters here? IMO it improves GitHub readability to avoid this and name parameters
(nit)

src/NuGet.Services.SearchService/Controllers/SearchController.cs
```diff
@@ -107,6 +107,37 @@ public async Task<HttpResponseMessage> GetStatusAsync(HttpRequestMessage request
...
+            bool prerelease = false,
+            string semVerLevel = null,
+            string q = null,
+            string id = null,
+            bool debug = false)
+        {
+            await EnsureInitializedAsync();
+
+            // If only "id" is provided, find package versions. Otherwise, find package Ids.
+            var type = (q != null || id == null)
```
joelverhagen: Is this behavior consistent with existing implementation? e.g.

- if both are provided with real values
- if either is provided with empty string, e.g. `q=&id=newtonsoft.json`
- if neither is provided

src/NuGet.Services.AzureSearch/SearchService/AzureSearchQueryBuilder.cs
```diff
@@ -38,171 +38,175 @@ internal class AzureSearchQueryBuilder
...
         public AzureSearchQueryBuilder()
         {
-            _nonFieldScopedValues = new List<string>();
-            _fieldScopedValues = new Dictionary<string, List<string>>();
+            _result = new StringBuilder();
+            _clauses = 0;
         }
 
-        public void AddNonFieldScopedValues(IEnumerable<string> values)
+        public void AddTerms(IReadOnlyList<string> terms)
```
cristinamanum: I think adding comments for the methods will be valuable.  Is there a specific AzureSearchQuery syntax that we try to follow? If yes, a link to docs would be useful . 

Is it allowed to call AddTerms and AddScopedTerms on the same instance or invoke multiple times the same method?

src/NuGet.Services.AzureSearch/SearchService/SearchResponseBuilder.cs
cristinamanum: Will the data need to be ordered?

src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
```diff
@@ -53,48 +53,73 @@ public string V3Search(V3SearchRequest request)
...
+            {
+                return MatchAllDocumentsQuery;
+            }
+
+            // Generate a query on package ids. This will be a prefix search
+            // if we are autocompleting package ids.
+            var builder = new AzureSearchQueryBuilder();
+
+            builder.AddScopedTerms(
+                IndexFields.PackageId,
```
cristinamanum: AddScopedTerms can throw, what will be the experience in this case?

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
```diff
@@ -102,6 +102,46 @@ public void ThrowsWhenTermIsTooBig(string query)
             }
         }
 
+        public class Autocomplete : FactsBase
```
cristinamanum: I did not see tests for AzureSearchQueryBuilder. In case that there are not, IMO i think it would be good to add a set of tests for that area.

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
```diff
@@ -102,6 +102,46 @@ public void ThrowsWhenTermIsTooBig(string query)
...
+            {
+                var request = new AutocompleteRequest
+                {
+                    Query = input,
+                    Type = AutocompleteRequestType.PackageVersions
+                };
+
+                var result = _target.Autocomplete(request);
+
+                Assert.Equal(expected, result);
```
cristinamanum: The expected result does not have version. Should not be ?

src/NuGet.Services.AzureSearch/SearchService/SearchParametersBuilder.cs
```diff
@@ -13,6 +13,9 @@ public class SearchParametersBuilder : ISearchParametersBuilder
         public const int DefaultTake = 20;
         private const int MaximumTake = 1000;
 
+        private static readonly List<string> PackageIdsAutocompleteSelect = new List<string> { IndexFields.PackageId };
```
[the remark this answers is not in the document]
loic-sharma: Yup, the SDK accepts an `IList` which isn't compatible with `IReadOnlyList`

src/NuGet.Services.AzureSearch/SearchService/SearchResponseBuilder.cs
loic-sharma: @joelverhagen I'll check that the results object has only one document as expected. `SelectMany` handles the case when there's no results succinctly, so I'll keep that as it is.

@cristinamanum The data does not need to be ordered, as per [the documentation](https://docs.microsoft.com/en-us/nuget/api/search-autocomplete-service-resource#response-1). However, I believe that versions will be ordered due to implementation details.

src/NuGet.Services.SearchService/Controllers/SearchController.cs
loic-sharma: Yup:

* If both are provided with real values, we return package ids. [Link](https://api-v2v3search-1.nuget.org/autocomplete?q=newtonsoft.json&id=newtonsoft.json)
* If both are provided but "id" is empty, we return package ids. [Link](https://api-v2v3search-1.nuget.org/autocomplete?q=Test&id=)
* If both are provided but "q" is empty, we return package ids. [Link](https://api-v2v3search-1.nuget.org/autocomplete?q=&id=Test)
* If neither are provided, we return package ids. [Link](https://api-v2v3search-1.nuget.org/autocomplete)

src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
loic-sharma: That's a very good question. This will throw an uncaught exception and result in a HTTP 500 Internal Error. I'll fix this

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
loic-sharma: This tests the Azure Search query we generate, but doesn't test the result of running the Azure Search query. Please let me know if I misunderstood your question!

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
loic-sharma: Good question. `AzureSearchQueryBuilder` should only be used by `SearchTextBuilder`, it is effectively only an implementation detail. All the unit tests for `AzureSearchQueryBuilder` are covered by `SearchTextBuilder`'s unit tests.

I'll make this clearer by making `AzureSearchQueryBuilder` a nested type on `SearchTextBuilder`.

src/NuGet.Services.AzureSearch/SearchService/SearchResponseBuilder.cs
loic-sharma: An exception is thrown now if the results are malformed

src/NuGet.Services.AzureSearch/SearchService/SearchResponseBuilder.cs
loic-sharma: Fair enough. Removed

src/NuGet.Services.AzureSearch/SearchService/AzureSearchQueryBuilder.cs
loic-sharma: Made parameters required

src/NuGet.Services.AzureSearch/SearchService/AzureSearchQueryBuilder.cs
loic-sharma: I've improved the documentation in this class. Please let me know if you have any additional suggestions!

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
loic-sharma: `AzureSearchQueryBuilder` is now nested inside of `SearchTextBuilder`.

src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
loic-sharma: This should be fixed now

src/NuGet.Services.SearchService/Controllers/SearchController.cs
```diff
@@ -80,11 +81,19 @@ public async Task<HttpResponseMessage> GetStatusAsync(HttpRequestMessage request
                 ShowDebug = debug,
             };
 
-            return await _searchService.V2SearchAsync(request);
+            try
+            {
+                return Json(await _searchService.V2SearchAsync(request));
```
joelverhagen: Maybe add a helper so we have the try/catch defined once and call the helper in these three places

loic-sharma: Good suggestion, cleaned up the code quite nicely

PR #490, opened 2019-03-14
Remove System.Random usage.

cristinamanum:
Change System.Random with RNGCryptoServiceProvider also remove the use of ticks as a randomizer.


src/Catalog/ValidatePackageHashHandler.cs
```diff
@@ -52,7 +52,6 @@ public async Task ProcessPackageAsync(CatalogIndexEntry packageEntry, ICloudBloc
             using (var packageStream = await _httpClient.GetStreamAsync(blob.Uri))
             {
                 var hashBytes = hashAlgorithm.ComputeHash(packageStream);
-
```
joelverhagen: Unnecessary change

PR #498, opened 2019-04-08
Add OwnerDataClient to read/write the owner information from storage

joelverhagen:

### DIFF
--- a/src/NuGet.Services.AzureSearch/Owners2AzureSearch/IOwnerDataClient.cs
+++ b/src/NuGet.Services.AzureSearch/Owners2AzureSearch/IOwnerDataClient.cs
@@ -1,0 +1,41 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using NuGetGallery;
+
+namespace NuGet.Services.AzureSearch.Owners2AzureSearch
+{
+    /// <summary>
+    /// The purpose of this interface is allow reading and writing owner information from storage. The Catalog2Owners
+    /// job does a comparison of latest owner data from the database with a snapshot of information stored in Azure
+    /// Blob Storage. This interface handles the reading and writing of that snapshot from storage.
+    /// </summary>
+    public interface IOwnerDataClient
+    {
+        /// <summary>
+        /// Read all of the latest indexed owners from storage. Also, return the current etag to allow optimistic
+        /// concurrency checks on the writing of the file. The returned dictionary has a key which is the package ID
+        /// and a value which is the owners of that package ID. The dictionary and the sets are case-insensitive.
+        /// </summary>
+        Task<ResultAndAccessCondition<SortedDictionary<string, SortedSet<string>>>> ReadLatestIndexedAsync();
+
+        /// <summary>
+        /// Replace the existing latest indexed owners file (i.e. "owners.v2.json" file).
+        /// </summary>
+        /// <param name="newData">The new data to be serialized into storage.</param>
+        /// <param name="accessCondition">The access condition (i.e. etag) to use during the upload.</param>
+        Task ReplaceLatestIndexedAsync(
+            SortedDictionary<string, SortedSet<string>> newData,
+            IAccessCondition accessCondition);
+
+        /// <summary>
+        /// Write a list of owners to storage. The file name that will be used in storage will be a timestamp so
+        /// subsequent calls should not conflict.
+        /// </summary>
+        /// <param name="packageIds">A non-empty list of package IDs that had owner changes.</param>
+        Task UploadChangeHistoryAsync(IReadOnlyList<string> packageIds);
+    }
+}
+
--- a/src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerDataClient.cs
+++ b/src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerDataClient.cs
@@ -1,0 +1,173 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Net;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Microsoft.WindowsAzure.Storage;
+using Newtonsoft.Json;
+using NuGetGallery;
+
+namespace NuGet.Services.AzureSearch.Owners2AzureSearch
+{
+    public class OwnerDataClient : IOwnerDataClient
+    {
+        private static readonly JsonSerializer Serializer = new JsonSerializer();
+
+        private readonly ICloudBlobClient _cloudBlobClient;
+        private readonly IOptionsSnapshot<AzureSearchJobConfiguration> _options;
+        private readonly ILogger<VersionListDataClient> _logger;
+        private readonly Lazy<ICloudBlobContainer> _lazyContainer;
+
+        public OwnerDataClient(
+            ICloudBlobClient cloudBlobClient,
+            IOptionsSnapshot<AzureSearchJobConfiguration> options,
+            ILogger<VersionListDataClient> logger)
+        {
+            _cloudBlobClient = cloudBlobClient ?? throw new ArgumentNullException(nameof(cloudBlobClient));
+            _options = options ?? throw new ArgumentNullException(nameof(cloudBlobClient));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+            _lazyContainer = new Lazy<ICloudBlobContainer>(
+                () => _cloudBlobClient.GetContainerReference(_options.Value.StorageContainer));
+        }
+
+        private ICloudBlobContainer Container => _lazyContainer.Value;
+
+        public async Task<ResultAndAccessCondition<SortedDictionary<string, SortedSet<string>>>> ReadLatestIndexedAsync()
+        {
+            var blobName = GetLatestIndexedBlobName();
+            var blobReference = Container.GetBlobReference(blobName);
+
+            _logger.LogInformation("Reading the latest indexed owners from {BlobName}.", blobName);
+
+            var builder = new PackageIdToOwnersBuilder(_logger);
+            IAccessCondition accessCondition;
+            try
+            {
+                using (var stream = await blobReference.OpenReadAsync(AccessCondition.GenerateEmptyCondition()))
+                {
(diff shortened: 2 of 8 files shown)


src/NuGet.Services.AzureSearch/Owners2AzureSearch/IOwnerDataClient.cs
```diff
@@ -0,0 +1,19 @@
...
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using NuGetGallery;
+
+namespace NuGet.Services.AzureSearch.Owners2AzureSearch
+{
+    public interface IOwnerDataClient
+    {
+        Task<ResultAndAccessCondition<SortedDictionary<string, SortedSet<string>>>> ReadLatestIndexedAsync();
```
loic-sharma: Could you add comments about the purpose of the owners data and when the data is updated? Ideally someone could get a gist of today's design review from the comments on this class.

src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerDataClient.cs
```diff
@@ -0,0 +1,171 @@
...
+            ICloudBlobClient cloudBlobClient,
+            IOptionsSnapshot<AzureSearchJobConfiguration> options,
+            ILogger<VersionListDataClient> logger)
+        {
+            _cloudBlobClient = cloudBlobClient ?? throw new ArgumentNullException(nameof(cloudBlobClient));
+            _options = options ?? throw new ArgumentNullException(nameof(cloudBlobClient));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+            _lazyContainer = new Lazy<ICloudBlobContainer>(
+                () => _cloudBlobClient.GetContainerReference(_options.Value.StorageContainer));
```
loic-sharma: AFAIK `GetContainerReference` just builds an object. Why does it need to be wrapped with a lazy?

src/NuGet.Services.AzureSearch/Owners2AzureSearch/IOwnerDataClient.cs
```diff
@@ -0,0 +1,19 @@
...
+        Task ReplaceLatestIndexedAsync(
+            SortedDictionary<string, SortedSet<string>> newData,
+            IAccessCondition accessCondition);
+        Task UploadChangeHistoryAsync(IEnumerable<string> packageIds);
```
loic-sharma: Can the `IEnumerable` be an `IReadOnlyList`? I've been bitten too many times by LINQ :)

src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerDataClient.cs
```diff
@@ -0,0 +1,171 @@
...
+                accessCondition = AccessConditionWrapper.GenerateIfNotExistsCondition();
+                _logger.LogInformation("The blob {BlobName} does not exist.", blobName);
+            }
+
+            return new ResultAndAccessCondition<SortedDictionary<string, SortedSet<string>>>(
+                builder.GetResult(),
+                accessCondition);
+        }
+
+        public async Task UploadChangeHistoryAsync(IEnumerable<string> packageIds)
```
loic-sharma: Is it the caller's responsibility to ensure that `packageIds` is non-empty?

src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerDataClient.cs
```diff
@@ -0,0 +1,171 @@
...
+                Guard.Assert(jsonReader.Read(), "There should be a second token.");
+                while (jsonReader.TokenType == JsonToken.PropertyName)
+                {
+                    var id = (string)jsonReader.Value;
+
+                    Guard.Assert(jsonReader.Read(), "There should be a token after the property name.");
+                    Guard.Assert(jsonReader.TokenType == JsonToken.StartArray, "The token after the property name should be the start of an object.");
+
+                    var owners = Serializer.Deserialize<List<string>>(jsonReader);
+                    foreach (var owner in owners)
```
loic-sharma: Thoughts on using `Add(string id, IEnumerable<string> owners)` here to dedup the foreach?

src/NuGet.Services.AzureSearch/Owners2AzureSearch/IOwnerDataClient.cs
loic-sharma: Same comment with line 25 of `PackageIdToOwnersBuilder.cs`

src/NuGet.Services.AzureSearch/Owners2AzureSearch/PackageIdToOwnersBuilder.cs
```diff
@@ -0,0 +1,70 @@
...
+        public void Add(string id, string username)
+        {
+            _addCount++;
+            if (_addCount % 10000 == 0)
+            {
+                _logger.LogInformation("{AddCount} ownership records have been added so far.", _addCount);
+            }
+
+            // Use a single instance of each username string.
+            if (!_usernameToUsername.TryGetValue(username, out var existingUsername))
```
loic-sharma: Why do you use a dictionary here instead of `string.Intern`?

src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerDataClient.cs
joelverhagen: Pattern of existing code. Does not really on the wrapper implementation or underlying implementation to be lightweight.

src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerDataClient.cs
joelverhagen: Throwing is a good idea.

src/NuGet.Services.AzureSearch/Owners2AzureSearch/PackageIdToOwnersBuilder.cs
joelverhagen: `string.Intern` never gets GC'd. First thing I learned on NuGet team thanks to @yishaigalatzer 👍.

src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerDataClient.cs
joelverhagen: Yup, sounds good.

src/NuGet.Services.AzureSearch/Owners2AzureSearch/IOwnerDataClient.cs
joelverhagen: `IEnumerable` is not LINQ per se and should be acceptable as a method parameter. It's the implementor's responsibility to enumerate it an appropriate number of times. In 99% of our code there is a single implementation of these interfaces so it's easy to get right in the one place it needs to. But sure, I don't care much so I'll switch it.

src/NuGet.Services.AzureSearch/Owners2AzureSearch/IOwnerDataClient.cs
joelverhagen: Yeah, sounds good.

src/NuGet.Services.AzureSearch/Owners2AzureSearch/PackageIdToOwnersBuilder.cs
loic-sharma: I’d be more concerned by the global lock used in `string.Intern`’s implementation. Regardless, the set of owners on nuget.org is relatively stable :)

joelverhagen: Concerning the lock, that's their implementation detail and would not be a problem given we're doing this pseudo-interning serially.

Concerning the set of owners being stable, this is true but the job does not need to hold onto this memory why it is sleeping. Current design is to run every 24 hours so a lot of the time the memory will be used when it is not needed.

loic-sharma: Ah that's a good point 👍

src/NuGet.Services.AzureSearch/Owners2AzureSearch/IOwnerDataClient.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// <param name="accessCondition">The access condition (i.e. etag) to use during the upload.</param>
+        Task ReplaceLatestIndexedAsync(
+            SortedDictionary<string, SortedSet<string>> newData,
+            IAccessCondition accessCondition);
+
+        /// <summary>
+        /// Write a list of owners to storage. The file name that will be used in storage will be a timestamp so
+        /// subsequent calls should not conflict.
+        /// </summary>
+        Task UploadChangeHistoryAsync(IReadOnlyList<string> packageIds);
```
loic-sharma: `packageIds` is missing a `<param ...`. I'd mention it cannot be empty

src/NuGet.Services.AzureSearch/Owners2AzureSearch/IOwnerDataClient.cs
loic-sharma: Looks great! 👌

src/NuGet.Services.AzureSearch/Owners2AzureSearch/IOwnerDataClient.cs
joelverhagen: Sounds good.

src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerDataClient.cs
```diff
@@ -0,0 +1,173 @@
...
+                _logger.LogInformation("The blob {BlobName} does not exist.", blobName);
+            }
+
+            return new ResultAndAccessCondition<SortedDictionary<string, SortedSet<string>>>(
+                builder.GetResult(),
+                accessCondition);
+        }
+
+        public async Task UploadChangeHistoryAsync(IReadOnlyList<string> packageIds)
+        {
```
zhhyu: Null check for packageIds?

joelverhagen: In general I avoid null checks all of the place in our methods. You'll notice 90% of our code doesn't null check method parameters. I think this is acceptable given our target consumer ... us!

It's a trade-off.

If we were releasing these .NET APIs publicly as packages and encouraging people to use them for their own projects, then I think we would want to be more rigorous about this.

We can maybe talk in the future within the team about adopting:
https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#nullable-reference-types

PR #499, opened 2019-04-09
Add OwnerSetComparer which detects ownership changes

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/6475

### DIFF
--- a/src/NuGet.Services.AzureSearch/Owners2AzureSearch/IOwnerSetComparer.cs
+++ b/src/NuGet.Services.AzureSearch/Owners2AzureSearch/IOwnerSetComparer.cs
@@ -1,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace NuGet.Services.AzureSearch.Owners2AzureSearch
+{
+    public interface IOwnerSetComparer
+    {
+        SortedDictionary<string, string[]> Compare(
+            SortedDictionary<string, SortedSet<string>> oldData,
+            SortedDictionary<string, SortedSet<string>> newData);
+    }
+}
--- a/src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerSetComparer.cs
+++ b/src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerSetComparer.cs
@@ -1,0 +1,101 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.Logging;
+
+namespace NuGet.Services.AzureSearch.Owners2AzureSearch
+{
+    public class OwnerSetComparer : IOwnerSetComparer
+    {
+        private static readonly string[] EmptyStringArray = new string[0];
+
+        private readonly ILogger _logger;
+
+        public OwnerSetComparer(ILogger logger)
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public SortedDictionary<string, string[]> Compare(
+            SortedDictionary<string, SortedSet<string>> oldData,
+            SortedDictionary<string, SortedSet<string>> newData)
+        {
+            if (oldData.Comparer != StringComparer.OrdinalIgnoreCase)
+            {
+                throw new ArgumentException("The old data should have a case-insensitive comparer.", nameof(oldData));
+            }
+
+            if (newData.Comparer != StringComparer.OrdinalIgnoreCase)
+            {
+                throw new ArgumentException("The new data should have a case-insensitive comparer.", nameof(newData));
+            }
+
+            // We use a very simplistic algorithm here. Perform one pass on the new data to find the added or changed
+            // package IDs. Then perform a second pass on the old data to find removed package IDs. We can optimize
+            // this later if necessary.
+            //
+            // We emit all of the usernames when a package ID's owners have changed instead of the delta. This is
+            // because Azure Search does not have a way to append or remove a specific item from a field that is an
+            // array. The entire new array needs to be provided.
+            var result = new SortedDictionary<string, string[]>(StringComparer.OrdinalIgnoreCase);
+
+            // First pass: find added or changed sets.
+            foreach (var pair in newData)
+            {
+                var id = pair.Key;
+                var newOwners = pair.Value;
+                if (!oldData.TryGetValue(id, out var oldOwners))
+                {
+                    // ADDED: The package ID does not exist in the old data, which means the package ID was added.
+                    result.Add(id, newOwners.ToArray());
+                    _logger.LogInformation(
+                        "The package ID {ID} has been added, with {AddedCount} owners.",
+                        id,
+                        newOwners.Count);
+                }
+                else
+                {
+                    // The package ID exists in the old data. We need to check if the owner set has changed. Perform
+                    // an ordinal comparison to allow username case changes to flow through.
+                    var removedUsernames = oldOwners.Except(newOwners, StringComparer.Ordinal).ToList();
+                    var addedUsernames = newOwners.Except(oldOwners, StringComparer.Ordinal).ToList();
+
+                    if (removedUsernames.Any() || addedUsernames.Any())
+                    {
+                        // CHANGED: The username set has changed.
+                        result.Add(id, newOwners.ToArray());
+                        _logger.LogInformation(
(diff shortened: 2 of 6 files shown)


src/NuGet.Services.AzureSearch/Owners2AzureSearch/IOwnerSetComparer.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace NuGet.Services.AzureSearch.Owners2AzureSearch
+{
+    public interface IOwnerSetComparer
+    {
+        SortedDictionary<string, string[]> Compare(
```
skofman1: comments?

src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerSetComparer.cs
```diff
@@ -0,0 +1,91 @@
...
+            // because Azure Search does not have a way to append or remove a specific item from a field that is an
+            // array. The entire new array needs to be provided.
+            var result = new SortedDictionary<string, string[]>(StringComparer.OrdinalIgnoreCase);
+
+            // First pass: find added or changed sets.
+            foreach (var pair in newData)
+            {
+                var id = pair.Key;
+                var newOwners = pair.Value;
+                if (!oldData.TryGetValue(id, out var oldOwners))
```
skofman1: >id [](start = 41, length = 2)

are the ids normalized before this method? I'm thinking about cases when we change capitalization of id by customer request.

src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerSetComparer.cs
```diff
@@ -0,0 +1,91 @@
...
+                {
+                    // The package ID exists in the old data. We need to check if the owner set has changed. Perform
+                    // an ordinal comparison to allow username case changes to flow through.
+                    var removedUsernames = oldOwners.Except(newOwners, StringComparer.Ordinal).ToList();
+                    var addedUsernames = newOwners.Except(oldOwners, StringComparer.Ordinal).ToList();
+
+                    if (removedUsernames.Any() || addedUsernames.Any())
+                    {
+                        // CHANGED: The username set has changed.
+                        result.Add(id, newOwners.ToArray());
```
skofman1: >newOwners.ToArray [](start = 39, length = 17)

I remember conversations about owners sorting in search results. This is probably not part of this code but I wonder if you considered this.

joelverhagen: Yup, I considered this. I chose `OrdinalIgnoreCase` since it is the simplest and fastest. If we want another ordering (like the one specified in https://github.com/NuGet/NuGetGallery/issues/6258) we would do this in either search service or the gallery, not here.

src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerSetComparer.cs
joelverhagen: nope, they are the original casing found in the DB.

src/NuGet.Services.AzureSearch/Owners2AzureSearch/IOwnerSetComparer.cs
joelverhagen: Sure

src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerSetComparer.cs
skofman1: so might cause a problem, no?

joelverhagen: `TryGetValue` observes the container's comparer, which is `StringComparer.OrdinalIgnoreCase`. If that doesn't address what you mean, I don't understand.

src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerSetComparer.cs
```diff
@@ -0,0 +1,91 @@
...
+        public OwnerSetComparer(ILogger logger)
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public SortedDictionary<string, string[]> Compare(
+            SortedDictionary<string, SortedSet<string>> oldData,
+            SortedDictionary<string, SortedSet<string>> newData)
+        {
+            // We use a very simplistic algorithm here. Perform one pass on the new data to find the added or changed
```
scottbommarito: put this on `///<remarks>` instead?

src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerSetComparer.cs
scottbommarito: I think @skofman1 is concerned that one set may have normalized IDs while the other doesn't, so comparisons will fail. Given that the comparison is case-insensitive, there isn't any issue here, right?

src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerSetComparer.cs
```diff
@@ -0,0 +1,91 @@
...
+                }
+            }
+
+            // Second pass: find removed sets.
+            foreach (var pair in oldData)
+            {
+                var id = pair.Key;
+                var oldOwners = pair.Value;
+
+                if (!newData.TryGetValue(id, out var newOwners))
```
scottbommarito: Assuming that the two sets are mostly the same, we're doing a comparison here that we've _usually_ already done above. Is there a way we could filter out the IDs we've already processed from this second pass, or rewrite the algorithm so that we only process each ID once? This could become very expensive if we're comparing millions of packages owners at once.

For example, provided that `Distinct` isn't very expensive, potentially something like this could be much faster.

```
var ids = oldData.Keys.Concat(newData.Keys).Distinct();
foreach (var id in ids)
{
    var wasInOld = oldData.TryGetValue(id, out var oldOwners);
    var isInNew = newData.TryGetValue(id, out var newOwners);
    if (wasInOld)
    {
        if (isInNew)
        {
            // check for changes
        }
        else
        {
            // removed
        }
    }
    else
    {
        if (isInNew)
        {
            // added
        }
        else
        {
            // throw exception because this shouldn't happen
        }
    }
}
```

src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerSetComparer.cs
```diff
@@ -0,0 +1,91 @@
...
+        private static readonly string[] EmptyStringArray = new string[0];
+
+        private readonly ILogger _logger;
+
+        public OwnerSetComparer(ILogger logger)
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public SortedDictionary<string, string[]> Compare(
```
loic-sharma: >Compare [](start = 50, length = 7)

Should we assert that the `oldData` and `newData` have the expected `.Comparer`?

src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerSetComparer.cs
joelverhagen: IMO this is more about the implementation than the interface. I think it should be inline because of this.

src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerSetComparer.cs
joelverhagen: Certainly doesn't hurt.

src/NuGet.Services.AzureSearch/Owners2AzureSearch/OwnerSetComparer.cs
joelverhagen: I am not exactly sure how `Distinct` is implemented by my guess is that it uses a hash set internally. This means that in the case where `oldData` and `newData` are the same (the extremem closest to what is going to happen the most), we'll be performing 4 * N hash look-ups. 2 * N in `Distinct` and 2 * N in the `foreach`. This should become obvious with some basic performance testing. I'll try to find some time today to do this.

joelverhagen: DEV data, 100 iterations, no debugger attached, my devbox. Averaged values.

Existing algorithm in PR: 381ms.
Suggested algorithm: 427ms.

In short, `Distinct` is not magic. I looked at the implementation. It uses a hash set.

PR #501, opened 2019-04-10
[Azure Search 1] Add DatabaseOwnerFetcher to query for current owner information

joelverhagen:

Query inspired by:
https://github.com/NuGet/NuGet.Jobs/blob/master/src/Search.GenerateAuxiliaryData/SqlScripts/Owners.sql

Progress on https://github.com/NuGet/NuGetGallery/issues/6475

### DIFF
--- a/src/NuGet.Services.AzureSearch/Owners2AzureSearch/IOwnerSetComparer.cs
+++ b/src/NuGet.Services.AzureSearch/Owners2AzureSearch/IOwnerSetComparer.cs
@@ -5,8 +5,20 @@ using System.Collections.Generic;
 
 namespace NuGet.Services.AzureSearch.Owners2AzureSearch
 {
+    /// <summary>
+    /// Used to compare two sets of owners to determine the changes.
+    /// </summary>
     public interface IOwnerSetComparer
     {
+        /// <summary>
+        /// Compares two sets of owners to determine the package IDs that have changed. The returned dictionary
+        /// contains owners that were added, removed, and changed. For the "added" and "changed" cases, the owner set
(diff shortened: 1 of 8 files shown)


joelverhagen:
There are two problems with unit tests here:

1. There's not much useful to assert with regards to the SQL string itself.
1. The factory returns `SqlConnection` which is not mockable.

Because of these hurdles and since the code is small, I chose not to unit test it.

src/NuGet.Services.AzureSearch/Owners2AzureSearch/IOwnerSetComparer.cs
```diff
@@ -5,8 +5,18 @@
...
+    /// </summary>
     public interface IOwnerSetComparer
     {
+        /// <summary>
+        /// Compares two sets of owners to determine the package IDs that have changed. The returned dictionary
+        /// contains owners that were added, removed, and changed. For the "added" and "changed" cases, the owner set
+        /// is the new data. For the "removed" case, the set is empty.
+        /// </summary>
+        /// <param name="oldData">The old owner information, typically from storage.</param>
+        /// <param name="newData">The new owner information, typically from gallery DB.</param>
```
shishirx34: may be specify the return type of <id, list<owners>>?

loic-sharma:
@joelverhagen The `azuresearch` branch was merged into the `dev` branch. I updated this to target the `dev` branch

src/NuGet.Services.AzureSearch/Owners2AzureSearch/IOwnerSetComparer.cs
joelverhagen: Done.

PR #502, opened 2019-04-10
[Azure Search 2] Add UpdateOwners model to update just the owners field

joelverhagen:
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/501.

Adds an Azure Search model so we can update just the owners field on a search document.

### DIFF
--- a/src/NuGet.Services.AzureSearch/Models/SearchDocument.cs
+++ b/src/NuGet.Services.AzureSearch/Models/SearchDocument.cs
@@ -27,7 +27,7 @@ namespace NuGet.Services.AzureSearch
         /// Used when processing <see cref="SearchIndexChangeType.AddFirst"/>.
         /// </summary>
         [SerializePropertyNamesAsCamelCase]
-        public class AddFirst : UpdateLatest
+        public class AddFirst : UpdateLatest, IOwners
         {
             [IsSearchable]
             [Analyzer(ExactMatchCustomAnalyzer.Name)]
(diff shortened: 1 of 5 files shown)


tests/NuGet.Services.AzureSearch.Tests/SearchDocumentBuilderFacts.cs
```diff
@@ -138,6 +138,42 @@ public async Task SetsExpectedProperties()
...
+        public class UpdateOwners : BaseFacts
+        {
+            public UpdateOwners(ITestOutputHelper output) : base(output)
+            {
+            }
+
+            [Fact]
+            public async Task SetsExpectedProperties()
+            {
+                var document = _target.UpdateOwners(
```
shishirx34: may be also assert the expected owners list before invoking the `UpdateOwners` that way we know its changed?

src/NuGet.Services.AzureSearch/Models/SearchDocument.cs
```diff
@@ -60,6 +60,17 @@ public class UpdateVersionList : CommittedDocument, IVersions
             public bool? IsLatest { get; set; }
         }
 
+        /// <summary>
+        /// Used when updating just the owners of a document. Note that this model does not need any analyzer or
+        /// other Azure Search attributes since it is not used for index creation. The <see cref="Full"/> and it's
```
loic-sharma:
```suggestion
        /// other Azure Search attributes since it is not used for index creation. The <see cref="Full"/> and its
```

joelverhagen: Woah, what is this!

joelverhagen: Okay, that's cool.

tests/NuGet.Services.AzureSearch.Tests/SearchDocumentBuilderFacts.cs
joelverhagen: The rest of the tests assume they are constant. Will leave as-is.

PR #508, opened 2019-04-25
Initial implementation of self-contianed icon ingestion for catalog2dnx

agr:
Addresses https://github.com/NuGet/NuGetGallery/issues/7091, https://github.com/NuGet/NuGetGallery/issues/7072

Needed an indication of the embedded icon presence for catalog2dnx job. For that purpose an instance of `ICatalogClient` is now used by `DnxCatalogCollector`. Since `CatalogClient` implementation requires an `HttpClient` in one of its dependencies constructor which is not available at `DnxCatalogCollector` construction time, `DnxCatalogCollector` was modified to accept a factory of `ICatalogClient`s, to be able to instantiate catalog client as needed.

Icon is extracted from the nupkg, two implementations are provided:
1. The main implementation that uses `IAzureStorage` and the ability to seek blob streams;
2. Fallback implementation that downloads nupkg to a temp file.

Content type determination is done by file extension and for now will rely on Gallery doing its job when validating the icon. [Later](https://github.com/nuget/nugetgallery/issues/7061), we can replace it with proper format determination.


src/Catalog/Dnx/DnxCatalogCollector.cs
```diff
@@ -248,12 +253,17 @@ private async Task<bool> AreRequiredPropertiesPresentAsync(Storage destinationSt
             string packageId,
             string normalizedPackageVersion,
             Uri sourceUri,
+            Uri catalogLeafUri,
             Dictionary<string, string> telemetryProperties,
             CancellationToken cancellationToken)
         {
+            var catalogClient = _catalogClientFactory(client);
+            var catalogPage = await catalogClient.GetPackageDetailsLeafAsync(catalogLeafUri.AbsoluteUri);
```
loic-sharma: >catalogPage [](start = 16, length = 11)

This should be `catalogLeaf`

src/Catalog/Dnx/DnxMaker.cs
```diff
@@ -260,6 +274,122 @@ private async Task<Uri> SaveNupkgAsync(Stream nupkgStream, Storage storage, stri
...
+                iconFilename,
+                destinationStorage,
+                packageId,
+                normalizedPackageVersion,
+                cancellationToken,
+                () => Task.FromResult(nupkgStream),
+                disposeOfPackageStream: false);
+        }
+
+        private async Task CopyIconAsync(
```
loic-sharma: >CopyIconAsync [](start = 27, length = 13)

I feel as if this method would be cleaner if the package stream was managed by its caller. The caller can create the stream and choose whether or not to dispose it.

src/NuGet.Indexing/IndexingConfiguration.cs
```diff
@@ -9,7 +9,7 @@ namespace NuGet.Indexing
     /// <summary>
     /// Provides configuration for Indexing.
     /// </summary>
-    public class IndexingConfiguration : Configuration
+    public class IndexingConfiguration : NuGet.Services.Configuration.Configuration
```
loic-sharma: >uGet.Services.Configuration. [](start = 42, length = 28)

Is this change intended?

src/Catalog/Dnx/DnxMaker.cs
```diff
@@ -260,6 +274,122 @@ private async Task<Uri> SaveNupkgAsync(Stream nupkgStream, Storage storage, stri
...
+        {
+            if (iconPath.EndsWith(".png", StringComparison.InvariantCultureIgnoreCase))
+            {
+                return "image/png";
+            }
+            if (iconPath.EndsWith(".jpg", StringComparison.InvariantCultureIgnoreCase) || iconPath.EndsWith(".jpeg", StringComparison.InvariantCultureIgnoreCase))
+            {
+                return "image/jpeg";
+            }
+            return "image/jpeg";
```
loic-sharma: >image [](start = 20, length = 5)

Should we maybe throw here?

agr: I don't want package publishing to fail because of some icon issues. I'd rather have it published with wrong type and have client do its magic.

src/NuGet.Indexing/IndexingConfiguration.cs
agr: Without the change it fails to build with 
> error CS0118: 'Configuration' is a namespace but is used like a type
I'm not sure how it works in our main branches.

src/Catalog/Dnx/DnxMaker.cs
loic-sharma: > I don't want package publishing to fail because of some icon issues. 

I agree, but we should be able to catch this specific problem in the Gallery when the package is uploaded, right?

> I'd rather have it published with wrong type and have client do its magic.

Will the client be able to properly handle an image with the wrong mime type?

scottbommarito: Is there a mimetype that means "image but not sure what kind"? JPEG is misleading.

We should definitely have telemetry when this event happens.
Perhaps we should even have a metric that logs the file types used by images?
At a bare minimum, I think we should log a warning when it's not as JPEG or PNG.

Also, add a comment to this code so we know what's going on.

(nit: add a line between this and the `}`)

src/NuGet.Indexing/Extraction/CatalogPackageReader.cs
```diff
@@ -87,5 +87,20 @@ public override Task<byte[]> GetArchiveHashAsync(HashAlgorithmName hashAlgorithm
         {
             throw new NotImplementedException();
         }
+
+        public override bool CanVerifySignedPackages(SignedPackageVerifierSettings verifierSettings)
+        {
+            throw new NotImplementedException();
```
scottbommarito: Shouldn't this be `false` instead?

src/Catalog/Dnx/DnxMaker.cs
agr: Gallery is supposed to validate everything related to the embedded icon. Client already handles invalid mime types since there are 0 guarantees regarding where iconUrl points to.

src/Catalog/Dnx/DnxMaker.cs
```diff
@@ -4,37 +4,50 @@
...
             string normalizedPackageVersion,
+            string iconFilename,
             CancellationToken cancellationToken)
         {
             if (nupkgStream == null)
             {
                 throw new ArgumentNullException(nameof(nupkgStream));
             }
 
+            if (!nupkgStream.CanSeek)
```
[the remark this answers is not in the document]
agr: In that particular method `nupkgStream` is a file stream pointing to the temp file on local disk, so it is always seekable.

src/Catalog/Dnx/DnxMaker.cs
```diff
@@ -260,6 +296,116 @@ private async Task<Uri> SaveNupkgAsync(Stream nupkgStream, Storage storage, stri
...
+            string normalizedPackageVersion,
+            CancellationToken cancellationToken)
+        {
+            var packageFileName = PackageUtility.GetPackageFileName(packageId, normalizedPackageVersion);
+            var sourceUri = sourceStorage.ResolveUri(packageFileName);
+            var packageSourceBlob = await sourceStorage.GetCloudBlockBlobReferenceAsync(sourceUri);
+            return await packageSourceBlob.GetStreamAsync(cancellationToken);
+        }
+
+        private string GetIconContentType(string iconPath)
```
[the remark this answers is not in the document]
agr: Only png and jpg files are allowed to be packaged into nupkg files.

src/NuGet.Indexing/Extraction/CatalogPackageReader.cs
agr: It is never called neither in our code, nor in the client library, so shouldn't matter. I will change though.

src/Catalog/Dnx/DnxMaker.cs
agr: I did not find a mime type that means "some sort of image". But I ran some tests: browsers seem to ignore the specific type of an image (i.e. if server returns jpeg and says it's `image/png` browsers display it correctly). VS client does something similar.

loic-sharma: Interesting. Did the image have an extension in your tests?

src/NuGet.Protocol.Catalog/Models/PackageDetailsCatalogLeaf.cs
```diff
@@ -78,5 +78,8 @@ public class PackageDetailsCatalogLeaf : CatalogLeaf
 
         [JsonProperty("verbatimVersion")]
         public string VerbatimVersion { get; set; }
+
+        [JsonProperty("icon")]
+        public string IconFilename { get; set; }
```
joelverhagen: Why isn't this `IconFileName`?

agr: Because that's a single [word](https://en.wikipedia.org/wiki/Filename)? :)

joelverhagen: Oh, TIL. Is this consistent across our docs/client code?

src/Catalog/Dnx/DnxMaker.cs
```diff
@@ -114,6 +149,7 @@ public async Task DeletePackageAsync(string id, string version, CancellationToke
 
             await DeleteNuspecAsync(storage, id, normalizedVersion, cancellationToken);
             await DeleteNupkgAsync(storage, id, normalizedVersion, cancellationToken);
+            await DeleteIconAsync(storage, id, normalizedVersion, cancellationToken);
```
joelverhagen: Is license deleted?

agr: Catalog2dnx does not extract the license, so it does not clean it up. It is deleted by Gallery though.

src/Catalog/Dnx/DnxMaker.cs
```diff
@@ -91,6 +116,16 @@ public DnxMaker(StorageFactory storageFactory)
             var destinationStorage = _storageFactory.Create(packageId);
             var nuspecUri = await SaveNuspecAsync(destinationStorage, packageId, normalizedPackageVersion, nuspec, cancellationToken);
             var nupkgUri = await CopyNupkgAsync(sourceStorage, destinationStorage, packageId, normalizedPackageVersion, cancellationToken);
+            if (!string.IsNullOrWhiteSpace(iconFilename))
```
joelverhagen: This should be done before .nupkg copy. In `DnxCatalogCollector` the .nupkg is used as a marker that this code has finished running.

src/Catalog/Dnx/DnxMaker.cs
joelverhagen: We need a comment about the arbitrary mime-type selection. What if you just leave the mime-type blank in this case?

src/Catalog/Dnx/DnxMaker.cs
```diff
@@ -302,6 +458,13 @@ public static string GetRelativeAddressNupkg(string id, string version)
             return $"{normalizedVersion}/{id}.{normalizedVersion}.nupkg";
         }
 
+        private static string GetRelativeAddressIcon(string id, string version)
+        {
+            var normalizedVersion = NuGetVersion.Parse(version).ToNormalizedString();
+
+            return $"{normalizedVersion}/icon";
```
joelverhagen: Does client know where they are storing icons locally yet? Originally, this folder structure was supposed to mimic the V3 folder structure on disk.

src/Catalog/Dnx/DnxMaker.cs
joelverhagen: Okay, out of scope then. Could you link the issue here tracking this improper design?

src/Catalog/Dnx/DnxMaker.cs
joelverhagen: Spoke offline. Maybe we should drop the mime-type determination till we know what gallery is doing here in terms of validation.

src/Catalog/xslt/nuspec.xslt
```diff
@@ -125,7 +125,22 @@
                 </xsl:when>
               </xsl:choose>
             </xsl:when>
-            
+
+            <xsl:when test="self::nuget:icon">
```
joelverhagen: 👍

src/Catalog/Dnx/DnxMaker.cs
agr: https://github.com/NuGet/NuGetGallery/issues/7110

src/NuGet.Protocol.Catalog/Models/PackageDetailsCatalogLeaf.cs
agr: Renamed to `iconFile` to be consistent with catalog entry.

src/Catalog/Dnx/DnxMaker.cs
agr: Removed for now.

src/Catalog/Dnx/DnxMaker.cs
agr: We already deviated from that with licenses. Those are stored as `/license` under appropriate packageid/version path. I am being consistent with licenses here.

src/Catalog/Dnx/DnxMaker.cs
```diff
@@ -260,6 +298,114 @@ private async Task<Uri> SaveNupkgAsync(Stream nupkgStream, Storage storage, stri
...
+        {
+            using (var zipArchive = new ZipArchive(packageStream, ZipArchiveMode.Read, leaveOpen: true))
+            {
+                var iconEntry = zipArchive.Entries.FirstOrDefault(e => e.FullName.Equals(iconPath, StringComparison.InvariantCultureIgnoreCase));
+                if (iconEntry != null)
+                {
+                    using (var iconStream = iconEntry.Open())
+                    {
+                        _logger.LogInformation("Extracting icon to the destination storage {DestinationUri}", destinationUri);
+                        // TODO: align the mime type determination with Gallery
```
joelverhagen: TODO should have a github issue link

src/Catalog/Dnx/DnxMaker.cs
joelverhagen: Sounds good.

src/Catalog/NuGet.Services.Metadata.Catalog.csproj
```diff
@@ -269,7 +272,7 @@
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
     <PackageReference Include="NuGet.Packaging.Core">
-      <Version>4.8.0</Version>
+      <Version>5.0.0-preview1.5665</Version>
```
dominoFire: What feature do you need from this unreleased/preview version of `NuGet.Packaging.Core`? Is it good to include preview package versions?

agr: That specific version is to align with [Gallery](https://github.com/NuGet/NuGetGallery/blob/0ca0368afbfc1bbf7133e2d72389fc168d35f6ce/src/NuGetGallery/NuGetGallery.csproj#L2362-L2367). It is needed to have access to the `PathUtility.StripLeadingDirectorySeparators` method.

In Gallery that particular version is used because it is one of the latest preview versions that has license metadata parsing support and not yet targeted to .NET Framework 4.7.2.

PR #510, opened 2019-04-26
[Azure Search] Split tokens on whitespace without keeping the original token

loic-sharma:
I tested this by indexing all packages on DEV.

Fixes https://github.com/nuget/nugetgallery/issues/7045


src/NuGet.Services.AzureSearch/Analysis/PackageIdCustomTokenizer.cs
```diff
@@ -3,15 +3,15 @@
...
+    /// For example, "Foo.Bar Baz" becomes "Foo", "Bar", and "Baz".
     /// </summary>
     public static class PackageIdCustomTokenizer
     {
         public const string Name = "nuget_package_id_tokenizer";
 
         public static readonly PatternTokenizer Instance = new PatternTokenizer(
             Name,
-            @"[.\-_,;:'*#!~+()\[\]{}]");
+            @"[.\-_,;:'*#!~+()\[\]{}\s]");
```
shishirx34: BTW, Its fine to have here, but I think its only applicable to the DescriptionAnalyzer, since the ID wouldn't actually have spaces in it. In case you think we want it to be applied to specific analyzers.

loic-sharma: Yup, that's correct. I kept this as one so that we wouldn't need to mirror future changes to `PackageIdCustomTokenizer` to this second tokenizer. That being said, we can reassess if requirements change.

PR #512, opened 2019-04-30
[Hotfix] Make ng.exe 64-bit

joelverhagen:
Previously it was limited to 4 GB memory and was slowing down (many, many GCs) and eventually hitting an out-of-memory exception.

Mitigation for https://github.com/NuGet/NuGetGallery/issues/7101.

### DIFF
--- a/NuGet.Services.Metadata.sln
+++ b/NuGet.Services.Metadata.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.26719.0
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.28827.37
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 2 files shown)


scottbommarito:
A couple questions:
- I thought AnyCPU automatically ran as 64-bit on a 64-bit machine?
- Why don't we make all of our jobs and services 64-bit?

joelverhagen:
> Why don't we make all of our jobs and services 64-bit?

There is insignificant benefit in most cases. And we would need to validate that the dependencies work.

> I thought AnyCPU automatically ran as 64-bit on a 64-bit machine?

There are two kinds of AnyCPU. `anycpu32bitpreferred` is the default.

See for more details.
http://blogs.microsoft.co.il/sasha/2012/04/04/what-anycpu-really-means-as-of-net-45-and-visual-studio-11/

scottbommarito:
@joelverhagen - Any reason we can't make it `Any CPU` (not 32-bit preferred) instead? This is probably a minor concern, but if someone is trying to run the V3 jobs from a 32-bit machine they'll be blocked by this change.

PR #514, opened 2019-05-02
Remove unnecessary Exists check in AzureStorage

joelverhagen:
This PR does 3 things. A bit of polish find while investigating https://github.com/NuGet/NuGetGallery/issues/7101.

1. Don't call `Exists` before loading a blob from storage. Just handle the 404. This saves an HTTP request.
2. Remove superfluous logs. The ones I removed are not needed because the storage layer already logs each URL before and after the request. This made debugging easier.
3. Last week, we saw a mysterious JSON parse error of a page. It wasn't clear what URL that was so I beefed up the logging.


joelverhagen:
Note that I tried adding a throttle to the page downloads as originally surmised, but this had minimal impact on the catalog leafs that caused the out of memory exception. Yes requests got slower with massive parallelism, but they all ended up succeeding in a similar time as with a throttle. If we run into this later, we can fix it at that time. My guess is that until the degree of parallelism causes an HTTP timeout, there won't be any benefit from a throttle.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -339,13 +340,15 @@ protected override async Task<StorageContent> OnLoadAsync(Uri resourceUri, Cance
 
                 return new StringStorageContent(content);
             }
-
-            if (Verbose)
+            catch (StorageException ex) when (ex.RequestInformation.HttpStatusCode == (int)HttpStatusCode.NotFound)
```
loic-sharma: I believe we use `.RequestInformation?.HttpStatusCode` in other areas, do we need to follow that pattern here?

src/Catalog/Registration/RegistrationPersistence.cs
```diff
@@ -155,9 +151,8 @@ private static async Task<IGraph> LoadCatalog(IStorage storage, Uri resourceUri,
 
         private static async Task<IGraph> LoadCatalogPage(IStorage storage, Uri pageUri, CancellationToken cancellationToken)
         {
-            Trace.TraceInformation("RegistrationPersistence.LoadCatalogPage: pageUri = {0}", pageUri);
             string json = await storage.LoadStringAsync(pageUri, cancellationToken);
-            IGraph graph = Utils.CreateGraph(json);
+            IGraph graph = Utils.CreateGraph(pageUri, json);
```
loic-sharma: Nit: return directly?

src/Catalog/Persistence/AzureStorage.cs
joelverhagen: Hmm... that is a good point. I did not consider this being null. I think I investigated this in the past by digging through storage SDK code but I forget the conclusion. Best to be defensive in the `catch` case.

src/Catalog/Registration/RegistrationPersistence.cs
joelverhagen: I left this since I put a break point here a couple times while debugging to inspect the graph. This is the big guy that is eating memory.

PR #515, opened 2019-05-07
Complete Owners2AzureSearch job and wire-up the entry point

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/6475

### DIFF
--- a/src/NuGet.Services.AzureSearch/Owners2AzureSearch/Owners2AzureSearchCommand.cs
+++ b/src/NuGet.Services.AzureSearch/Owners2AzureSearch/Owners2AzureSearchCommand.cs
@@ -1,0 +1,103 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using NuGet.Services.Metadata.Catalog.Helpers;
+
+namespace NuGet.Services.AzureSearch.Owners2AzureSearch
+{
+    public class Owners2AzureSearchCommand
+    {
+        private readonly IDatabaseOwnerFetcher _databaseOwnerFetcher;
+        private readonly IOwnerDataClient _ownerDataClient;
+        private readonly IOwnerSetComparer _ownerSetComparer;
+        private readonly IOwnerIndexActionBuilder _indexActionBuilder;
+        private readonly Func<IBatchPusher> _batchPusherFactory;
+        private readonly IOptionsSnapshot<AzureSearchJobConfiguration> _options;
+        private readonly ILogger<Owners2AzureSearchCommand> _logger;
+
+        public Owners2AzureSearchCommand(
+            IDatabaseOwnerFetcher databaseOwnerFetcher,
+            IOwnerDataClient ownerDataClient,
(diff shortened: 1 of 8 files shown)


src/NuGet.Services.AzureSearch/Owners2AzureSearch/Owners2AzureSearchCommand.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            if (_options.Value.MaxConcurrentBatches <= 0)
+            {
+                throw new ArgumentException(
+                    $"The {nameof(AzureSearchJobConfiguration.MaxConcurrentBatches)} must be greater than zero.",
+                    nameof(options));
+            }
+        }
+
+        public async Task ExecuteAsync()
```
loic-sharma: >ExecuteAsync [](start = 26, length = 12)

Should we track the duration of `ExecuteAsync` with a metric? This would help us drive decisions like how frequently we should update owners.

joelverhagen: The jobs infra traces this duration per loop.

Tracked here https://github.com/NuGet/NuGetGallery/issues/6664.

src/NuGet.Services.AzureSearch/Owners2AzureSearch/Owners2AzureSearchCommand.cs
```diff
@@ -0,0 +1,106 @@
...
+            // Persist in storage the list of all package IDs that have owner changes. This allows debugging and future
+            // analytics on frequency of ownership changes.
+            _logger.LogInformation("Uploading the package IDs that have owner changes to blob storage.");
+            await _ownerDataClient.UploadChangeHistoryAsync(changes.Keys.ToList());
+
+            _logger.LogInformation("Uploading the new owner data to blob storage.");
+            await _ownerDataClient.ReplaceLatestIndexedAsync(databaseResult, storageResult.AccessCondition);
+        }
+
+        private async Task WorkAsync(ConcurrentBag<IdAndValue<string[]>> changesBag)
```
scottbommarito: I think this would be a useful thing to share--a method that grabs items from a `ConcurrentBag` and then processes items until none are left. I know I wrote something that did this when I worked on the monitoring jobs. Do we have anything like this today? Could we move this to a shared place if not?

src/NuGet.Services.AzureSearch/Owners2AzureSearch/Owners2AzureSearchCommand.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            if (!changes.Any())
+            {
+                return;
+            }
+
+            _logger.LogInformation(
+                "Starting {Count} workers pushing owners changes to Azure Search.",
+                _options.Value.MaxConcurrentBatches);
+            var workerTasks = Enumerable
```
scottbommarito: Maybe replace this with https://github.com/NuGet/NuGet.Services.Metadata/blob/master/src/Catalog/Helpers/ParallelAsync.cs?

joelverhagen: Yes, I can do that.

src/NuGet.Services.AzureSearch/Owners2AzureSearch/Owners2AzureSearchCommand.cs
joelverhagen: Spoke offline. This not easy to make generic and readable.

PR #516, opened 2019-05-09
Catalog2registration changes for icons

agr:
Addresses https://github.com/NuGet/NuGetGallery/issues/7062.

Catalog2registration will now emit iconUrl value depending on the presence of `iconFile` property in catalog.

Dropped a few lines from SPARQL query as they did nothing.


src/Catalog/Registration/RegistrationMakerCatalogItem.cs
```diff
@@ -214,5 +215,25 @@ public override IGraph CreatePageContent(CatalogContext context)
                 throw new Exception(string.Format("Exception processing catalog item {0}", _catalogUri), e);
             }
         }
+
+        private string GetIconUrl()
+        {
+            var subject = _catalogItem.CreateUriNode(_catalogUri);
+            var packageId = _catalogItem.GetTriplesWithSubjectPredicate(subject, _catalogItem.CreateUriNode(Schema.Predicates.Id)).FirstOrDefault().Object.ToString();
```
scottbommarito: I see we fetch the package ID and version like this in a couple places in this file...could we share this code?

src/Catalog/sparql/ConstructRegistrationPageContentGraph.rq
```diff
@@ -47,6 +47,8 @@ WHERE
 
 	BIND(@licenseUrl AS ?licenseUrl)
 
+	BIND(@iconUrl AS ?iconUrl)
```
zhhyu: So it will generate an empty url string for icon like licenseUrl if no catalog icon info?

src/Catalog/Registration/IPackagePathProvider.cs
```diff
@@ -4,10 +4,11 @@
 namespace NuGet.Services.Metadata.Catalog.Registration
 {
     /// <summary>
-    /// An interface for providing the relative path of a package in the base address
+    /// An interface for providing the relative path of a package or an icon in the base address
     /// </summary>
     public interface IPackagePathProvider
     {
+        string GetIconPath(string id, string version);
```
zhhyu: Is that better that we have another interface and implementation to get icon path? It will enhance single-responsibility principle.

src/Catalog/Registration/FlatContainerPackagePathProvider.cs
```diff
@@ -22,5 +22,13 @@ public string GetPackagePath(string id, string version)
 
             return $"{_container}/{idLowerCase}/{versionLowerCase}/{packageFileName}";
         }
+
+        public string GetIconPath(string id, string version)
```
cristinamanum: Can the arguments be null/empty ?

src/Catalog/sparql/ConstructRegistrationPageContentGraph.rq
```diff
@@ -57,8 +59,6 @@ WHERE
 	OPTIONAL { ?catalogEntry nuget:description ?optionalDescription . }
 	OPTIONAL { ?catalogEntry nuget:title ?optionalTitle . }
 	OPTIONAL { ?catalogEntry nuget:summary ?optionalSummary . }
-	OPTIONAL { ?catalogEntry nuget:iconUrl ?optionalIconUrl . }
-	OPTIONAL { ?catalogEntry nuget:licenseUrl ?licenseUrl . }
```
cristinamanum: Is deleting the license intended?

src/Catalog/Registration/RegistrationMakerCatalogItem.cs
```diff
@@ -214,5 +215,25 @@ public override IGraph CreatePageContent(CatalogContext context)
...
+            if (!string.IsNullOrWhiteSpace(iconFileTriple?.Object.ToString()) && !string.IsNullOrWhiteSpace(packageId) && !string.IsNullOrWhiteSpace(packageVersion))
+            {
+                // The embedded icon file case. We assume here that catalog2dnx did its job
+                // and extracted the icon file to the appropriate location.
+                string path = PackagePathProvider.GetIconPath(packageId, packageVersion);
+                return new Uri(_packageContentBaseAddress, path).AbsoluteUri;
+            }
+
+            // TODO: Update when iconUrl ingestion work is done: https://github.com/nuget/nugetgallery/issues/7061
+            return iconUrlTriple?.Object.ToString() ?? "";
```
cristinamanum: It is ok to have the empty entries in catalog?

src/Catalog/sparql/ConstructRegistrationPageContentGraph.rq
zhhyu: Previously, we will set an empty licenseUrl string in the registration. Client will decide whether to show the licenseUrl base on it's empty or not. If we don't inject the licenseUrl if the package has no license info, client should have the logic to handle this case rather than expect to check whether it's empty or not.

src/Catalog/sparql/ConstructRegistrationPageContentGraph.rq
agr: This just binds `?iconUrl` to whatever code produces and it produces an empty string in the case you describe, so yes.

src/Catalog/sparql/ConstructRegistrationPageContentGraph.rq
agr: Yes, it was intended. Removed lines did nothing, the value is always provided from the code.

src/Catalog/Registration/RegistrationMakerCatalogItem.cs
agr: This produces an empty string value in registration. And yes, that's existing behavior. See [here](https://api.nuget.org/v3/registration3-gz-semver2/e2e.signedpackage/page/1.0.0-180502.211641.6633165/2018.7.19-v193608.json) for example.

src/Catalog/Registration/FlatContainerPackagePathProvider.cs
agr: added checks.

src/Catalog/Registration/IPackagePathProvider.cs
agr: That would probably be nice if we had proper DI in V3 jobs. Right now it will cause too much noise. Currently they are logically grouped by container type (packages vs flat container), I can drop the "Packages" work from the interface name, but icon path is a package property, so current implementation seems logical for me.

zhhyu: Ok

src/Catalog/Registration/RegistrationMakerCatalogItem.cs
agr: Updated

PR #517, opened 2019-05-14
Initialize the owners.v2.json file in Db2AzureSearch

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/7114

Chose to make a small design change. Now all package registrations are processed, not just ones that have packages. If there are no versions, however, no index changes will be made.

This change is so the resulting owners.v2.json has owners even for package registrations currently with no versions. The source of this case is package deletes.

Also, come code was moved out og the `Owners2AzureSearch` namespace since it is now used by more than one job.

### DIFF
--- a/src/NuGet.Services.AzureSearch/Db2AzureSearch/Db2AzureSearchCommand.cs
+++ b/src/NuGet.Services.AzureSearch/Db2AzureSearch/Db2AzureSearchCommand.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
@@ -12,6 +13,7 @@ using NuGet.Protocol.Catalog;
 using NuGet.Services.AzureSearch.Catalog2AzureSearch;
 using NuGet.Services.Metadata.Catalog;
 using NuGet.Services.Metadata.Catalog.Persistence;
+using NuGetGallery;
 
 namespace NuGet.Services.AzureSearch.Db2AzureSearch
 {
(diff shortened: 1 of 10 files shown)


src/NuGet.Services.AzureSearch/Db2AzureSearch/Db2AzureSearchCommand.cs
```diff
@@ -134,6 +127,41 @@ private async Task InitializeAsync()
...
+            var ownersBuilder = new PackageIdToOwnersBuilder(_logger);
+            foreach (var owners in allOwners)
+            {
+                ownersBuilder.Add(owners.Id, owners.Value);
+            }
+
+            await _ownerDataClient.ReplaceLatestIndexedAsync(
+                ownersBuilder.GetResult(),
+                AccessConditionWrapper.GenerateIfNotExistsCondition());
+        }
```
loic-sharma: Consider logging when the work completes.

src/NuGet.Services.AzureSearch/Db2AzureSearch/NewPackageRegistrationProducer.cs
```diff
@@ -296,5 +297,21 @@ public PackageRegistrationRange(int minKey, int? maxKey, int packageCount)
             /// </summary>
             public int PackageCount { get; }
         }
+
+        public class PackageRegistrationInfo
```
loic-sharma: This isn't use by anyone else other than `NewPackageRegistrationProducer` right? Should this be `private`?

src/NuGet.Services.AzureSearch/Db2AzureSearch/Db2AzureSearchCommand.cs
joelverhagen: This is handled by the job infrastructure.

src/NuGet.Services.AzureSearch/Db2AzureSearch/NewPackageRegistrationProducer.cs
joelverhagen: Yeah, nice catch.

src/NuGet.Services.AzureSearch/Db2AzureSearch/Db2AzureSearchCommand.cs
loic-sharma: Oh I meant at the granularity of generating the owners.v2.json file. Considering it's a big file, this may be slower than expected.

joelverhagen: Oh okay, I'll add this in my current branch for the next set of changes.

PR #519, opened 2019-05-15
CollectorHttpClient should retry on OperationCanceledException

scottbommarito:
https://github.com/nuget/nugetgallery/issues/7138

`CollectorHttpClient` sometimes throws `OperationCanceledException` when making web requests while there is high network load (e.g. when processing 10,000 catalog leaves simultaneously). If a web request fails due to this reason, it suggests a client-side issue and the web request should be retried.


src/Catalog/RetryWithExponentialBackoff.cs
```diff
@@ -39,23 +39,28 @@ public async Task<HttpResponseMessage> SendAsync(HttpClient client, Uri address,
 
                     return httpResponse;
                 }
-                catch (HttpRequestException) when (IsTransientError(httpResponse))
+                catch (Exception e)
```
joelverhagen: UTs?

scottbommarito: lol good call

PR #520, opened 2019-05-15
Feed2Catalog should throttle its Azure storage requests

scottbommarito:
https://github.com/nuget/nugetgallery/issues/7138

Azure storage requests sometimes fail when there is high network load (e.g. when creating 10,000 catalog leaves simultaneously). This PR adds throttling (using `NuGet.Protocol`'s implementation) that prevents more storage requests than `ServicePointManager.DefaultConnectionLimit` from running simultaneously.


src/Catalog/NuGet.Services.Metadata.Catalog.csproj
```diff
@@ -254,6 +254,9 @@
     <PackageReference Include="NuGet.Packaging">
       <Version>5.0.0-preview1.5665</Version>
     </PackageReference>
+    <PackageReference Include="NuGet.Protocol">
+      <Version>4.8.0</Version>
```
joelverhagen: Align this version with NuGet.Packaging.

joelverhagen:
Please get one more sign off.

src/Ng/CommandHelpers.cs
```diff
@@ -27,6 +27,7 @@
 using CatalogAzureStorageFactory = NuGet.Services.Metadata.Catalog.Persistence.AzureStorageFactory;
 using CatalogAzureStorage = NuGet.Services.Metadata.Catalog.Persistence.AzureStorage;
 using CatalogFileStorageFactory = NuGet.Services.Metadata.Catalog.Persistence.FileStorageFactory;
+using NuGet.Protocol;
```
skofman1: sort usings

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -309,6 +292,7 @@ protected override async Task<StorageContent> OnLoadAsync(Uri resourceUri, Cance
 
             CloudBlockBlob blob = GetBlockBlobReference(name);
 
+            await _throttle.WaitAsync();
```
joelverhagen: This is one of the possible download paths in Feed2Catalog. There is also an `HttpClient`-based one. Do we ever hit that path in practice? If so, do we need the throttle there too?

src/Catalog/NuGet.Services.Metadata.Catalog.csproj
scottbommarito: done

src/Catalog/Persistence/AzureStorage.cs
scottbommarito: I wanted to be very cautious about how I applied this throttle. I could have applied this throttle to all of the V3 jobs but I only applied it to this job because it was the only one that seemed affected by the issue that the throttling mitigates.

Similarly, I never saw Feed2Catalog fail [in the `HttpClient`-based path](https://github.com/NuGet/NuGet.Services.Metadata/blob/master/src/Catalog/PackageCatalogItemCreator.cs#L191), so I don't feel like it's necessary to add it there.

Furthermore, the `HttpClient`-based path doesn't even seem to have any retry policy. I think we would want to add that first before adding throttling--but we haven't needed either in the past, so I'm hesitant to add it with this change.

joelverhagen: Sounds good.

PR #521, opened 2019-05-16
[Azure Search] Boost freshness

loic-sharma:
Add a scoring profile that gives reasonable relevancy by favoring packages with high download counts that have been updated recently. I tested this by indexing PROD and then trying queries from our OneNote by hand.

Addresses https://github.com/nuget/nugetgallery/issues/6904


src/NuGet.Services.AzureSearch/ScoringProfiles/NewlyPublishedAndDownloadCountBoosterProfile.cs
```diff
@@ -1,33 +1,45 @@
...
             Name,
             textWeights: new TextWeights
             {
                 Weights = new Dictionary<string, double>
                 {
-                    { IndexFields.PackageId, 10 },              // Exact match of the package id should be boosted higher.
-                    { IndexFields.TokenizedPackageId, 5 }
+                    { IndexFields.PackageId, 10 },              // Exact match of the package id should be boosted the highest.
+                    { IndexFields.TokenizedPackageId, 5 },
+                    { IndexFields.Tags, 1 },
```
loic-sharma: I believe @shishirx34 set this to be explicit, as it seems that a weight of `1` has no effect.  I will follow up with Azure Search team on this.

src/NuGet.Services.AzureSearch/ScoringProfiles/NewlyPublishedAndDownloadCountBoosterProfile.cs
```diff
@@ -1,33 +1,42 @@
...
-                    fieldName: IndexFields.Search.TotalDownloadCount,
+                    fieldName: IndexFields.Search.LogBase2DownloadCount,
                     boost: 100.0,
-                    parameters: new MagnitudeScoringParameters(2000, double.MaxValue, shouldBoostBeyondRangeByConstant: true),
-                    interpolation: ScoringFunctionInterpolation.Logarithmic)
+                    parameters: new MagnitudeScoringParameters(
+                        boostingRangeStart: 0,
+                        boostingRangeEnd: Math.Log(999_999_999_999, 2),
+                        shouldBoostBeyondRangeByConstant: true),
+                    interpolation: ScoringFunctionInterpolation.Linear)
```
loic-sharma: The `LogBase2DownloadCount` field has a smaller range than `TotalDownloadCount`, which makes comparing its boosted score to the `Published` freshness score significantly easier.

src/NuGet.Services.AzureSearch/ScoringProfiles/NewlyPublishedAndDownloadCountBoosterProfile.cs
```diff
@@ -1,33 +1,42 @@
...
 
+using System;
 using System.Collections.Generic;
 using Microsoft.Azure.Search.Models;
 using NuGet.Services.AzureSearch.SearchService;
 
 namespace NuGet.Services.AzureSearch.ScoringProfiles
 {
-    public class DownloadCountBoosterProfile
+    public class NewlyPublishedAndDownloadCountBoosterProfile
```
joelverhagen: Perhaps we should just call this `DefaultBoosterProfile` since it will iterate methodologies a lot? Or in the future will we add another class each time the approach changes?

As it stands right now, we're making a name that encodes the implementation detail which is a little unmaintainable.

src/NuGet.Services.AzureSearch/ScoringProfiles/NewlyPublishedAndDownloadCountBoosterProfile.cs
```diff
@@ -1,33 +1,42 @@
...
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using Microsoft.Azure.Search.Models;
 using NuGet.Services.AzureSearch.SearchService;
 
 namespace NuGet.Services.AzureSearch.ScoringProfiles
 {
-    public class DownloadCountBoosterProfile
```
joelverhagen: static class?

src/NuGet.Services.AzureSearch/ScoringProfiles/NewlyPublishedAndDownloadCountBoosterProfile.cs
```diff
@@ -1,33 +1,42 @@
...
+                    { IndexFields.PackageId, 10 },              // Exact match of the package id should be boosted the highest.
+                    { IndexFields.TokenizedPackageId, 5 },
                 }
             },
             functions: new List<ScoringFunction>()
             {
+                new FreshnessScoringFunction(
+                    fieldName: IndexFields.Published,
+                    boost: 5.0,
+                    boostingDuration: new TimeSpan(365, 0, 0, 0),
```
joelverhagen: Could you leave code comments about the rational on each of these functions?

joelverhagen: Each `ScoringFunction` that is

src/NuGet.Services.AzureSearch/ScoringProfiles/NewlyPublishedAndDownloadCountBoosterProfile.cs
```diff
@@ -1,33 +1,42 @@
...
                     boost: 100.0,
-                    parameters: new MagnitudeScoringParameters(2000, double.MaxValue, shouldBoostBeyondRangeByConstant: true),
-                    interpolation: ScoringFunctionInterpolation.Logarithmic)
+                    parameters: new MagnitudeScoringParameters(
+                        boostingRangeStart: 0,
+                        boostingRangeEnd: Math.Log(999_999_999_999, 2),
+                        shouldBoostBeyondRangeByConstant: true),
+                    interpolation: ScoringFunctionInterpolation.Linear)
             },
             functionAggregation: ScoringFunctionAggregation.Sum);
```
joelverhagen: Rational on `Sum` too.

src/NuGet.Services.AzureSearch/SearchDocumentBuilder.cs
```diff
@@ -201,6 +201,7 @@ public SearchDocument.LatestFlags LatestFlagsOrNull(VersionLists versionLists, S
                 owners: owners);
             DocumentUtilities.PopulateMetadata(document, packageId, package);
             document.TotalDownloadCount = totalDownloadCount;
+            document.LogBase2DownloadCount = Math.Log(totalDownloadCount, 2.0);
```
joelverhagen: We're real programmer now 🥇

loic-sharma: 👨‍🔬

loic-sharma:
@joelverhagen Would you mind giving a quick peek over the comments? Thanks!

tests/NuGet.Services.AzureSearch.Tests/SearchDocumentBuilderFacts.cs
```diff
@@ -493,6 +493,7 @@ public async Task SetsExpectedProperties(SearchFilters searchFilters, string exp
     {
       ""@search.action"": ""upload"",
       ""totalDownloadCount"": 1001,
+      ""logBase2DownloadCount"": 9.9672262588359928,
```
agr: This looks like it can be a source of test flakiness:
* It assumes that log2 function will always produce certain value (according to [Wolfram Alpha](https://www.wolframalpha.com/input/?i=log2(1001)) the last 2 digits here are wrong);
* It assumes the JSON serialization uses certain precision and rounding rules.

If any of the above changes (with version upgrade), test might start to fail.

src/NuGet.Services.AzureSearch/ScoringProfiles/DefaultScoringProfile.cs
```diff
@@ -0,0 +1,55 @@
...
+                        boostingRangeEnd: Math.Log(999_999_999_999, 2),
+                        shouldBoostBeyondRangeByConstant: true),
+                    interpolation: ScoringFunctionInterpolation.Linear),
+
+                // Boost results with a recent published date. We use a quadatric interpolation
+                // so that the boost decreases faster as the publish date nears the end of the boost range.
+                new FreshnessScoringFunction(
+                    fieldName: IndexFields.Published,
+                    boost: 5.0,
+                    boostingDuration: new TimeSpan(365, 0, 0, 0),
```
agr: `TimeSpan.FromDays(365)`?

tests/NuGet.Services.AzureSearch.Tests/SearchDocumentBuilderFacts.cs
loic-sharma: Let's reassess this if/when this test proves to be flaky.

src/NuGet.Services.AzureSearch/ScoringProfiles/DefaultScoringProfile.cs
```diff
@@ -0,0 +1,55 @@
...
+            },
+            functions: new List<ScoringFunction>()
+            {
+                // Greatly boost results with high download counts. We score off the log of the download count
+                // with linear interpolation so that the boost slows down at higher download counts. We cannot
+                // use the raw download count with a log interpolation as that would result in a large boosting
+                // range, which would need to be offset by an unmanageably high boosting factor.
+                new MagnitudeScoringFunction(
+                    fieldName: IndexFields.Search.LogBase2DownloadCount,
+                    boost: 100.0,
```
cristinamanum: Should these values be configurable in a configuration file?

loic-sharma: Good question. Currently, changing these values requires an index rebuild so I'm leaning more towards keeping these values in code. Please let me know if you feel strongly.

PR #522, opened 2019-05-17
Add deprecation information to the catalog if provided

scottbommarito:
Part of https://github.com/nuget/engineering/issues/2383

This PR adds the ability for deprecation information to flow into the catalog if it is provided by F2C (soon to be DB2C). Currently deprecation information is not fetched so this change will not have any effect.


src/Catalog/PackageCatalogItem.cs
```diff
@@ -94,6 +103,52 @@ public override IGraph CreateContentGraph(CatalogContext context)
...
+                // assert message to deprecation root node
+                if (Deprecation.Message != null)
+                {
+                    graph.Assert(
+                        deprecationRootNode,
+                        graph.CreateUriNode(Schema.Predicates.DeprecationMessage),
+                        graph.CreateLiteralNode(Deprecation.Message));
+                }
+
+                if (Deprecation.AlternatePackage != null)
```
joelverhagen: Our DB allows multiple deprecations. Do think it's right to start with it being an array of deprecations but today only ever have one or zero?

src/Catalog/Schema.cs
```diff
@@ -153,6 +153,13 @@ public static class Predicates
             public static readonly Uri LicenseFile = new Uri(Prefixes.NuGet + "licenseFile");
 
             public static readonly Uri IconFile = new Uri(Prefixes.NuGet + "iconFile");
+
+            public static readonly Uri Deprecation = new Uri(Prefixes.NuGet + "deprecation");
+            public static readonly Uri DeprecationReasons = new Uri(Prefixes.NuGet + "reasons");
+            public static readonly Uri DeprecationMessage = new Uri(Prefixes.NuGet + "message");
+            public static readonly Uri DeprecationAlternatePackage = new Uri(Prefixes.NuGet + "alternatePackage");
+            public static readonly Uri DeprecationAlternatePackageId = new Uri(Prefixes.NuGet + "id");
+            public static readonly Uri DeprecationAlternatePackageVersion = new Uri(Prefixes.NuGet + "version");
```
joelverhagen: This is a version range not version, right?

src/Catalog/PackageDeprecationItem.cs
```diff
@@ -0,0 +1,70 @@
...
+
+                if (version == null)
+                {
+                    throw new ArgumentNullException(nameof(version));
+                }
+
+                Version = new VersionRange(
+                    minVersion: NuGetVersion.Parse(version),
+                    maxVersion: NuGetVersion.Parse(version),
+                    includeMaxVersion: true).ToNormalizedString();
```
joelverhagen: I think we should be explicit about `includeMinVersion: true` for clarity.

src/Catalog/PackageDeprecationItem.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                throw new ArgumentNullException(nameof(reasons));
+            }
+
+            if (!reasons.Any())
+            {
+                throw new ArgumentException(nameof(reasons));
+            }
+
+            Reasons = reasons;
```
joelverhagen: nit: `IReadOnlyList` is a bit more usable

src/Catalog/PackageDeprecationItem.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.Linq;
+using NuGet.Versioning;
+
+namespace NuGet.Services.Metadata.Catalog
+{
+    public class PackageDeprecationItem
+    {
+        public PackageDeprecationItem(
+            IReadOnlyCollection<string> reasons,
+            string message = null,
```
joelverhagen: I think all of these parameters should be explicit since they are all known at the same time from DB, i.e. no defaults are necessary.

src/Catalog/PackageDeprecationItem.cs
```diff
@@ -0,0 +1,70 @@
...
+            Reasons = reasons;
+            Message = message;
+            AlternatePackage = alternatePackage;
+        }
+
+        public IReadOnlyCollection<string> Reasons { get; }
+        public string Message { get; }
+        public PackageDeprecationAlternatePackageItem AlternatePackage { get; }
+
+        public class PackageDeprecationAlternatePackageItem
```
joelverhagen: Nesting classes is not necessary here. Makes things more verbose than is necessary.

src/Catalog/PackageDeprecationItem.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            public PackageDeprecationAlternatePackageItem(
+                string id)
+            {
+                Id = id ?? throw new ArgumentNullException(nameof(id));
+
+                // There is a bug in NuGet.Versioning preventing us from using VersionRange to calculate this.
+                // https://github.com/NuGet/Home/issues/6697
+                Version = "*";
```
joelverhagen: This logic should not be here IMO. This should just be a dumb model that accepts the version range.

src/Catalog/Schema.cs
```diff
@@ -153,6 +153,13 @@ public static class Predicates
             public static readonly Uri LicenseFile = new Uri(Prefixes.NuGet + "licenseFile");
 
             public static readonly Uri IconFile = new Uri(Prefixes.NuGet + "iconFile");
+
+            public static readonly Uri Deprecation = new Uri(Prefixes.NuGet + "deprecation");
+            public static readonly Uri DeprecationReasons = new Uri(Prefixes.NuGet + "reasons");
```
joelverhagen: `deprecationReasons`

joelverhagen: I think all 5 of these latter deprecation predicates should include "deprecation" in their URI

src/Catalog/PackageDeprecationItem.cs
```diff
@@ -0,0 +1,70 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using NuGet.Versioning;
+
+namespace NuGet.Services.Metadata.Catalog
+{
+    public class PackageDeprecationItem
+    {
+        public PackageDeprecationItem(
```
joelverhagen: Can't this be the same entity as what gallery uses?

src/Catalog/PackageCatalogItem.cs
```diff
@@ -21,13 +21,22 @@ public class PackageCatalogItem : AppendOnlyCatalogItem
...
+            PackageDeprecationItem deprecation = null)
         {
             NupkgMetadata = nupkgMetadata;
             CreatedDate = createdDate;
             LastEditedDate = lastEditedDate;
             PublishedDate = publishedDate;
+            Deprecation = deprecation;
         }
 
         public override IGraph CreateContentGraph(CatalogContext context)
```
cristinamanum: Is there an UnitTest for this method?

src/Catalog/PackageCatalogItem.cs
```diff
@@ -94,6 +103,52 @@ public override IGraph CreateContentGraph(CatalogContext context)
...
+                // assert deprecation root node to subject
+                var deprecationPredicate = graph.CreateUriNode(Schema.Predicates.Deprecation);
+                var deprecationRootNode = graph.CreateUriNode(new Uri(resource.Subject.ToString() + "#deprecation"));
+                graph.Assert(resource.Subject, deprecationPredicate, deprecationRootNode);
+
+                // assert reasons to deprecation root node
+                var deprecationReasonRootNode = graph.CreateUriNode(Schema.Predicates.DeprecationReasons);
+                foreach (var reason in Deprecation.Reasons)
+                {
+                    var reasonNode = graph.CreateLiteralNode(reason);
```
cristinamanum: Can the reason be null?

joelverhagen:
Please verify that if there is a single reason, it is still put into an array.

joelverhagen:
If the domain only allows a single deprecation per package, I think we should assert on the graph right before rendering to JSON that there is the proper cardinality of properties (e.g. 0 or 1 deprecation per package, 1 to n reasons, zero or one alternative package IDs, etc).

src/Catalog/Schema.cs
scottbommarito: Yes, but the field is titled "version", which is why I chose this instead of "VersionRange" or similar.

joelverhagen: Hm... okay. Sounds good.

src/Catalog/Schema.cs
scottbommarito: I tried that, but what ended up happening was the object had the shape
```
"deprecation": {
    "deprecationReasons": [ ... ],
    "deprecationMessage": "...",
    "..."
}
```
which I didn't want.

If you know a way to change this URI in a way that makes it unique and keeps the name, I'd love too.

src/Catalog/PackageCatalogItem.cs
scottbommarito: Good point, however, even though we have multiple deprecations, the idea was always for there to only be a single ACTIVE deprecation.

Having multiple deprecations was always intended to be "this is a deprecation I think you should apply to your package" or "this is a deprecation that our automated security scanning added to your package" and not meant to be actually visible.

src/Catalog/PackageDeprecationItem.cs
scottbommarito: I have no problem with that--I mainly did it like this so it was very easy for @xavierdecoster to pass the deprecation info into this code.

src/Catalog/PackageCatalogItem.cs
joelverhagen: I see, so we don't foresee users ever seeing multiple deprecation models on one package at the same time.

src/Catalog/PackageDeprecationItem.cs
scottbommarito: Because we aren't using EF, it seems very odd to me to take a dependency on `NuGet.Services.Entities` only for the purpose of converting the output of a SQL query into an EF model before we serialize it into something else. If it was EF I'd agree.

src/Catalog/PackageCatalogItem.cs
scottbommarito: nope, and `Deprecation.Reasons` throws if it's null or empty.

src/Catalog/Schema.cs
joelverhagen: does `deprecation/reasons` work?

src/Catalog/PackageCatalogItem.cs
scottbommarito: AFAIK the only testing around this is through the `FeedHelpers` tests that I edited below. The methods on `FeedHelpers` that are tested call this method. If I saw tests on this method alone I would have added them.

src/Catalog/Schema.cs
joelverhagen: I don't know all of JSON-LD conventions.

src/Catalog/PackageDeprecationItem.cs
joelverhagen: Okay, sounds good.

src/Catalog/Schema.cs
scottbommarito: I think I tried it but I will double check. If I remember correct it did some very very strange things.

src/Catalog/PackageDeprecationItem.cs
scottbommarito: I wanted to better enforce that the only possible states for a `PackageDeprecationItem` are
- no alternate package
- alternate package with both ID and version specified

Now that I think about it, it would be easier to just enforce those things in the constructor.

joelverhagen: I don't understand how nesting classes and validation are related.

src/Catalog/Schema.cs
scottbommarito: Note: I looked at existing catalog pages and saw that dependency version ranges have `range` as the property name instead of `version`. Do you think I should switch to that syntax? I think it would improve the readability of the code.

src/Catalog/Schema.cs
joelverhagen: Either the field name is made more generic or the URI is made more specific. the current name of the field is very specific but the URI is very generic

src/Catalog/PackageCatalogItem.cs
scottbommarito: Yes, and I think even if we did, we would be combining them to a single visible deprecation for clients to consume.

src/Catalog/PackageDeprecationItem.cs
scottbommarito: I will probably just make this accept a string.

src/Catalog/Schema.cs
joelverhagen: Well, there is the RDF answer and the more down to earth answer for what we want our JSON to look like. You could make arguments either way. I personally think it's wrong to come up with new names for the same thing, but you could argue this is a different thing since package dependencies should not allow `*` but deprecation does.

In short, I think `range` is a more consistent name.

scottbommarito: switching to `range`

src/Catalog/Schema.cs
joelverhagen: Notice all other predicates today have the field name exactly matching the URI suffix. I would go that route, personally by removing "deprecation" from the field name

scottbommarito:
@joelverhagen - yup, individual reasons are not put into an array with this code, will fix

src/Catalog/PackageDeprecationItem.cs
scottbommarito: With a nested class, callers can only provide either null or a `PackageDeprecationAlternatePackageItem`. This enforces that they can't provide an invalid combination of IDs and versions to this constructor...but in retrospect it just moves that logic to the constructor or `PackageDeprecationAlternatePackageItem`. I removed the nested classes to simplify the code.

src/Catalog/Schema.cs
scottbommarito: `#deprecation/reasons` becomes `http://schema.nuget.org/reasons` in the output...I will try to find a better way

src/Catalog/PackageDeprecationItem.cs
```diff
@@ -0,0 +1,46 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NuGet.Services.Metadata.Catalog
+{
+    public class PackageDeprecationItem
+    {
+        public PackageDeprecationItem(
```
skofman1: please add a comment on how input is expected to look

tests/CatalogTests/Helpers/FeedHelpersTests.cs
```diff
@@ -303,12 +303,35 @@ public async Task DownloadMetadata2CatalogAsync_WhenCreatedPackagesIsTrue_WithNo
...
+
         [Theory]
-        [InlineData(true, true)]
-        [InlineData(false, true)]
-        [InlineData(true, false)]
-        [InlineData(false, false)]
-        public async Task DownloadMetadata2CatalogAsync_WithOnePackage_UpdatesStorage(bool createdPackages, bool updateCreatedFromEdited)
+        [MemberData(nameof(DownloadMetadata2CatalogAsync_WithOnePackage_UpdatesStorage_Data))]
+        public async Task DownloadMetadata2CatalogAsync_WithOnePackage_UpdatesStorage(bool createdPackages, bool updateCreatedFromEdited, PackageDeprecationItemState deprecationState)
         {
```
skofman1: add comments to the test:
// Arrange
// Act
// Assert

tests/CatalogTests/Helpers/FeedHelpersTests.cs
```diff
@@ -321,11 +344,44 @@ public async Task DownloadMetadata2CatalogAsync_WithOnePackage_UpdatesStorage(bo
...
+                    case PackageDeprecationItemState.DeprecatedWithAlternateVersion:
+                        deprecationItem = new PackageDeprecationItem(
+                            new[] { "first", "second" },
+                            "this is the message",
+                            "theId", 
+                            "2.4.5");
+                        break;
+                    default:
+                        throw new ArgumentException(nameof(deprecationState));
+                }
```
skofman1: move to a separate method, also, there are some opportunities to reduce code duplication.

tests/CatalogTests/TestData/CatalogTestData.cs
```diff
@@ -132,7 +133,7 @@ internal static class CatalogTestData
   ""authors"": ""James Newton-King"",
   ""catalog:commitId"": ""{1}"",
   ""catalog:commitTimeStamp"": ""{2}"",
-  ""created"": ""{3}"",
+  ""created"": ""{3}"",{6}
```
skofman1: why here? Perhaps under {5}? :)

src/Catalog/Schema.cs
scottbommarito: Figured out a way to do it using `@context`...but it's very ugly and will break if we add any other properties with the same name. I will make the field name more generic instead.

In case you were curious...

In `Schema.cs`:
```
            public static readonly Uri DeprecationReasons = new Uri(Prefixes.NuGet + "deprecationReasons");
            public static readonly Uri DeprecationMessage = new Uri(Prefixes.NuGet + "deprecationMessage");
            public static readonly Uri DeprecationAlternatePackage = new Uri(Prefixes.NuGet + "deprecationAlternatePackage");
            public static readonly Uri DeprecationAlternatePackageId = new Uri(Prefixes.NuGet + "deprecationAlternatePackageId");
            public static readonly Uri DeprecationAlternatePackageRange = new Uri(Prefixes.NuGet + "deprecationAlternatePackageRange");
```

In `PackageDetails.json`:
```
        "reasons" : { "@id": "deprecationReasons", "@container" : "@set" },
        "message" : { "@id": "deprecationMessage" },
        "alternatePackage" : { "@id": "deprecationAlternatePackage" },
        "id" : { "@id": "deprecationAlternatePackageId" },
        "range" : { "@id": "deprecationAlternatePackageRange" }
```

tests/CatalogTests/TestData/CatalogTestData.cs
scottbommarito: The deprecation info gets put here due because JSON-LD alphabetizes the fields.

It was either I do this or rearrange the existing format parameters. I thought this was a smaller additive change.

src/Catalog/PackageDeprecationItem.cs
scottbommarito: done

src/Catalog/PackageCatalogItem.cs
scottbommarito: I found some tests for the method. They seem more geared towards reading a `nupkg` from disk and asserting the output. Unfortunately, however, it seems substantially harder to add this additional logic to there. Given that I'm already testing the behavior in the other tests, I think it's fine not to add it here.

PR #524, opened 2019-05-17
Convert feed2catalog to db2catalog

xavierdecoster:
This PR converts `feed2catalog` into `db2catalog`.

All OData http requests have been replaced by SQL db queries. All package downloads going through the gallery's v2 API have been replaced by HTTP requests to the packages container, which is configurable through the new `-packageContentUrlFormat` commandline argument.

Also added new commandline arguments to support configuring cursor size and sql command timeout: `-cursorSize` and `-sqlCommandTimeoutInSeconds` respectively.


xavierdecoster:
Note: looking into the build failure for the `NgTests.csproj` caused by 

> Error CS1738: Named argument specifications must appear after all fixed arguments have been specified. Please use language version 7.2 or greater to allow non-trailing named arguments.

src/Catalog/Helpers/Db2CatalogSqlQueries.cs
```diff
@@ -0,0 +1,122 @@
...
+            SqlConnection sqlConnection,
+            SqlTransaction sqlTransaction,
+            int top,
+            DateTime since,
+            string columnName)
+        {
+            var nextTimestampsQuery = $@"
+                        SELECT TOP {top} P.[{columnName}]
+                        FROM [dbo].[Packages] AS P
+                        WHERE [{columnName}] IS NOT NULL AND [{columnName}] > @{CursorParameterName}
```
joelverhagen: We should only consider available packages, right?

src/Catalog/Helpers/Db2CatalogSqlQueries.cs
```diff
@@ -0,0 +1,122 @@
...
+                            P.[NormalizedVersion],
+                            P.[Created],
+                            P.[LastEdited],
+                            P.[Published],
+                            P.[Listed],
+                            P.[HideLicenseReport],
+                            P.[LicenseNames],
+                            P.[LicenseReportUrl]
+                        FROM [dbo].[Packages] AS P
+                        INNER JOIN [dbo].[PackageRegistrations] AS PR ON P.[PackageRegistrationKey] = PR.[Key]
```
joelverhagen: Same, only available packages right?

src/Catalog/Helpers/GalleryDatabaseHelper.cs
```diff
@@ -0,0 +1,116 @@
...
+        /// <returns>A task that represents the asynchronous operation.
+        /// The task result (<see cref="Task{TResult}.Result" />) returns an
+        /// <see cref="IList{FeedPackageDetails}" />.</returns>
+        private async Task<IReadOnlyCollection<FeedPackageDetails>> GetPackages(
+            DateTime since,
+            int top,
+            string columnName)
+        {
+            using (var sqlConnection = await _connectionFactory.OpenAsync())
+            using (var sqlTransaction = sqlConnection.BeginTransaction())
```
joelverhagen: We are doing two DB queries now (previous one via feed). How we we know this transaction is giving us the proper level of isolation and therefore self consistency? We need a load test? I am not an expert in isolation levels or SQL lock escalation so I can't see what I need to see to be convinced that this is working okay.

src/Catalog/Helpers/GalleryDatabaseHelper.cs
```diff
@@ -0,0 +1,116 @@
...
+            {
+                var timestamps = await _db2catalogQueries.GetTimestampsAsync(
+                    sqlConnection,
+                    sqlTransaction,
+                    top,
+                    since,
+                    columnName);
+
+                return await _db2catalogQueries.GetPackageDetailsForTimestampsAsync(
```
joelverhagen: Would it be possible to just do the old style query for now and later introduce the `GROUP BY` improvement? Maybe we can phase it in to reduce risk.

src/Catalog/Helpers/PackageContentUriBuilder.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        public PackageContentUriBuilder(string packageContentUrlFormat)
+        {
+            _packageContentUrlFormat = packageContentUrlFormat ?? throw new ArgumentNullException(nameof(packageContentUrlFormat));
+
+            if (!_packageContentUrlFormat.Contains("{id-lower}") || !_packageContentUrlFormat.Contains(VersionLowerPlaceholderString))
+            {
+                throw new ArgumentException(
+                    $"The package content URL format must contain the following placeholders to be valid: {IdLowerPlaceholderString}, {VersionLowerPlaceholderString}. " +
+                    "(e.g. https://api.nuget.org/v3-flatcontainer/{id-lower}/{version-lower}/{id-lower}.{version-lower}.nupkg)",
```
joelverhagen: We should use a more realistic example, i.e. packages container

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -295,48 +295,49 @@ protected override async Task<StorageContent> OnLoadAsync(Uri resourceUri, Cance
             await _throttle.WaitAsync();
             try
             {
-                string content;
-
-                using (var originalStream = new MemoryStream())
+                if (blob.Exists())
```
joelverhagen: Why did this change?

joelverhagen:
Is this based on latest `dev`?

src/Catalog/Helpers/Db2CatalogProjection.cs
```diff
@@ -0,0 +1,49 @@
...
+            var listed = dataRecord.GetBoolean(dataRecord.GetOrdinal("Listed"));
+            var hideLicenseReport = dataRecord.GetBoolean(dataRecord.GetOrdinal("HideLicenseReport"));
+
+            var packageContentUri = _packageContentUriBuilder.Build(
+                packageId.ToLowerInvariant(),
+                normalizedPackageVersion.ToLowerInvariant());
+
+            return new FeedPackageDetails(
+                packageContentUri,
+                dataRecord.ReadNullableUtcDateTime("Created"),
```
joelverhagen: `Created` is not nullable

src/Catalog/Helpers/Db2CatalogProjection.cs
```diff
@@ -0,0 +1,49 @@
...
+                dataRecord.ReadNullableUtcDateTime("LastEdited"),
+                listed ? dataRecord.ReadNullableUtcDateTime("Published") : Constants.UnpublishedDate,
```
joelverhagen: `Published` is not nullable

src/Ng/Helpers/EmptySecretInjector.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Threading.Tasks;
+using NuGet.Services.KeyVault;
+
+namespace Ng.Helpers
+{
+    public class EmptySecretInjector : ISecretInjector
```
joelverhagen: Why is this necessary?

src/Ng/Jobs/Db2CatalogJob.cs
```diff
@@ -119,8 +122,23 @@ protected override void Init(IDictionary<string, string> arguments, Cancellation
...
+            PackageContentUriBuilder = new PackageContentUriBuilder(
+                arguments.GetOrThrow<string>(Arguments.PackageContentUrlFormat));
+
+            var connectionString = arguments.GetOrThrow<string>(Arguments.ConnectionString);
+            GalleryDbConnection = new AzureSqlConnectionFactory(
+                connectionString,
+                new EmptySecretInjector(),
+                LoggerFactory.CreateLogger<AzureSqlConnectionFactory>());
+
+            const int timeoutInSeconds = 300;
```
joelverhagen: IMO this should be configurable. We've had to plumb this to config/params so many times for other DB jobs (like stats)

src/Ng/Jobs/Db2CatalogJob.cs
```diff
@@ -252,42 +270,12 @@ protected override async Task RunInternalAsync(CancellationToken cancellationTok
         }
 
         // Wrapper function for CatalogUtility.CreateHttpClient
-        // Overriden by NgTests.TestableFeed2CatalogJob
+        // Overriden by NgTests.TestableDb2CatalogJob
```
joelverhagen: Use XML docs so you can point to the actual type.

tests/NuGet.Services.BasicSearchTests/NuGet.Services.BasicSearchTests.csproj
```diff
@@ -131,7 +131,7 @@
       <Version>2.28.0</Version>
     </PackageReference>
     <PackageReference Include="xunit">
-      <Version>2.1.0</Version>
+      <Version>2.4.1</Version>
```
joelverhagen: This seems duplicate with https://github.com/NuGet/NuGet.Services.Metadata/pull/523

src/Ng/Jobs/Catalog2DnxJob.cs
```diff
@@ -3,9 +3,11 @@
 
 using System;
 using System.Collections.Generic;
+using System.Net;
```
skofman1: this is the only change in this file.. is this needed?

src/Catalog/Helpers/PackageContentUriBuilder.cs
```diff
@@ -0,0 +1,56 @@
...
+        public const string VersionLowerPlaceholderString = "{version-lower}";
+
+        private const string NuPkgExtension = ".nupkg";
+        private readonly string _packageContentUrlFormat;
+
+        public PackageContentUriBuilder(string packageContentUrlFormat)
+        {
+            _packageContentUrlFormat = packageContentUrlFormat ?? throw new ArgumentNullException(nameof(packageContentUrlFormat));
+
+            if (!_packageContentUrlFormat.Contains("{id-lower}") || !_packageContentUrlFormat.Contains(VersionLowerPlaceholderString))
```
skofman1: >{id-lower}" [](start = 52, length = 11)

use the const you already defined

src/Catalog/Extensions/DateTimeExtensions.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace System
+{
+    public static class DateTimeExtensions
+    {
+        public static DateTime ForceUtc(this DateTime date)
```
cristinamanum: UnitTest?

src/Catalog/Extensions/IDataRecordExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+            {
+                throw new ArgumentNullException(nameof(dataRecord));
+            }
+
+            if (columnName == null)
+            {
+                throw new ArgumentNullException(nameof(columnName));
+            }
+
+            return dataRecord.GetDateTime(dataRecord.GetOrdinal(columnName));
```
cristinamanum: What is the behavior of the column type if not DateTime?

src/Catalog/Helpers/Db2CatalogProjection.cs
```diff
@@ -0,0 +1,49 @@
...
+        public FeedPackageDetails FromDataRecord(IDataRecord dataRecord)
+        {
+            if (dataRecord == null)
+            {
+                throw new ArgumentNullException(nameof(dataRecord));
+            }
+
+            var packageId = dataRecord["Id"].ToString();
+            var normalizedPackageVersion = dataRecord["NormalizedVersion"].ToString();
+            var listed = dataRecord.GetBoolean(dataRecord.GetOrdinal("Listed"));
```
cristinamanum: Do the IDataRecord.GetBoolean throw if the conversion fails or return a default value? If silently fails is there any way to ensure the expected type?

src/Catalog/Helpers/Db2CatalogSqlQueries.cs
skofman1: Yup. Make sure not to grab packages in "Validating" state.

---
In reply to: [285178974](https://github.com/NuGet/NuGet.Services.Metadata/pull/524#discussion_r285178974) [](ancestors = 285178974)

src/Catalog/Extensions/DateTimeExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+        {
+            if (date.Kind == DateTimeKind.Unspecified)
```
cristinamanum: The name may imply there the date will be changed to UTC. This is correct only if the DateTimeKind in Unspecified. What about Local?

src/Catalog/Helpers/Db2CatalogSqlQueries.cs
```diff
@@ -0,0 +1,122 @@
...
+                while (await timestampsReader.ReadAsync())
+                {
+                    timestamps.Add(timestampsReader.ReadDateTime(columnName).ForceUtc());
+                }
+            }
+
+            return timestamps;
+        }
+
+        public async Task<IReadOnlyCollection<FeedPackageDetails>> GetPackageDetailsForTimestampsAsync(
```
skofman1: should be renamed (perhaps separate PR)

src/Catalog/Helpers/Db2CatalogSqlQueries.cs
```diff
@@ -0,0 +1,122 @@
...
+            _db2catalogProjection = new Db2CatalogProjection(packageContentUriBuilder);
+            _commandTimeout = commandTimeout;
+        }
+
+        public async Task<IReadOnlyCollection<DateTime>> GetTimestampsAsync(
+            SqlConnection sqlConnection,
+            SqlTransaction sqlTransaction,
+            int top,
+            DateTime since,
+            string columnName)
```
cristinamanum: What would be the behavior for random columnNames? Should the column names be restricted?

src/Catalog/Helpers/Db2CatalogSqlQueries.cs
```diff
@@ -0,0 +1,122 @@
...
+            return timestamps;
+        }
+
+        public async Task<IReadOnlyCollection<FeedPackageDetails>> GetPackageDetailsForTimestampsAsync(
+            SqlConnection sqlConnection,
+            SqlTransaction sqlTransaction,
+            string columnName,
+            IReadOnlyCollection<DateTime> timestamps)
+        {
+            var packages = new List<FeedPackageDetails>();
```
cristinamanum: Can this lead to OutOfMemory? Should we consider a max count of Packages retrieved?

src/Catalog/Helpers/GalleryDatabaseHelper.cs
```diff
@@ -0,0 +1,116 @@
...
+
+            return OrderPackagesByKeyDate(allPackages, keyDateFunc);
+        }
+
+
+        /// <summary>
+        /// Returns a <see cref="SortedList{DateTime, IList{FeedPackageDetails}}"/> of packages.
+        /// </summary>
+        /// <param name="keyDateFunc">The <see cref="DateTime"/> field to sort the <see cref="FeedPackageDetails"/> on.</param>
+        public static SortedList<DateTime, IList<FeedPackageDetails>> OrderPackagesByKeyDate(
```
cristinamanum: UnitTest?

src/Catalog/Helpers/Db2CatalogSqlQueries.cs
```diff
@@ -0,0 +1,122 @@
...
+                var packagesThisTimestamp = new List<FeedPackageDetails>();
+
+                var packagesCommand = new SqlCommand(packageQuery, sqlConnection, sqlTransaction)
+                {
+                    CommandTimeout = _commandTimeout
+                };
+
+                packagesCommand.Parameters.AddWithValue("Cursor", timestamp);
+
+                using (var packagesReader = await packagesCommand.ExecuteReaderAsync())
```
skofman1: correct me if I'm wrong, this might execute 20 SQL queries in a single loop (if we read 20 timestamps, and run a query per timestamp. 

Wouldn't it be more performant to create a single query? There are many options to achieve this. One example: get packages in the range between min timestamp and max timestamp. Also, MaxPageSize can be part of the query (you might need to add logic to ignore packages from the last timestamp if there's more then page size).

src/Catalog/Helpers/Db2CatalogProjection.cs
```diff
@@ -0,0 +1,49 @@
...
+            var hideLicenseReport = dataRecord.GetBoolean(dataRecord.GetOrdinal("HideLicenseReport"));
+
+            var packageContentUri = _packageContentUriBuilder.Build(
+                packageId.ToLowerInvariant(),
```
skofman1: you can drop the ToLowerInvariant since it's already done by the implementation of PackageContentUriBuilder

src/Ng/CommandHelpers.cs
```diff
@@ -97,13 +95,16 @@ public static void AssertAzureStorage(IDictionary<string, string> arguments)
...
 
-        public static RegistrationStorageFactories CreateRegistrationStorageFactories(IDictionary<string, string> arguments, bool verbose)
+        public static RegistrationStorageFactories CreateRegistrationStorageFactories(
+            IDictionary<string, string> arguments,
+            bool verbose,
+            IThrottle throttle = null)
         {
             CatalogStorageFactory legacyStorageFactory;
-            var semVer2StorageFactory = CreateSemVer2StorageFactory(arguments, verbose);
+            var semVer2StorageFactory = CreateSemVer2StorageFactory(arguments, verbose, throttle);
```
skofman1: you should merge with latest dev..

src/Ng/CommandHelpers.cs
```diff
@@ -97,13 +95,16 @@ public static void AssertAzureStorage(IDictionary<string, string> arguments)
             }
         }
 
-        public static RegistrationStorageFactories CreateRegistrationStorageFactories(IDictionary<string, string> arguments, bool verbose)
+        public static RegistrationStorageFactories CreateRegistrationStorageFactories(
+            IDictionary<string, string> arguments,
+            bool verbose,
+            IThrottle throttle = null)
```
cristinamanum: Where is this(without null value) used ?

src/Catalog/Extensions/DateTimeExtensions.cs
xavierdecoster: I just moved this code from [here](https://github.com/NuGet/NuGet.Services.Metadata/pull/524/files/f358a0d4e8c8a844f32097e9499d8bd70090ce16#diff-ef9bbac124ec39795cf902ad6b03bfc2L161). As no tests existed for it, I'll create them.

src/Catalog/Extensions/DateTimeExtensions.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/b0ab67f19470e9ad3e6e9b6c9e6cd85817e18bcc

src/Catalog/Extensions/DateTimeExtensions.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/b0ab67f19470e9ad3e6e9b6c9e6cd85817e18bcc

src/Catalog/Helpers/GalleryDatabaseHelper.cs
xavierdecoster: Currently, the default READ COMMITTED isolation level is applied. I think, however, that we should be using SNAPSHOT isolation: https://docs.microsoft.com/en-us/sql/t-sql/statements/set-transaction-isolation-level-transact-sql?view=sql-server-2017

I'm no expert either which is why it is good to get this reviewed ;)
Maybe @dtivel wants to have a say in this?

src/Catalog/Helpers/PackageContentUriBuilder.cs
xavierdecoster: Oops. Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/63879c738456c91777a2798a7f2e2e6b0e260224

tests/NuGet.Services.BasicSearchTests/NuGet.Services.BasicSearchTests.csproj
xavierdecoster: Hence the separate PR ;) merged with dev.

src/Catalog/Helpers/PackageContentUriBuilder.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/63879c738456c91777a2798a7f2e2e6b0e260224

src/Catalog/Helpers/Db2CatalogProjection.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/eef8cc37c3f36b12511305ed4d7608899ef04d90

src/Catalog/Helpers/Db2CatalogProjection.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/d9f99d7c369333fa52e42b7ef95f5b43ace38063

src/Catalog/Helpers/Db2CatalogProjection.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/d9f99d7c369333fa52e42b7ef95f5b43ace38063

src/Catalog/Persistence/AzureStorage.cs
xavierdecoster: This change looks unintented. Adding @scottbommarito who did this commit on this branch fd1ebbbb608d6bf89019677da161bb5697938b3b

I think I can revert the changes to this file as the throttling change went in via separate PR (https://github.com/NuGet/NuGet.Services.Metadata/pull/520/files), but want to get confirmation from Scott first.

src/Ng/CommandHelpers.cs
xavierdecoster: I did. May be a left-over of an earlier throttling implementation on this branch by @scottbommarito before the change went into `dev` via separate PR. See also my comment [here](https://github.com/NuGet/NuGet.Services.Metadata/pull/524/files#r285221099).

src/Catalog/Persistence/AzureStorage.cs
scottbommarito: Yeah I think I must have accidentally reverted the changes to this file from https://github.com/NuGet/NuGet.Services.Metadata/pull/514 when pulling it into my branch.

src/Ng/Helpers/EmptySecretInjector.cs
xavierdecoster: This job has not been converted yet to use the configuration JSON files to be used by the secret reader.

src/Ng/Jobs/Db2CatalogJob.cs
xavierdecoster: This is an old comment. Also, the jobs don't reference the test projects ;)

build.ps1
```diff
@@ -117,7 +117,7 @@ Invoke-BuildStep 'Creating artifacts' {
             "src\Ng\Catalog2Lucene.nuspec", `
             "src\Ng\Catalog2Monitoring.nuspec", `
             "src\Ng\Catalog2Registration.nuspec", `
-            "src\Ng\Feed2Catalog.nuspec", `
+            "src\Ng\Db2Catalog.nuspec", `
```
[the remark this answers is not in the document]
xavierdecoster: Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/34aa8e2ef94a48633ec4a3558a45ef025eef7396

src/Catalog/Extensions/IDataRecordExtensions.cs
xavierdecoster: That would depend on the concrete implementation of `IDataRecord.GetDateTime`.

src/Catalog/Helpers/Db2CatalogSqlQueries.cs
```diff
@@ -0,0 +1,122 @@
...
+using System.Data.SqlClient;
+using System.Data.SqlTypes;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.Metadata.Catalog.Helpers
+{
+    /// <summary>
+    /// Utility class for all SQL queries invoked by Db2Catalog.
+    /// </summary>
+    public class Db2CatalogSqlQueries : IDb2CatalogQueries
```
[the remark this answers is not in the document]
xavierdecoster: Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/b494ec6629999bdde8eb6a42f612ef42ac36f817

src/Ng/CommandHelpers.cs
xavierdecoster: Fixed!

src/Catalog/Helpers/Db2CatalogSqlQueries.cs
xavierdecoster: Restricted in refactoring. See https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/6ff842b66eeacec3f6e19a0dd57a5936bc911173

src/Catalog/Helpers/Db2CatalogSqlQueries.cs
xavierdecoster: Obsolete after refactoring.

src/Catalog/Helpers/Db2CatalogSqlQueries.cs
xavierdecoster: Good catch. Fixed.

src/Catalog/Helpers/Db2CatalogSqlQueries.cs
xavierdecoster: Fixed.

src/Catalog/Helpers/GalleryDatabaseHelper.cs
xavierdecoster: No more need for read-only transaction; now is a single SQL query.

src/Catalog/Persistence/AzureStorage.cs
xavierdecoster: Fixed.

src/Ng/Jobs/Db2CatalogJob.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/0a025e89b7819fd5991d4d5980e222acf8deeecb

src/Catalog/Helpers/Db2CatalogProjection.cs
xavierdecoster: Depends on the concrete implementation. In the case of `SqlDataReader`, an `InvalidCastException` would be thrown if the data type doesn't match (no conversions happen): https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqldatareader.getboolean?view=netframework-4.8

src/Catalog/Helpers/GalleryDatabaseHelper.cs
xavierdecoster: This is existing code which lacked test coverage. Added a test asserting behavior in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/67cd6d2138b28d6fe984e4c27be3c1def5df9bf1.

src/Catalog/Helpers/GalleryDatabaseQueryService.cs
```diff
@@ -0,0 +1,150 @@
...
+                CommandTimeout = _commandTimeout
+            };
+
+            packagesCommand.Parameters.AddWithValue(CursorParameterName, cursor.CursorValue);
+
+            using (var packagesReader = await packagesCommand.ExecuteReaderAsync())
+            {
+                while (await packagesReader.ReadAsync())
+                {
+                    packages.Add(_db2catalogProjection.FromDataRecord(packagesReader));
```
scottbommarito: One note about this implementation--it ALWAYS grabs the next 20 packages with ties, even if the last timestamp has 10,000 packages sharing it. In the previous implementation, it would have grabbed the next 19 packages and then saved the last timestamp for a batch on its own. I'm not sure about the implications of this--if the 19 packages on their own are massive (e.g. have a lot of versions) this could wreck our pipeline. Could we make it so that it splits large timestamps into their own batch?

xavierdecoster: Yes, that's indeed something to fix still. Are you suggesting to reintroduce the Group By in this single query?

scottbommarito: No, I'm suggesting that after we get the data from the server we filter it out--I think keeping the single query behavior is most important.

You could probably do it in [`OrderPackagesByKeyDate`](https://github.com/NuGet/NuGet.Services.Metadata/pull/524/files#diff-9238f95b1acac0dc5e93d264bfd09e36R65) somewhere.

xavierdecoster: Addressed with https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/dc4291d70161ac97bda66d952962c468eafbfede

src/Catalog/Helpers/GalleryDatabaseQueryService.cs
```diff
@@ -0,0 +1,166 @@
...
+                        P.[Created],
+                        P.[LastEdited],
+                        P.[Published],
+                        P.[Listed],
+                        P.[HideLicenseReport],
+                        P.[LicenseNames],
+                        P.[LicenseReportUrl]
+                    FROM [dbo].[Packages] AS P
+                    INNER JOIN [dbo].[PackageRegistrations] AS PR ON P.[PackageRegistrationKey] = PR.[Key]
+                    WHERE P.[PackageStatusKey] = {PackageStatus.Available}
```
joelverhagen: Does `{PackageStatus.Available}` become `0` or `Available`?

src/Catalog/Helpers/GalleryDatabaseQueryService.cs
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        private async Task<IReadOnlyCollection<FeedPackageDetails>> GetPackageDetailsAsync(
+            SqlConnection sqlConnection,
+            DB2CatalogCursor cursor)
+        {
+            var packages = new List<FeedPackageDetails>();
+            var packageQuery = BuildDB2CatalogSqlQuery(cursor);
+
+            var packagesCommand = new SqlCommand(packageQuery, sqlConnection)
```
joelverhagen: Dispose

src/Catalog/Helpers/PackageContentUriBuilder.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        public PackageContentUriBuilder(string packageContentUrlFormat)
+        {
+            _packageContentUrlFormat = packageContentUrlFormat ?? throw new ArgumentNullException(nameof(packageContentUrlFormat));
+
+            if (!_packageContentUrlFormat.Contains(IdLowerPlaceholderString) || !_packageContentUrlFormat.Contains(VersionLowerPlaceholderString))
+            {
+                throw new ArgumentException(
+                    $"The package content URL format must contain the following placeholders to be valid: {IdLowerPlaceholderString}, {VersionLowerPlaceholderString}. " +
+                    "(e.g. https://storageaccountname.blob.core.windows.net/packages/{id-lower}.{version-lower}.nupkg)",
```
joelverhagen: nit: use constants in exception message

src/Catalog/NuGet.Services.Metadata.Catalog.csproj
```diff
@@ -252,12 +259,18 @@
     <PackageReference Include="dotNetRDF">
       <Version>1.0.8.3533</Version>
     </PackageReference>
-    <PackageReference Include="NuGet.Packaging">
-      <Version>5.0.0-preview1.5665</Version>
+    <PackageReference Include="NuGet.Services.Entities">
+      <Version>2.41.0</Version>
```
joelverhagen: NuGet.Services.Entities is not 4.4.X versions since it is made in gallery. We should be pulling in latest of these guys.

src/Catalog/NuGet.Services.Metadata.Catalog.csproj
```diff
@@ -252,12 +259,18 @@
...
+    </PackageReference>
+    <PackageReference Include="NuGet.Services.Sql">
+      <Version>2.46.0</Version>
```
joelverhagen: Is this version consistent with the rest of the sln?

src/Ng/Helpers/EmptySecretInjector.cs
joelverhagen: Copyright header

Also, can you leave an XML doc explaining that the secret injection is done at a higher level and therefore this type doesn't need to do anything

src/Ng/NgJobFactory.cs
```diff
@@ -14,14 +14,14 @@ public static class NgJobFactory
         public static IDictionary<string, Type> JobMap = new Dictionary<string, Type>()
         {
             { "package2catalog", typeof(Package2CatalogJob) },
-            { "feed2catalog", typeof(Feed2CatalogJob) },
+            { "db2catalog", typeof(DB2CatalogJob) },
```
joelverhagen: Should I change Db2AzureSearch to DB2AzureSearch?

nit: tbh I've seen "Db" in our code more than "DB". For example `DbSet`, `GalleryDb` config section name, `TestDbAsyncQueryProvider` for mocking. Because of this, I think we should prefer `Db2Catalog` + `Db2Lucene` instead of `DB2Catalog` + `DB2Lucene`.

tests/NgTests/Infrastructure/TestLogger.cs
```diff
@@ -0,0 +1,43 @@
...
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace NgTests.Infrastructure
+{
+    internal class TestLogger : ILogger
+    {
+        public void Log(LogLevel logLevel, int eventId, object state, Exception exception, Func<object, Exception, string> formatter)
+        {
+            // Console.WriteLine($"{logLevel}: {formatter(state, exception)}");
```
joelverhagen: A test logger should accept Xunit's `ITestOutputHelper` right?

src/Ng/Jobs/DB2CatalogJob.cs
```diff
@@ -9,39 +9,42 @@
...
+        protected ISqlConnectionFactory GalleryDbConnection;
+        protected PackageContentUriBuilder PackageContentUriBuilder;
+        protected IGalleryDatabaseQueryService GalleryDatabaseQueryService;
         protected IStorage CatalogStorage;
         protected IStorage AuditingStorage;
         protected IStorage PreferredPackageSourceStorage;
         protected DateTime? StartDate;
         protected TimeSpan Timeout;
-        protected int Top;
+        protected int Top = 20;
```
joelverhagen: can this be a const?

src/Ng/Jobs/DB2CatalogJob.cs
```diff
@@ -303,8 +289,7 @@ private static Uri MakeLastEditedUri(string source, DateTime since, int top)
                 if (!string.IsNullOrEmpty(auditEntry.PackageId) && !string.IsNullOrEmpty(auditEntry.PackageVersion) && auditEntry.TimestampUtc > since)
                 {
                     // Mark the package "deleted"
-                    IList<FeedPackageIdentity> packages;
-                    if (!result.TryGetValue(auditEntry.TimestampUtc.Value, out packages))
+                    if (!result.TryGetValue(auditEntry.TimestampUtc.Value, out IList<FeedPackageIdentity> packages))
```
joelverhagen: super nit: you can do `out var packages`

src/Ng/Jobs/Monitoring2MonitoringJob.cs
```diff
@@ -4,9 +4,11 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
```
joelverhagen: Are these necessary changes?

src/Ng/Jobs/MonitoringProcessorJob.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
```
joelverhagen: Is this a necessary change?

src/Catalog/Helpers/GalleryDatabaseQueryService.cs
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        /// <summary>
+        /// Builds the SQL query string for db2catalog.
+        /// </summary>
+        /// <param name="cursor">The <see cref="DB2CatalogCursor"/> to be used.</param>
+        /// <returns>The SQL query string for the db2catalog job, build from the the provided <see cref="DB2CatalogCursor"/>.</returns>
+        internal static string BuildDB2CatalogSqlQuery(DB2CatalogCursor cursor)
+        {
+            return $@"SELECT TOP {cursor.Top} WITH TIES
```
joelverhagen: Could you paste the SQL generated by gallery for the `LastEdited`/`Created` OData queries? It's probably worth having multiple eyes on the existing SQL to made sure it's really 1:1

src/Catalog/Helpers/DB2CatalogCursor.cs
```diff
@@ -0,0 +1,34 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Data.SqlTypes;
+
+namespace NuGet.Services.Metadata.Catalog.Helpers
+{
+    public sealed class DB2CatalogCursor
```
skofman1: nit: since this is used only by GalleryDatabaseQueryService it can be an internal class inside it

src/Ng/NgJobFactory.cs
scottbommarito: `DB2AzureSearch` is the correct pascalcase but it's frequently mistaken. I don't think it's a big deal--if it's a ton of work to switch to it, I wouldn't worry about it, but if it's at the stage that you could quickly change to it then I think you should.

[Basically, abbreviations with only 2 letters are both capitalized with pascalcase.](https://stackoverflow.com/questions/15526107/acronyms-in-camelcase) I remember I used to get this wrong as well until @shishirx34 showed me.

joelverhagen: Then why did EF do it `Db`?

joelverhagen: In short, I think we should be consistent with what we already have (`Db`) -- unless we have many places that are also `DB` then I guess it's a toss-up.

scottbommarito: EF likely got it wrong too...it's a common mistake.

I don't think it's a big deal but `DB` is technically correct. For example, `System.IO` is not `System.Io`. I don't think we should be resistant to using the correct syntax just because we've done it incorrectly in the past. I'm fine with whatever we choose. Xavier did have it as `Db2Catalog` before and I asked him to change it to `DB2Catalog` so I feel bad for having him change it back.

src/Catalog/Helpers/GalleryDatabaseQueryService.cs
scottbommarito: The OData query is likely going to pull in a lot of unnecessary fields because the shape of the object is formed on the client-side--I don't think it's a big deal if it's not 1:1.

joelverhagen: True. I'm more interested in any DML that changes the number of records in the results (e.g. `WHERE`)

src/Catalog/NuGet.Services.Metadata.Catalog.csproj
scottbommarito: The latest version should be `4.4.5`, right?

src/Ng/Jobs/DB2CatalogJob.cs
scottbommarito: IMO it should be configurable--20 is a very arbitrary number and I think we could get slightly better perf out of the pipeline if we increased it--but that's out of scope of this PR.

src/Ng/NgJobFactory.cs
scottbommarito: Talked offline. Although the Microsoft guidelines on pascalcase seem to imply that `DB` is the right casing, `Db` seems to be used everywhere, in not only EF but also `System.Data.Common`. Given that, it seems like being consistent is a much better choice here.

tests/CatalogTests/Helpers/DB2CatalogCursorTests.cs
```diff
@@ -0,0 +1,63 @@
...
+using System;
+using System.Data.SqlTypes;
+using NuGet.Services.Metadata.Catalog.Helpers;
+using Xunit;
+
+namespace CatalogTests.Helpers
+{
+    public class DB2CatalogCursorTests
+    {
+        public class TheByCreatedMethod
```
scottbommarito: nit: these tests are identical between the two methods, you could de-dupe them by using inheritance or `Theory`

tests/CatalogTests/Helpers/GalleryDatabaseQueryServiceTests.cs
```diff
@@ -0,0 +1,279 @@
...
+using NuGet.Services.Entities;
+using NuGet.Services.Metadata.Catalog.Helpers;
+using Xunit;
+using CatalogConstants = NuGet.Services.Metadata.Catalog.Constants;
+
+namespace CatalogTests.Helpers
+{
+    public class GalleryDatabaseQueryServiceTests
+    {
+        public class TheBuildDB2CatalogSqlQueryMethod
```
scottbommarito: If we're just asserting each line of the query in a different test, why don't we just have a single test that asserts the exact query?

src/Catalog/Helpers/DB2CatalogCursor.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/ca3107303239c439127b30fb3b1c9ba507c64843

src/Ng/Jobs/MonitoringProcessorJob.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/81d386ceb95e2f5839d6e63029ee23ca1b22326b

src/Ng/Jobs/Monitoring2MonitoringJob.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/81d386ceb95e2f5839d6e63029ee23ca1b22326b

src/Ng/Jobs/DB2CatalogJob.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/81d386ceb95e2f5839d6e63029ee23ca1b22326b

src/Catalog/Helpers/GalleryDatabaseQueryService.cs
xavierdecoster: OMG this would 've been horrible! Great catch! Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/6e5c213ee5e0a0cd44229a5f964c16baaf7be007

src/Catalog/Helpers/PackageContentUriBuilder.cs
xavierdecoster: They're all constants. Maybe the example URL placeholders are causing confusion?

src/Catalog/NuGet.Services.Metadata.Catalog.csproj
xavierdecoster: AFAICT the latest version of NuGet.Services.Entities is `2.41.0`.

src/Catalog/NuGet.Services.Metadata.Catalog.csproj
xavierdecoster: Yep!

src/Ng/NgJobFactory.cs
xavierdecoster: Ok, reverting it all to `Db` lower case :)

tests/CatalogTests/Helpers/GalleryDatabaseQueryServiceTests.cs
xavierdecoster: Readability: because the method name clearly states why that line is tested in the test report. Of course this could all be a single test. Should I merge them into one? Or would you agree there's value in these separate assertions? I typically try to assert only one thing per test.

tests/CatalogTests/Helpers/DB2CatalogCursorTests.cs
xavierdecoster: Had to make the `Db2CatalogCursor` a public type again for this, but fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/22da5bfc272abc369d36ac8a05eeb1e7f210c607

src/Ng/NgJobFactory.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/33591d3f6ae50b3af5a9d637a7339a5f0db5742b and https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/f21017a88db277bb418a977f507bf96f2525af7a

src/Ng/Helpers/EmptySecretInjector.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/3e5c1faea9e244e35d5f69b516630ff94ee90935

src/Catalog/Helpers/GalleryDatabaseQueryService.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/6a4b503ac01e97f319c50333e653fafd0fed1630

tests/NgTests/Infrastructure/TestLogger.cs
xavierdecoster: This is existing code I just moved to its own file :) But yes, this should accept Xunit's `ITestOutputHelper`, will fix.

xavierdecoster: Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/2db90127244969eb8cb3e67787811c569b81d4b0

src/Ng/Jobs/DB2CatalogJob.cs
xavierdecoster: Made it configurable using commandline argument `-cursorSize`. Defaults to `20`. 
See https://github.com/NuGet/NuGet.Services.Metadata/pull/524/commits/8a0b180295c72ae523d1cd4e78db8716565759c7

src/Catalog/Helpers/GalleryDatabaseQueryService.cs
xavierdecoster: The OData queries had the following format:
`https://www.nuget.org/api/v2/Packages?$filter=Created%20gt%20DateTime%272019-05-24T00:00:00%27&$orderby=Created&$top=20&$select=Id,NormalizedVersion,Created,LastEdited,Published,LicenseNames,LicenseReportUrl&semVerLevel=2.0.0`

The generated SQL query looked like this:

```
DECLARE @p__linq__5 AS SQL_VARIANT;
DECLARE @p__linq__6 AS SQL_VARIANT;
DECLARE @p__linq__4 AS SQL_VARIANT;
SET @p__linq__5 = NULL;
SET @p__linq__6 = NULL;
SET @p__linq__4 = NULL;

SELECT TOP (20) 
    [Project1].[PackageRegistrationKey] AS [PackageRegistrationKey], 
    [Project1].[C3] AS [C1], 
    [Project1].[C4] AS [C2], 
    [Project1].[Version] AS [Version], 
    [Project1].[C5] AS [C3], 
    [Project1].[C2] AS [C4], 
    [Project1].[C6] AS [C5], 
    [Project1].[C1] AS [C6], 
    [Project1].[C7] AS [C7], 
    [Project1].[C8] AS [C8], 
    [Project1].[C9] AS [C9], 
    [Project1].[LastEdited] AS [LastEdited], 
    [Project1].[C10] AS [C10], 
    [Project1].[C11] AS [C11], 
    [Project1].[C12] AS [C12], 
    [Project1].[NormalizedVersion] AS [NormalizedVersion], 
    [Project1].[C13] AS [C13], 
    [Project1].[Id] AS [Id]
    FROM ( SELECT 
        [Extent1].[PackageRegistrationKey] AS [PackageRegistrationKey], 
        [Extent1].[Created] AS [Created], 
        [Extent1].[LastEdited] AS [LastEdited], 
        [Extent1].[Version] AS [Version], 
        [Extent1].[NormalizedVersion] AS [NormalizedVersion], 
        [Extent2].[Id] AS [Id], 
        CASE WHEN ((@p__linq__5 <> 1) OR ([Extent1].[HideLicenseReport] = 1)) THEN CAST(NULL AS varchar(1)) ELSE [Extent1].[LicenseNames] END AS [C1], 
        CASE WHEN ((@p__linq__6 <> 1) OR ([Extent1].[HideLicenseReport] = 1)) THEN CAST(NULL AS varchar(1)) ELSE [Extent1].[LicenseReportUrl] END AS [C2], 
        N'40103e1f-cbe5-4f2e-bad3-4622bf14a541' AS [C3], 
        N'Version' AS [C4], 
        N'LicenseReportUrl' AS [C5], 
        N'LicenseNames' AS [C6], 
        N'Published' AS [C7], 
         CAST( CASE WHEN ([Extent1].[Listed] = 1) THEN [Extent1].[Published] ELSE @p__linq__4 END AS datetime2) AS [C8], 
        N'LastEdited' AS [C9], 
        N'Created' AS [C10], 
         CAST( [Extent1].[Created] AS datetime2) AS [C11], 
        N'NormalizedVersion' AS [C12], 
        N'Id' AS [C13]
        FROM  [dbo].[Packages] AS [Extent1]
        INNER JOIN [dbo].[PackageRegistrations] AS [Extent2] ON [Extent1].[PackageRegistrationKey] = [Extent2].[Key]
        WHERE (0 = [Extent1].[PackageStatusKey]) AND ([Extent1].[SemVerLevelKey] = 2 OR [Extent1].[SemVerLevelKey] IS NULL) AND ([Extent1].[Created] > convert(datetime2, '2018-01-01 00:00:00.0000000', 121))
    )  AS [Project1]
    ORDER BY [Project1].[Created] ASC
```

The only difference I see in the `WHERE` clause is the SemVer2 related filter, but I think we can safely ignore that.

joelverhagen: Awesome! Thanks for gathering this.

PR #525, opened 2019-05-17
Add deprecation metadata to registration when it is present in catalog

scottbommarito:
https://github.com/nuget/engineering/issues/2384


joelverhagen:
Can we have some unit tests that assert the actual JSON shape in the deprecated an non deprecated cases?

loic-sharma:
image

scottbommarito:
@joelverhagen - I found some tests (written by @dtivel) that I can add to--stay tuned!

scottbommarito:
@joelverhagen - I finished writing the tests and also caught a bug in the process. It's ready for reviews again!

tests/CatalogTests/Helpers/RegistrationPackageDeprecationDetails.cs
```diff
@@ -0,0 +1,50 @@
...
+        public string[] Reasons { get; }
+
+        [JsonProperty(CatalogConstants.Message)]
+        public string Message { get; }
+
+        [JsonProperty(CatalogConstants.AlternatePackage)]
+        public RegistrationPackageDeprecationAlternatePackage AlternatePackage { get; }
+    }
+
+    public class RegistrationPackageDeprecationAlternatePackage
```
joelverhagen: Can we stick to one class per file please? makes github finding easier

agr: Sounds like a github could use some search improvements :)

scottbommarito: @anangaur - pass it on!

PR #526, opened 2019-05-21
Search service - add resilience for AAD tenant migrations

skofman1:
What do you think about this approach to resiliency? IMO the benefit here is that the code change is minimal, however, this is a hack.
Another approach can be to use cached configuration if we fail to retrieve from KeyVault.


src/NuGet.Services.BasicSearch/Startup.cs
```diff
@@ -210,6 +212,23 @@ public void Configuration(IAppBuilder app)
             Configuration(app, new ConfigurationFactory(configurationProvider), null, null);
         }
 
+        private async Task<BasicSearchConfiguration> GetConfiguration()
+        {
+            try
+            {
+                return await _configFactory.Get<BasicSearchConfiguration>();
+            }
+            catch (KeyVaultClientException e) //when (e.Status == HttpStatusCode.Unauthorized || e.Status == HttpStatusCode.Forbidden)
```
joelverhagen: Let's move this `when` to a comment in English. This just looks like forgotten code

skofman1:
Fixes https://github.com/nuget/engineering/issues/2412

PR #530, opened 2019-05-27
Add additional telemetry for SQL queries invoked by Db2Catalog

xavierdecoster:


src/Catalog/Helpers/GalleryDatabaseQueryService.cs
```diff
@@ -153,18 +156,32 @@ private async Task<IReadOnlyCollection<FeedPackageDetails>> GetPackages(Db2Catal
             {
                 packagesCommand.Parameters.AddWithValue(CursorParameterName, cursor.CursorValue);
 
-                using (var packagesReader = await packagesCommand.ExecuteReaderAsync())
+                var properties = GetTelemetryProperties(cursor);
+                using (_telemetryService.TrackDuration(TelemetryConstants.GetPackageDetailsSeconds, properties))
```
agr: I would wrap the `TrackDuration` call into the helper method instead of wrapping the dictionary generation, so it would look like `using (TrackGetPackageDetailsDuration(cursor)) {...}` or `using (TrackGetPackageDetailsDuration(_telemetryService, cursor)) {...}`. We don't have a wrapper telemetry service there like we have in [Orchestrator](https://github.com/NuGet/NuGet.Jobs/blob/master/src/NuGet.Services.Validation.Orchestrator/Telemetry/TelemetryService.cs) for example, right?

loic-sharma: Another possibility is to create a new interface with high level methods like `TrackGetPackageDetailsDuration`. Your code would become `using (_newTelemetryService.TrackGetPackageDetailsDuration(cursor))` which can be unit tested easily.

xavierdecoster: Updated in https://github.com/NuGet/NuGet.Services.Metadata/pull/530/commits/aabd9fa13f6058a32b30087aaa70a7b39b316201

PR #531, opened 2019-05-28
Upgrade NuGetGallery.Core dependency in NuGet.Indexing

xavierdecoster:
Required to get the following changes:
* Entities have moved out of `NuGetGallery.Core` into `NuGet.Services.Entities`
* `CVE` and `CWE` entities no longer exist but are still referenced by outdated `IEntitiesContext` interface
* `ICloudBlobContainer.ExistsAsync` public API has breaking change

This will unblock adoption of latest `NuGet.Services.Entities` in the solution, required to bring in package deprecation support to `ng.exe` jobs.


src/NuGet.Services.AzureSearch/BlobContainerBuilder.cs
```diff
@@ -84,9 +84,9 @@ public async Task CreateAsync(bool retryOnConflict)
             _logger.LogInformation("Done creating blob container {ContainerName}.", _options.Value.StorageContainer);
         }
 
-        public async Task CreateIfNotExistsAsync()
+        public async Task CreateIfNotExistsAsync(BlobRequestOptions options = null, OperationContext operationContext = null)
```
joelverhagen: These parameters are not needed in this context. Null can be passed for both of these on line 89.

src/NuGet.Indexing/NuGet.Indexing.csproj
```diff
@@ -190,7 +190,7 @@
       <Version>2.2.2</Version>
     </PackageReference>
     <PackageReference Include="NuGetGallery.Core">
-      <Version>4.4.4-master-41290</Version>
+      <Version>4.4.5-master-2699519</Version>
```
joelverhagen: This dependency comes into NuGet.Services.AzureSearch via NuGet.Services.Validation.Common.Job. Are the versions the same? If not, how do we know there are no runtime issues?

src/NuGet.Services.AzureSearch/BlobContainerBuilder.cs
joelverhagen: This should reduce the number of changes greatly.

xavierdecoster: Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/531/commits/a43276a228128b0a265e1dcf63f1200fbbd8e79f

src/NuGet.Indexing/NuGet.Indexing.csproj
xavierdecoster: Looked a bit more closely into this, turns out `NuGet.Indexing` only needs `NuGet.Services.Entities`. Replaced the dependency here, and [aligned](https://github.com/NuGet/NuGet.Services.Metadata/pull/531/files#diff-bec7f637013541c5349ba50fe16cb67aR262) the `NuGet.Services.Entities` reference in `NuGet.Services.Metadata.Catalog` accordingly

tests/NuGet.Services.AzureSearch.Tests/Catalog2AzureSearch/Integration/InMemoryCloudBlobContainer.cs
```diff
@@ -30,7 +31,7 @@ public Task<bool> DeleteIfExistsAsync()
             throw new NotImplementedException();
         }
 
-        public Task<bool> ExistsAsync()
+        public Task<bool> ExistsAsync(BlobRequestOptions options = null, OperationContext operationContext = null)
```
xavierdecoster: This change is required as the `ExistsAsync` API on the `NuGetGallery.ICloudBlobContainer` interface changed.

src/NuGet.Services.AzureSearch/NuGet.Services.AzureSearch.csproj
```diff
@@ -215,6 +215,9 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
+    <PackageReference Include="NuGetGallery.Core">
+      <Version>4.4.5-master-2699519</Version>
```
xavierdecoster: Explicitly adding a reference here to the latest version of `NuGetGallery.Core` to force the transitive dependency upgrade of `NuGet.Services.Validation.Common.Job`'s dependency on it.

I think we need to look into why `NuGet.Services.Validation.Common.Job` brings in the `NuGetGallery.Core` dependency. AFAICT, the only thing used from that dependency is some `IDiagnosticsService` related types and our cloud blob APIs and wrappers. Perhaps we could simplify things by extracting those into their own nupkg, e.g. `NuGet.Services.BlobStorage` or something.

There's really no need to drag along all of the other transitive dependencies that `NuGetGallery.Core` brings along, and we could untangle our nupkgs inter-dependencies by splitting up those APIs a bit more.

image

joelverhagen: This was intentional. There was a period when the churn in the validation jobs was extremely high (3 people working in parallel) and we used the NuGet.Services.Validation.Common.Job project to consolidate a lot of the duplicate code between job initialization (e.g. DI).

The config, storage, and entities context DI wire-up is shared in the project, part of this a dependency on NuGetGallery.Core. I think a better way of looking at this project name is "NuGet.Jobs.Common++". Yes it has bit of validation specific code but I think now that the high churn in validation is decreased, we should consider merging NuGet.Services.Validation.Common.Job down into NuGet.Jobs.Common. With this cross repo world we are in, the fewer projects we have the easier it is to maintain things.

I guess what I was getting at is that we've running into a lot of pain (some of which only found at runtime) when we mix package versions and have a lot of little libraries.

Anyhow, as long as runtime works for all of our entry points and there are no weird restore/build warnings, this is fine too.

PR #533, opened 2019-05-30
Add root page for Azure Search Service and warmup config

joelverhagen:
1. Make the root page "/" return something (`{"Success":true}`)
1. Add a web.config setting to tell Azure Web Apps to call a specific endpoint for warming an instance

Given that gallery has two health probe endpoint implementations by design, I introduced a `SearchStatusOptions` to select exactly what dependencies we want to check for a health probe. This will make course changes easier for Azure Search service in the future.

### DIFF
--- a/src/NuGet.Services.SearchService/Controllers/SearchController.cs
+++ b/src/NuGet.Services.SearchService/Controllers/SearchController.cs
@@ -43,11 +43,27 @@ namespace NuGet.Services.SearchService.Controllers
 
         [HttpGet]
         [ResponseType(typeof(SearchStatusResponse))]
+        public async Task<HttpResponseMessage> IndexAsync(HttpRequestMessage request)
+        {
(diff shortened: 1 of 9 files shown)


src/NuGet.Services.SearchService/Controllers/SearchController.cs
```diff
@@ -41,13 +41,29 @@ public class SearchController : ApiController
...
         }
 
+        [HttpGet]
+        [ResponseType(typeof(SearchStatusResponse))]
+        public async Task<HttpResponseMessage> IndexAsync(HttpRequestMessage request)
+        {
+            var result = await GetStatusAsync(SearchStatusOptions.All);
+            var statusCode = result.Success ? HttpStatusCode.OK : HttpStatusCode.InternalServerError;
+
+            // Hide all information exception the success boolean. This is the root page so we can keep it simple.
```
joelverhagen: except

PR #535, opened 2019-06-02
[Azure Search] Don't allow KeyVault secret cache to expire

joelverhagen:
I was seeing the Azure Search Service hang after 24 hours. This is due to the fact that the KeyVault secret injector was running in a request context causing a deadlock.
https://github.com/NuGet/ServerCommon/blob/36b22e856d12c4d614ead56fe7cf41378c7077a2/src/NuGet.Services.Configuration/KeyVaultInjectingConfigurationProvider.cs#L48

In short, app service recycle will handle secret rotation as with gallery.

I have verified the problem locally by reducing the KeyVault cache duration to 1 minute. I have two regions (EA and SEA) deployed with this fix right now and two with the problem (USNC and USSC).

We avoid this in **gallery** by evaluating the secrets on the startup thread and caching them forever in a `Lazy<T>`. We don't use this exact type but the context is the same here: 
https://github.com/NuGet/NuGetGallery/blob/aa628079fd17c94848700ad36f454c021503b2fe/src/NuGetGallery/Configuration/ConfigurationService.cs#L41-L44

We avoid this in **legacy search** by evaluating on the startup thread and not blocking on the reload. We may also avoid this because we don't use ASP.NET MVC, but I'm not sure of this hypothetical. This is a fire-and-forget in legacy search where we get the new secrets:
https://github.com/NuGet/NuGet.Services.Metadata/blob/41c72dd96c46447b706d9a5c29a22cc082318d10/src/NuGet.Services.BasicSearch/Startup.cs#L234

We avoid this in **jobs** because there is not the same synchronization context restrictions.

I chose this approach instead of a `Lazy<T>` because I don't want to use `ConfigurationService` like gallery and wanted to use the Microsoft.Extensions.Configuration stuff.

### DIFF
--- a/src/NuGet.Services.SearchService/App_Start/WebApiConfig.cs
+++ b/src/NuGet.Services.SearchService/App_Start/WebApiConfig.cs
@@ -8,7 +8,6 @@ using System.Threading;
 using System.Threading.Tasks;
 using System.Web.Hosting;
 using System.Web.Http;
-using System.Web.Http.Dependencies;
 using Autofac;
 using Autofac.Extensions.DependencyInjection;
 using Autofac.Integration.WebApi;
@@ -33,7 +32,14 @@ namespace NuGet.Services.SearchService
     {
         private const string ControllerSuffix = "Controller";
         private const string ConfigurationSectionName = "SearchService";
-        private static readonly TimeSpan KeyVaultSecretCachingTimeout = TimeSpan.FromDays(1);
+
+        /// <summary>
+        /// We use <see cref="int.MaxValue"/> number of seconds for the caching timeout to essentially cache the
+        /// KeyVault secrets for the entire life of the process. This duration is over 68 years so... close enough. 
+        /// This is necessary because the KeyVault secret injector waits for an asynchronous operation to complete
+        /// in a synchronous context. This can cause deadlocks in ASP.NET when this occurs in request context.
+        /// </summary>
+        private static readonly TimeSpan KeyVaultSecretCachingTimeout = TimeSpan.FromSeconds(int.MaxValue);
 
         public static void Register(HttpConfiguration config)
         {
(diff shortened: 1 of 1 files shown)


skofman1:
> In short, app service recycle will handle secret rotation as with gallery.

How often does the recycle happen? What is it triggered by?

src/NuGet.Services.SearchService/App_Start/WebApiConfig.cs
```diff
@@ -33,7 +32,14 @@ public static class WebApiConfig
...
         private const string ConfigurationSectionName = "SearchService";
-        private static readonly TimeSpan KeyVaultSecretCachingTimeout = TimeSpan.FromDays(1);
+
+        /// <summary>
+        /// We use <see cref="int.MaxValue"/> number of seconds for the caching timeout to essentially cache the
+        /// KeyVault secrets for the entire life of the process. This duration is over 68 years so... close enough. 
+        /// This is necessary because the KeyVault secret injector waits for an asynchronous operation to complete
+        /// in a synchronous context. This can cause deadlocks in ASP.NET when this occurs in request context.
+        /// </summary>
+        private static readonly TimeSpan KeyVaultSecretCachingTimeout = TimeSpan.FromSeconds(int.MaxValue);
```
skofman1: If you don't intend to refresh configuration in the process and load it during startup only, why use CachingSecretReader at all? You can use SecretReader directly. I don't have a strong opinion. Just sayin..

joelverhagen: I haven't dug through the various implementations we have fully -- maybe I am missing something. I just looked through ServerCommon. The `KeyVaultReader` (couldn't find "SecretReader") doesn't cache at all so each config activation via DI would be hitting KeyVault if we used that.

I considered resolving each of the strings directly then putting the resolved settings object in the DI container as a singleton but this would require a fair bit of custom code or the construction of two DI service providers (one for the first resolution one to hold a singleton). I favored this approach I implemented because I was hoping we could fix our KeyVault adapter to Microsoft.Extensions.Configuration and get auto-secret rotation in all the places (just based off of an arbitrary cache lifetime). Then we could just reduce the cache duration here and be done with it.

Happy to take another approach I just don't see the simpler path.

joelverhagen:
> How often does the recycle happen? What is it triggered by?

I cases where the process didn't deadlock I observed 24 hours. Couldn't find docs. I presume it's an IIS app pool recycle of sorts.

I will continue to observe the PID duration after this fix is applied.

PR #539, opened 2019-06-04
Hotfix binding redirects for search services

xavierdecoster:
Cherry-picked the changes from `master` and resolved merge conflicts in separate branch.
Supersedes https://github.com/NuGet/NuGet.Services.Metadata/pull/538

PR #540, opened 2019-06-04
Fix unit test failure

xavierdecoster:
There seems to be a unit test setup issue causing build failures. This PR fixes them.


xavierdecoster:
> How did this get into `dev` if there is a unit test failure?

I am wondering the same thing! I don't know. Noticed the test failure as part of [this PR](https://github.com/NuGet/NuGet.Services.Metadata/pull/539). Then locally switched to the `dev` branch to verify where the failure occured (which branch), noticed it was in `dev`...

scottbommarito:
I explicitly remember adding this to the unit tests...maybe it got corrupted by the merge?

xavierdecoster:
@scottbommarito could be! I was surprised to see this, as I remember you adding it.

PR #542, opened 2019-06-05
Package Deprecation - Import deprecation information into catalog

xavierdecoster:


src/Catalog/Helpers/Db2CatalogProjection.cs
```diff
@@ -19,29 +22,58 @@ public Db2CatalogProjection(PackageContentUriBuilder packageContentUriBuilder)
...
+                hideLicenseReport ? null : dataReader[Db2CatalogProjectionColumnNames.LicenseReportUrl]?.ToString(),
+                deprecationInfo);
+        }
+
+        public FeedPackageDeprecationInfo ReadDeprecationInfoFromDataRecord(DbDataReader dataReader)
+        {
+            var deprecationReasons = new List<string>();
+            var deprecationStatus = (PackageDeprecationStatus)dataReader.ReadNullableInt32(Db2CatalogProjectionColumnNames.DeprecationStatus);
+
+            if (deprecationStatus.HasFlag(PackageDeprecationStatus.Legacy))
```
scottbommarito: Could we use a method on `Enum` for this so that the method does need to be updated for every new `PackageDeprecationStatus` flag?

[Based on this thread from StackOverflow:](https://stackoverflow.com/questions/4171140/iterate-over-values-in-flags-enum)

```
foreach (var deprecationStatusFlag in Enum.GetValues(typeof(PackageDeprecationStatus)))
{
    if (deprecationStatusFlag == PackageDeprecationStatus.NotDeprecated)
    {
        continue;
    }

    if (deprecationStatus.HasFlag(deprecationStatusFlag))
    {
        deprecationReasons.Add(deprecationStatusFlag.ToString());
    }
}
```

src/Catalog/Extensions/DbDataReaderExtensions.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace System.Data.Common
+{
+    public static class DbDataReaderExtensions
```
agr: Two methods here are almost identical, how about refactoring it to not have copy-pasted code?
I would suggest adding `private static T ReadColumn<T>(DbDataReader dr, string cn, Func<int, T> provideValue)` and make other use it supplying type-specific `provideValue` delegates.

src/Catalog/Helpers/GalleryDatabaseQueryService.cs
```diff
@@ -20,19 +20,6 @@ public class GalleryDatabaseQueryService : IGalleryDatabaseQueryService
         private const string PackageIdParameterName = "PackageId";
         private const string PackageVersionParameterName = "PackageVersion";
 
-        private static readonly string Db2CatalogSqlSubQuery = $@"PR.[Id],
```
scottbommarito: I expect that as part of the `MonitoringProcessor` changes for deprecation, we will need to revert this refactoring so that `GetPackageOrNull` can return deprecation information. I would rather that we don't refactor this and then just verify that `MonitoringProcessor` doesn't validate the deprecation information with this change.

src/Catalog/Helpers/FeedPackageDeprecationInfo.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace NuGet.Services.Metadata.Catalog.Helpers
+{
+    public sealed class FeedPackageDeprecationInfo
```
scottbommarito: Why don't we just reuse `PackageDeprecationItem`? Having this additional class doesn't really provide much IMO.

src/Catalog/Helpers/Db2CatalogProjection.cs
```diff
@@ -19,29 +22,58 @@ public Db2CatalogProjection(PackageContentUriBuilder packageContentUriBuilder)
...
+            if (deprecationStatus.HasFlag(PackageDeprecationStatus.Other))
+            {
+                deprecationReasons.Add(PackageDeprecationStatus.Other.ToString());
+            }
+
+            return new FeedPackageDeprecationInfo(
+                deprecationReasons,
+                dataReader.ReadNullableString(Db2CatalogProjectionColumnNames.DeprecationMessage),
+                dataReader.ReadNullableString(Db2CatalogProjectionColumnNames.AlternatePackageId),
+                dataReader.ReadNullableString(Db2CatalogProjectionColumnNames.AlternatePackageVersion));
```
scottbommarito: This is incorrect. We need to pass in the version range, not the version string directly.

This would be either
- `*` when version is null
- `[{version}, {version}]` when a version is specified.

src/Catalog/Helpers/GalleryDatabaseQueryService.cs
```diff
@@ -167,7 +154,25 @@ public async Task<FeedPackageDetails> GetPackageOrNull(string id, string version
...
+                        P.[NormalizedVersion],
+                        P.[Created],
+                        P.[LastEdited],
+                        P.[Published],
+                        P.[Listed],
+                        P.[HideLicenseReport],
+                        P.[LicenseNames],
+                        P.[LicenseReportUrl],
+                        PD.[Status] AS '{Db2CatalogProjectionColumnNames.DeprecationStatus}',
+                        APR.[Id] AS '{Db2CatalogProjectionColumnNames.AlternatePackageId}',
```
scottbommarito: When a package version is chosen instead of just the registration, `AlternatePackageRegistrationKey` will be null.
You need to fetch the package ID using `AlternatePackageKey`'s `PackageRegistrationKey`.

As a result, you need to pull in all the following things
`AlternatePackageRegistrationId`
`AlternatePackageId`
`AlternatePackageVersion`

src/Catalog/Helpers/Db2CatalogProjection.cs
```diff
@@ -19,29 +22,58 @@ public Db2CatalogProjection(PackageContentUriBuilder packageContentUriBuilder)
...
                 packageId,
                 normalizedPackageVersion,
-                hideLicenseReport ? null : dataRecord[Db2CatalogProjectionColumnNames.LicenseNames]?.ToString(),
-                hideLicenseReport ? null : dataRecord[Db2CatalogProjectionColumnNames.LicenseReportUrl]?.ToString());
+                hideLicenseReport ? null : dataReader[Db2CatalogProjectionColumnNames.LicenseNames]?.ToString(),
+                hideLicenseReport ? null : dataReader[Db2CatalogProjectionColumnNames.LicenseReportUrl]?.ToString(),
+                deprecationInfo);
+        }
+
+        public FeedPackageDeprecationInfo ReadDeprecationInfoFromDataRecord(DbDataReader dataReader)
```
scottbommarito: Any tests for this method and its changes?

src/Catalog/Extensions/DbDataReaderExtensions.cs
xavierdecoster: I managed to reduce code, but the tricky part is that the generic constraints for value types and reference types are contradictory. Managed to work around that though using the `dynamic` keyword :) Please have another look. I could duplicate code again by making a proper value type and reference type variant of the method.

src/Catalog/Helpers/Db2CatalogProjection.cs
xavierdecoster: Fixed!

src/Catalog/Helpers/GalleryDatabaseQueryService.cs
xavierdecoster: Reverted query duplication in https://github.com/NuGet/NuGet.Services.Metadata/pull/542/commits/e57f596536a1840dfa2a5191c85946f149b99062

src/Catalog/Helpers/FeedPackageDeprecationInfo.cs
xavierdecoster: Yep! Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/542/commits/b163c15e78dfd6b998ac885c806d36c6c3899c74

src/Catalog/Helpers/Db2CatalogProjection.cs
xavierdecoster: Added in https://github.com/NuGet/NuGet.Services.Metadata/pull/542/commits/8b9758d296cee6dc79a5f80535fab1f2aad67f7c

src/Catalog/Helpers/GalleryDatabaseQueryService.cs
xavierdecoster: Now using the `IsNull` SQL construct on the join in https://github.com/NuGet/NuGet.Services.Metadata/pull/542/commits/8b9758d296cee6dc79a5f80535fab1f2aad67f7c

If no alternate package version is defined, `AP.[PackageRegistrationKey]` will yield `null`, and the join will happen on the `PackageDeprecation.AlternatePackageRegistrationKey` instead.

src/Catalog/Helpers/Db2CatalogProjection.cs
xavierdecoster: Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/542/commits/8b9758d296cee6dc79a5f80535fab1f2aad67f7c#diff-3ef253b665c20ee4412e5ad2f110f4f6R89

src/Catalog/Helpers/CatalogWriterHelper.cs
```diff
@@ -0,0 +1,164 @@
...
+        /// Asynchronously writes package metadata to the catalog.
+        /// </summary>
+        /// <param name="packageCatalogItemCreator">A package catalog item creator.</param>
+        /// <param name="packages">Packages to download metadata for.</param>
+        /// <param name="storage">Storage.</param>
+        /// <param name="lastCreated">The catalog's last created datetime.</param>
+        /// <param name="lastEdited">The catalog's last edited datetime.</param>
+        /// <param name="lastDeleted">The catalog's last deleted datetime.</param>
+        /// <param name="maxDegreeOfParallelism">The maximum degree of parallelism for package processing.</param>
+        /// <param name="createdPackages"><c>true</c> to include created packages; otherwise, <c>false</c>.</param>
```
cristinamanum: Why not called includeCreatedPackages instead of createdPackages?

src/Catalog/Helpers/CatalogWriterHelper.cs
```diff
@@ -0,0 +1,164 @@
...
+        /// <see cref="DateTime}" /> that was processed.</returns>
+        public static async Task<DateTime> WritePackageDetailsToCatalogAsync(
+            IPackageCatalogItemCreator packageCatalogItemCreator,
+            SortedList<DateTime, IList<FeedPackageDetails>> packages,
+            IStorage storage,
+            DateTime lastCreated,
+            DateTime lastEdited,
+            DateTime lastDeleted,
+            int maxDegreeOfParallelism,
+            bool? createdPackages,
```
cristinamanum: What is the meaning of a null value?

src/Catalog/Helpers/CatalogWriterHelper.cs
```diff
@@ -0,0 +1,164 @@
...
+
+            await workItems.ForEachAsync(maxDegreeOfParallelism, async workItem =>
+            {
+                workItem.PackageCatalogItem = await packageCatalogItemCreator.CreateAsync(
+                    workItem.FeedPackageDetails,
+                    workItem.Timestamp,
+                    cancellationToken);
+            });
+
+            lastDate = packages.Last().Key;
```
cristinamanum: Maybe this line moved closer to line 94? I do not think that the packages are changed line 95 - line 107 code block.

src/Catalog/Extensions/DbDataReaderExtensions.cs
scottbommarito: Why does it need a `return` type of `dynamic` and not just `T`?

src/Catalog/Helpers/CatalogWriterHelper.cs
```diff
@@ -0,0 +1,164 @@
...
+            foreach (var workItem in workItems.Where(workItem => workItem.PackageCatalogItem != null))
+            {
+                writer.Add(workItem.PackageCatalogItem);
+
+                logger?.LogInformation("Add metadata from: {PackageDetailsContentUri}", workItem.FeedPackageDetails.ContentUri);
+            }
+
+            if (createdPackages.HasValue)
+            {
+                lastEdited = !createdPackages.Value ? lastDate : lastEdited;
```
cristinamanum: Should be createdPackages.Value instead of !createdPackages.Value?  There are not any checks for the order of lastEdited > or < lastDate . Is any logic needed for this?

src/Catalog/Helpers/Db2CatalogProjection.cs
```diff
@@ -19,29 +23,70 @@ public Db2CatalogProjection(PackageContentUriBuilder packageContentUriBuilder)
...
+                }
+            }
+
+            var alternatePackageVersion = dataReader.ReadStringOrNull(Db2CatalogProjectionColumnNames.AlternatePackageVersion);
+
+            return new PackageDeprecationItem(
+                deprecationReasons,
+                dataReader.ReadStringOrNull(Db2CatalogProjectionColumnNames.DeprecationMessage),
+                dataReader.ReadStringOrNull(Db2CatalogProjectionColumnNames.AlternatePackageId),
+                alternatePackageVersion == null ? "*" : $"[{alternatePackageVersion}, {alternatePackageVersion}]");
```
scottbommarito: If `AlternatePackageId` and `AlternatePackageVersion` are both null, it looks like you'll be passing `null` and `*` into this method, [which will cause `PackageDeprecationItem`'s constructor to throw](https://github.com/NuGet/NuGet.Services.Metadata/blob/master/src/Catalog/PackageDeprecationItem.cs#L41).

I'd also suggest making sure the unit tests cover:
null ID and null version
not null ID and null version
not null ID and not null version

src/Catalog/Helpers/CatalogWriterHelper.cs
```diff
@@ -0,0 +1,164 @@
...
+            var commitMetadata = PackageCatalog.CreateCommitMetadata(writer.RootUri, new CommitMetadata(lastCreated, lastEdited, lastDeleted));
+
+            await writer.Commit(commitMetadata, cancellationToken);
+
+            logger?.LogInformation("COMMIT metadata to catalog.");
+
+            return lastDate;
+        }
+
+        private static DateTime DetermineLastDate(DateTime lastCreated, DateTime lastEdited, bool? createdPackages)
```
cristinamanum: It does not matter which of the lastEdited lastCreated is newer? If there are certain expectations for the order should add logs if the expectations are not met?

src/Catalog/Helpers/FeedHelpers.cs
```diff
@@ -144,155 +141,11 @@ public static async Task<IList<FeedPackageDetails>> GetPackages(HttpClient clien
                     packageId,
                     packageNormalizedVersion,
                     licenseNames,
-                    licenseReportUrl));
+                    licenseReportUrl,
+                    deprecationInfo: null));
```
scottbommarito: Does anything still use this? We should clean up all the V2 feed code eventually--maybe in a separate PR?

src/Catalog/PackageCatalogItemCreator.cs
```diff
@@ -136,7 +144,8 @@ public sealed class PackageCatalogItemCreator : IPackageCatalogItemCreator
                             packageItem.PublishedDate,
                             licenseNames: null,
                             licenseReportUrl: null,
-                            packageHash: packageHash);
+                            packageHash: packageHash,
+                            deprecationItem: deprecationItem);
```
scottbommarito: Can we not just pass `packageItem.DeprecationInfo` here directly? Why do we need to clone the object?

src/Catalog/PackageCatalogItemCreator.cs
```diff
@@ -198,19 +207,28 @@ private async Task<PackageCatalogItem> GetPackageViaHttpAsync(FeedPackageDetails
...
+                        : null;
+
                     item = Utils.CreateCatalogItem(
                         packageItem.ContentUri.ToString(),
                         stream,
                         packageItem.CreatedDate,
                         packageItem.LastEditedDate,
-                        packageItem.PublishedDate);
+                        packageItem.PublishedDate,
+                        deprecationItem: deprecationItem);
```
scottbommarito: Same as above: can we not just pass `packageItem.DeprecationInfo` here directly? Why do we need to clone the object?

src/Catalog/Extensions/DbDataReaderExtensions.cs
xavierdecoster: Actually, that is possible now since I'm passing in the `nullValue` as an argument.
Updated https://github.com/NuGet/NuGet.Services.Metadata/pull/542/commits/8b7ab6b8a064942a58efd99a811fa889445dbb63

src/Catalog/Helpers/CatalogWriterHelper.cs
xavierdecoster: This is existing code which just moved to another type.

src/Catalog/Helpers/CatalogWriterHelper.cs
xavierdecoster: This is existing code which just moved to another type.

src/Catalog/Helpers/CatalogWriterHelper.cs
xavierdecoster: This is existing code which just moved to another type.

src/Catalog/Helpers/CatalogWriterHelper.cs
xavierdecoster: This is existing code which just moved to another type.

src/Catalog/Helpers/CatalogWriterHelper.cs
xavierdecoster: This is existing code which just moved to another type.

src/Catalog/Helpers/FeedHelpers.cs
xavierdecoster: Still used by `package2catalog` job.

scottbommarito: `package2catalog` will be broken by this work...we should probably remove it entirely

alternatively, we could update it to use the DB, but I don't think the job has been used in many years so it's safe to just delete

src/Catalog/PackageCatalogItemCreator.cs
xavierdecoster: No longer needed indeed. Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/542/commits/c9193c2f337d98a07658c107296d03b815243964

src/Catalog/PackageCatalogItemCreator.cs
xavierdecoster: No longer needed indeed. Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/542/commits/c9193c2f337d98a07658c107296d03b815243964

src/Catalog/Helpers/FeedHelpers.cs
xavierdecoster: @joelverhagen @skofman1 anyone remember the purpose of `package2catalog`? Should we keep maintaining it? Or can I go ahead and remove it completely?

And should it be part of this PR or tracked by separate issue? :)

src/Catalog/Helpers/Db2CatalogProjection.cs
xavierdecoster: Good catch! This was hidden by a faulty test setup.
Fixed in https://github.com/NuGet/NuGet.Services.Metadata/pull/542/commits/592665ecec20b491925cc20e76805c167608fc3b

src/Catalog/Helpers/FeedHelpers.cs
scottbommarito: Talked to @joelverhagen offline (@skofman1 is on vacation this week and next). We came to the consensus that it should be removed, along with all the code that is only remaining because of it. I would do it in a separate PR though.

xavierdecoster: Tracking issue: https://github.com/nuget/engineering/issues/2504

PR #543, opened 2019-06-05
Test warning fixes

agr:
VS shows dozens of xUnit warnings now. This addresses most of them.


tests/CatalogTests/RetryWithExponentialBackoffTests.cs
```diff
@@ -74,6 +74,8 @@ public static IEnumerable<object[]> ReturnsFalseIfResponseStatusBelow500OrWhitel
                 }
             }
 
+            [Theory]
+            [MemberData(nameof(ReturnsFalseIfResponseStatusBelow500OrWhitelisted_Data))]
```
agr: @scottbommarito, check here and below that `MemberData` is the expected one.

scottbommarito: yup

PR #544, opened 2019-06-06
[Azure Search] Make scoring profile configurable

loic-sharma:
This allows us to change the scoring profile through configs instead of code changes. I verified this change by creating indexes using `db2azuresearch` and `catalog2azuresearch` on my own Azure Search indexes.

Depends on [NuGetDeployment#946](https://nuget.visualstudio.com/NuGetMicrosoft/_git/NuGetDeployment/pullrequest/946?_a=overview)


src/NuGet.Services.AzureSearch/AzureSearchScoringConfiguration.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// <summary>
+        /// Controls the log base of <see cref="SearchDocument.Full.LogOfDownloadCount"/>.
+        /// Decreasing the base increases the range of values, thereby increasing the boost
+        /// for packages with high download counts.
+        /// </summary>
+        public double DownloadCountLogBase { get; set; }
+
+        /// <summary>
+        /// The <see cref="BaseMetadataDocument.PackageId"/> field weight.
+        /// This field is used to match a package's id exactly and should have a high weight.
```
joelverhagen: nit: `ID`

src/NuGet.Services.AzureSearch/ScoringProfiles/DefaultScoringProfile.cs
```diff
@@ -12,44 +12,47 @@ public static class DefaultScoringProfile
...
-                        boostingRangeStart: 0,
-                        boostingRangeEnd: Math.Log(999_999_999_999, 2),
-                        shouldBoostBeyondRangeByConstant: true),
-                    interpolation: ScoringFunctionInterpolation.Linear),
+                    Weights = new Dictionary<string, double>
+                    {
+                        // Exact match of the package id should be boosted the highest,
+                        // followed by a tokenized match on the package id.
+                        { IndexFields.PackageId, config.PackageIdFieldWeight },
+                        { IndexFields.TokenizedPackageId, config.TokenizedPackageIdFieldWeight },
```
joelverhagen: Thoughts on putting this entire dictionary in config? Not sure if you are anticipating other fields getting weights.

tests/NuGet.Services.AzureSearch.Tests/IndexBuilderFacts.cs
```diff
@@ -148,7 +151,7 @@ public BaseFacts(ITestOutputHelper output)
 
                 _options
                     .Setup(x => x.Value)
-                    .Returns(() => _config);
+                    .Returns(_config);
```
joelverhagen: nit I typically us `() => _config` as a pattern so that if a test wants to replace an instance it's easier.

src/NuGet.Services.AzureSearch/ScoringProfiles/DefaultScoringProfile.cs
loic-sharma: That sounds good to me, it'll give us better flexibility

joelverhagen: Only drawback is that now the field names are set in two repos. Not sure of the behavior if the field does not exist.

tests/NuGet.Services.AzureSearch.Tests/IndexBuilderFacts.cs
joelverhagen: (It's `readonly` right now but doesn't really have to if it makes test setup easier)

loic-sharma: Ah interesting. That makes sense, I'll revert this

PR #546, opened 2019-06-07
Deprecation metadata should only be present in SemVer 2.0.0 hive

scottbommarito:
https://github.com/nuget/nugetgallery/issues/7251

We only want to expose deprecation metadata in the SemVer 2.0.0 hive because we don't want to have to make any decision about how to handle alternate packages that are SemVer 2.0.0 in the SemVer 1.0.0 hive and SemVer 1.0.0 clients won't even read deprecation metadata anyway.

Most of the code changes involve passing down a `RegistrationMakerCatalogItem.PostProcessGraph` from `RegistrationCollector` to `RegistrationMakerCatalogItem`


tests/CatalogTests/Registration/RegistrationMakerCatalogItemTests.cs
```diff
@@ -265,67 +272,77 @@ public static IEnumerable<object[]> CreatePageContent_SetsDeprecationInformation
...
+                    Assert.Empty(deprecationPredicateTriples);
+                    return;
+                }
+
+                var deprecationObjectNode = deprecationPredicateTriples
                     .Single()
                     .Object;
 
-                var deprecationTriples = _graph.GetTriplesWithSubject(deprecationObjectNode);
+                var deprecationTriples = content.GetTriplesWithSubject(deprecationObjectNode);
```
scottbommarito: This test was silently broken before: we were testing `_graph`, the input to the `RegistrationMakerCatalogItem`, instead of `content`, which is the `RegistrationMakerCatalogItem`'s output.

src/Catalog/Registration/RegistrationMakerCatalogItem.cs
```diff
@@ -108,9 +119,7 @@ private DateTime GetPublishedDate()
 
                 if (pubTriple != null)
                 {
-                    ILiteralNode node = pubTriple.Object as ILiteralNode;
-
-                    if (node != null)
+                    if (pubTriple.Object is ILiteralNode node)
```
scottbommarito: I fixed a couple VS info messages while making this change--I have no problem reverting them if there's push-back.

src/Catalog/Registration/RegistrationCollector.cs
```diff
@@ -24,7 +25,8 @@ public class RegistrationCollector : SortingGraphCollector
 
         private readonly StorageFactory _legacyStorageFactory;
         private readonly StorageFactory _semVer2StorageFactory;
-        private readonly ShouldIncludeRegistrationPackage _shouldIncludeSemVer2;
+        private readonly ShouldIncludeRegistrationPackage _shouldIncludeSemVer2ForLegacyStorageFactory;
+        private readonly RegistrationMakerCatalogItem.PostProcessGraph _postProcessGraphForLegacyStorageFactory;
```
joelverhagen: Why nest the delegate inside a class?

scottbommarito: I wanted to make it more obvious that it's strictly used by `RegistrationMakerCatalogItem`.
For example, where `ShouldIncludeRegistrationPackage` is used is much harder to find.

PR #547, opened 2019-06-07
Add "unknown" package status to monitoring jobs

scottbommarito:
https://github.com/NuGet/Engineering/issues/2038

This PR allows validators to distinguish between
1. A validation that does not apply to this package
2. A validation that needs to run against a future catalog entry of this package

Previously, if we processed a package and determined we didn't have the latest catalog entry for it, we would mark it as `PackageState.Valid`. As a result, there was no way for us to find these packages without doing a very expensive full-scan of the `valid` container.

With this change, packages marked `PackageState.Unknown` will now be placed in their own container, where we can monitor whether or not they've been reprocessed.

Our monitoring specifically looks for packages marked `PackageState.Invalid`, so this change won't cause any existing alerts to go off. We will need to update the monitoring to track `PackageState.Unknown` as well after this change.


src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageState.cs
```diff
@@ -18,6 +18,12 @@ public enum PackageState
...
         /// Its metadata is missing or in an unexpected state.
         /// </summary>
-        Invalid
+        Invalid,
+
+        /// <summary>
+        /// The package's state could not be determined.
+        /// Its metadata is in an unknown state.
+        /// </summary>
+        Unknown
```
joelverhagen: nit: trailing comma to reduce diff noise

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/Catalog/PackageHasSignatureValidator.cs
```diff
@@ -26,13 +26,8 @@ public sealed class PackageHasSignatureValidator : Validator<CatalogEndpoint>
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
         }
 
-        protected override async Task<bool> ShouldRunAsync(ValidationContext context)
+        protected override async Task<ShouldRunTestResult> ShouldRunAsync(ValidationContext context)
         {
-            if (!ShouldRunValidator(context))
```
scottbommarito: note: i will need to add this back

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/Validator.cs
```diff
@@ -87,10 +89,14 @@ protected virtual async Task<bool> ShouldRunAsync(ValidationContext context)
...
+            {
+                // If the timestamp metadata in the catalog is EQUAL to that of the feed, we are looking at the latest catalog entry that corresponds with this package, so run the test.
+                return ShouldRunTestResult.Yes;
+            }
 
-            // If the timestamp metadata in the catalog is LESS than that of the feed, we must not be looking at the latest entry that corresponds with this package, so skip the test for now.
-            // If the timestamp metadata in the catalog is EQUAL to that of the feed, we are looking at the latest catalog entry that corresponds with this package, so run the test.
-            return timestampCatalog.Last == timestampV2.Last;
+            // If the timestamp metadata in the catalog is LESS than that of the feed, we must not be looking at the latest entry that corresponds with this package, so we must attempt this test again later with more information.
+            return ShouldRunTestResult.Unknown;
```
loic-sharma: Instead of `ShouldRunTestResult.Unknown`, what do you think of `ShouldRunTestResult.RetryLater`? I think it makes the intent clearer

scottbommarito: I was actually thinking about this but forgot to update the PR. I like this much better.

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Result/TestResult.cs
```diff
@@ -12,13 +12,20 @@ public enum TestResult
...
         Fail,
+
+        /// <summary>
+        /// The test was skipped (<see cref="Validator.ShouldRunAsync(ValidationContext)"/> returned <see cref="ShouldRunTestResult.No"/>).
+        /// </summary>
+        Skip,
+
         /// <summary>
-        /// The test was skipped (<see cref="Validator.ShouldRunAsync(ValidationContext)"/> threw or returned false).
+        /// The result of the test could not be determined (<see cref="Validator.ShouldRunAsync(ValidationContext)"/> returned <see cref="ShouldRunTestResult.Unknown"/>)
```
loic-sharma: What do you think of a name like `Pending` to correspond to `ShouldRunTestResult.RetryLater`? If so, I would change the comment to something like, `The result of the test has not been determined`

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Result/ShouldRunTestResult.cs
```diff
@@ -0,0 +1,48 @@
...
+        Yes,
+
+        /// <summary>
+        /// The test can be safely skipped.
+        /// </summary>
+        No,
+
+        /// <summary>
+        /// It could not be determined if the test should be run.
+        /// The test must be attempted again when more information is available.
```
loic-sharma: I would mention that the validation will be done by a future leaf and not the current leaf.

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Result/TestResult.cs
loic-sharma: I would also mention that this will need a new leaf to move to a `Pass`/`Fail` final state

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Result/ShouldRunTestResult.cs
scottbommarito: Added your feedback to all the docs.

PR #549, opened 2019-06-08
Add ETag support to MonitoringProcessor

scottbommarito:
https://github.com/NuGet/NuGetGallery/issues/7254

This is a proposed solution for adding ETag support to the `NuGet.Services.Metadata.Persistence.Storage` APIs.

Rather than changing the `Save` and `Load` APIs themselves, I've added two additional `StorageContent` types.
1. `StringStorageContentWithAccessCondition` - for specifying `AccessCondition`s for a `Save`
2. `StringStorageContentWithETag` - for retrieving the `ETag` of a `Load`

Unfortunately, I did have to make changes to the `Delete` API. I added an optional ` DeleteRequestOptions` parameter. Callers can specify a `DeleteRequestOptionsWithAccessCondition` to specify an `AccessCondition` for a `Delete` operation on an `AzureStorage` class.

The benefit of this approach is that it's fairly light and requires mostly only changes to the `AzureStorage` implementation of `Storage`. Additionally, if a caller doesn't care about `ETag`s or `AccessCondition`s, these things are completely abstracted away from them.

After doing this, I was able to easily add this support to `PackageMonitoringStatusService` to prevent overwriting of status blobs.


src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -226,7 +226,15 @@ protected override async Task OnSaveAsync(Uri resourceUri, StorageContent conten
...
                     destinationStream.Seek(0, SeekOrigin.Begin);
 
-                    await blob.UploadFromStreamAsync(destinationStream, cancellationToken);
+                    var accessCondition = (content as StringStorageContentWithAccessCondition)?.AccessCondition 
+                        ?? AccessCondition.GenerateEmptyCondition();
+
+                    await blob.UploadFromStreamAsync(
+                        destinationStream, 
+                        accessCondition, 
+                        new BlobRequestOptions(),
```
loic-sharma: I think these can be null instead of empty

scottbommarito: done

src/Ng/Jobs/MonitoringProcessorJob.cs
```diff
@@ -172,14 +174,33 @@ private async Task ProcessPackagesAsync(CancellationToken token)
...
                 }
             }
 
             // If we failed to run validations and failed to save the failed validation, we cannot remove the message from the queue.
             if (messageWasProcessed)
             {
-                await _queue.RemoveAsync(queueMessage, token);
+                try
+                {
+                    await _queue.RemoveAsync(queueMessage, token);
```
scottbommarito: I observed locally that occasionally, with many instances running, an instance would take longer than the message invisibility duration to process a message, which lead to additional instances processing it at the same time and failing because when deleting because "message not found".

In general, there is no case for alarm if this call fails because either:
- the message was already processed, so all the work to do for it is finished anyway
- the message is still there and will be attempted again later
So I wrapped this call in a `try-catch` so the process will not need to restart in this case.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -226,7 +226,15 @@ protected override async Task OnSaveAsync(Uri resourceUri, StorageContent conten
...
                     destinationStream.Seek(0, SeekOrigin.Begin);
 
-                    await blob.UploadFromStreamAsync(destinationStream, cancellationToken);
+                    var accessCondition = (content as StringStorageContentWithAccessCondition)?.AccessCondition 
+                        ?? AccessCondition.GenerateEmptyCondition();
+
+                    await blob.UploadFromStreamAsync(
+                        destinationStream, 
+                        accessCondition, 
+                        null,
```
joelverhagen: >l [](start = 26, length = 1)

name null params

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -339,13 +347,15 @@ protected override async Task<StorageContent> OnLoadAsync(Uri resourceUri, Cance
             }
         }
 
-        protected override async Task OnDeleteAsync(Uri resourceUri, CancellationToken cancellationToken)
+        protected override async Task OnDeleteAsync(Uri resourceUri, DeleteRequestOptions deleteRequestOptions, CancellationToken cancellationToken)
         {
             string name = GetName(resourceUri);
 
+            var accessCondition = (deleteRequestOptions as DeleteRequestOptionsWithAccessCondition)?.AccessCondition;
```
joelverhagen: >i [](start = 26, length = 1)

default to empty condition?

scottbommarito: Is there a difference between null and `GenerateEmptyCondition`?
I wanted to avoid affecting the existing code path as much as possible.

joelverhagen: Check in ILSpy. I believe not. Just saying this is inconsistent with the `??` in your changeset above.

src/Catalog/Persistence/StringStorageContentWithETag.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Services.Metadata.Catalog.Persistence
+{
+    public class StringStorageContentWithETag : StringStorageContent
```
joelverhagen: >StringStorageContentWithETag [](start = 17, length = 28)

This is sort of "response" content. XML docs plz

src/Catalog/Persistence/StringStorageContentWithAccessCondition.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.WindowsAzure.Storage;
+
+namespace NuGet.Services.Metadata.Catalog.Persistence
+{
+    public class StringStorageContentWithAccessCondition : StringStorageContent
```
joelverhagen: >StringStorageContentWithAccessCondition [](start = 17, length = 39)

This is sort of "request" content. XML docs plz

src/Catalog/Persistence/AzureStorage.cs
scottbommarito: done

src/Catalog/Persistence/AzureStorage.cs
scottbommarito: I'm going to use `null` to be consistent with the existing behavior when an ETag is not supplied. The `ClockBlockBlob` docs say "If null, no condition is used.".

src/Catalog/Persistence/StringStorageContentWithAccessCondition.cs
scottbommarito: Added docs to `StringStorageContentWithAccessCondition`, `StringStorageContentWithETag`, and `DeleteRequestOptions`.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -351,7 +350,7 @@ protected override async Task OnDeleteAsync(Uri resourceUri, DeleteRequestOption
         {
             string name = GetName(resourceUri);
 
-            var accessCondition = (deleteRequestOptions as DeleteRequestOptionsWithAccessCondition)?.AccessCondition;
+            var accessCondition = (deleteRequestOptions as DeleteRequestOptionsWithAccessCondition)?.AccessCondition ?? null;
```
joelverhagen: wait isn't `?? null` kind of a no-op?

loic-sharma: Same comment for this line: https://github.com/NuGet/NuGet.Services.Metadata/pull/549/files#diff-7d744c5a2a0334b0e4d8c4c000aa6469R229

src/Ng/Jobs/MonitoringProcessorJob.cs
```diff
@@ -172,14 +174,33 @@ private async Task ProcessPackagesAsync(CancellationToken token)
...
+                }
+                catch (StorageException storageException)
+                {
+                    if (storageException.RequestInformation.ExtendedErrorInformation.ErrorCode == QueueErrorCodeStrings.MessageNotFound)
+                    {
+                        Logger.LogWarning(
+                            "Queue message for {PackageId} {PackageVersion} no longer exists. Message was likely handled by another instance of the job.",
```
loic-sharma: I would consider including the exception just in case we want to look deeper into this case in the future.

src/NuGet.Services.Metadata.Catalog.Monitoring/Status/PackageMonitoringStatusService.cs
```diff
@@ -86,17 +87,27 @@ public async Task<PackageMonitoringStatus> GetAsync(FeedPackageIdentity package,
...
             {
-                return statusesWithTimeStamps.OrderByDescending(s => s.ValidationResult.CatalogEntries.Max(c => c.CommitTimeStamp)).First();
+                orderedStatuses = statusesWithTimeStamps.OrderByDescending(s => s.ValidationResult.CatalogEntries.Max(c => c.CommitTimeStamp));
             }
             else
             {
                 // No statuses have timestamps (they all failed to process).
                 // Because they are all in a bad state, choose an arbitrary one.
-                return statuses.First();
+                orderedStatuses = statuses;
```
loic-sharma: Consider preferring invalid statuses over valid and pending statuses, and consider preferring pending statuses over valid statuses.

src/Ng/Jobs/MonitoringProcessorJob.cs
scottbommarito: added

src/Catalog/Persistence/AzureStorage.cs
scottbommarito: lmao good catch

PR #552, opened 2019-06-12
Add telemetry to Azure Search jobs and service

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/6664

I think the usefulness of unit tests on telemetry is less than the cost of implementing them. This is perhaps not true if we alert on the specific telemetry, but we have no plans at this time to alert on these. When we add an alert and have any concerns about telemetry reliability, we can add the tests at that point.

There is an "integration test" suite which uses a concrete implementation of `AzureSearchTelemetryService` so we have coverage there for the simplest of cases.

See unit tests for `ILogger` calls.

### DIFF
--- a/src/NuGet.Services.AzureSearch/DatabaseOwnerFetcher.cs
+++ b/src/NuGet.Services.AzureSearch/DatabaseOwnerFetcher.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Data.Entity;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
@@ -18,6 +19,7 @@ namespace NuGet.Services.AzureSearch
 
         private readonly ISqlConnectionFactory<GalleryDbConfiguration> _connectionFactory;
         private readonly IEntitiesContextFactory _entitiesContextFactory;
+        private readonly IAzureSearchTelemetryService _telemetryService;
         private readonly ILogger<DatabaseOwnerFetcher> _logger;
 
         private const string Sql = @"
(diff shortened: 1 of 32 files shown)


src/NuGet.Services.AzureSearch/DatabaseOwnerFetcher.cs
```diff
@@ -66,13 +71,17 @@ public async Task<string[]> GetOwnersOrEmptyAsync(string id)
                 var sortedOwners = owners
                     .OrderBy(o => o, StringComparer.OrdinalIgnoreCase)
                     .ToArray();
+
+                stopwatch.Stop();
+                _telemetryService.TrackGetOwnersForPackageId(sortedOwners.Length, stopwatch.Elapsed);
```
loic-sharma: You can use a `TrackDuration` here if you use `TrackDuration<TProperties>`. Example: https://github.com/NuGet/ServerCommon/pull/168

I don't feel strongly about this, your current code is succinct enough and doesn't require a new class to store the `sortedOwners`

loic-sharma: Same comment for `GetPackageIdToOwnersAsync`

joelverhagen: If there were mutable anonymous types this would be slam dunk. I don't think it's worth coming up with private classes for each of these cases.

joelverhagen: I'll try named tuple and see if it feels better

joelverhagen: Spoke offline. Decided against it.

PR #553, opened 2019-06-12
Delete V3PerPackage since it has served its purpose and is leading to unnecessary maintenance

joelverhagen:
Address https://github.com/NuGet/Engineering/issues/2501

I will link to the specific Git commit in the issue so that we can resurrect the code if necessary.


joelverhagen:
> What was V3PerPackage project? Good riddance :D

I wrote it to run all packages through V3. We did this before repo signing since we thought some packages would block the pipeline.

PR #556, opened 2019-06-13
Catalog2Lucene updates for embedded icon support

agr:
Addresses https://github.com/NuGet/NuGetGallery/issues/7067.

Added 2 new arguments that specify Flat container base URL and container name. Icon URL generation code was updated to check the presence of icon file metadata.


src/NuGet.Indexing/Extraction/CatalogPackageMetadataExtraction.cs
```diff
@@ -64,12 +83,31 @@ private class Extractor
...
+                    var packageVersion = JTokenToString(_catalog[MetadataConstants.NormalizedVersionPropertyName]);
+
+                    var pathProvider = new FlatContainerPackagePathProvider(flatContainerContainerName);
+                    var path = pathProvider.GetIconPath(packageId, packageVersion);
+                    var flatContainerIconUrl = new Uri(flatContainerBaseAddress, path);
+                    _metadata[MetadataConstants.IconUrlPropertyName] = flatContainerIconUrl.AbsoluteUri;
+                    return;
+                }
+
+                AddString(MetadataConstants.IconUrlPropertyName);
```
joelverhagen: this would be clearer in an else

src/Ng/Jobs/Catalog2LuceneJob.cs
```diff
@@ -117,6 +136,8 @@ protected override async Task RunInternalAsync(CancellationToken cancellationTok
                     commitTimeout: _commitTimeout,
                     baseAddress: _catalogBaseAddress,
                     galleryBaseAddress: _galleryBaseAddress == null ? null : new Uri(_galleryBaseAddress),
+                    flatContainerBaseAddress: _flatContainerBaseAddress,
+                    flatContainerConatinerName: _flatContainerContainerName,
```
joelverhagen: >flatContainerConatinerName [](start = 20, length = 26)

`Container`

src/NuGet.Indexing/Extraction/CatalogPackageMetadataExtraction.cs
```diff
@@ -64,12 +83,31 @@ private class Extractor
...
 
+            private void AddIconUrl(Uri flatContainerBaseAddress, string flatContainerContainerName)
+            {
+                var iconFile = JTokenToString(_catalog[MetadataConstants.IconFilePropertyName]);
+                if (!string.IsNullOrWhiteSpace(iconFile))
+                {
+                    var packageId = JTokenToString(_catalog[MetadataConstants.IdPropertyName]);
+                    var packageVersion = JTokenToString(_catalog[MetadataConstants.NormalizedVersionPropertyName]);
+
+                    var pathProvider = new FlatContainerPackagePathProvider(flatContainerContainerName);
```
joelverhagen: >FlatContainerPackagePathProvider [](start = 43, length = 32)

Can't we just pass the provider down instead of two params?

agr: The choice was between:
* `Uri flatContainerBaseAddress, string flatContainerContainerName`;
* `Uri flatContainerBaseAddress, IPackagePathProvider pathProvider`;
I.e. we can't get rid of `flatContainerBaseAddress` (see the few lines below, path provider does not provide full URL). Passing `IPackagePathProvider` provides a bit more freedom than I want in this method: it only supports flat container paths, so path provider must be `FlatContainerPackagePathProvider`.

joelverhagen: Sounds good!

PR #558, opened 2019-06-14
Azure search functional test configuration support

shishirx34:
These scripts are used by the VSTS Azure search functional tests build to run the tests with the configured environment.

I have separated the scripts for running the old service functional tests and the new azure search functional tests for easy editing and removal once we have migrated over to the new search service.

See the VSTS config PR for the new changes.

I am going to integrate this build with the EV2(WIP)


tests/Scripts/Configure-AzureSearchFunctionalTests.ps1
```diff
@@ -0,0 +1,8 @@
+[CmdletBinding(DefaultParameterSetName='RegularBuild')]
+param (
+    [string]$Slot
+)
+
+Write-Host "Setting environment variable for the slot: " $Slot
+$env:Slot = $Slot
+Write-Host "##vso[task.setvariable variable=Slot;]$Slot"
```
shishirx34: I am thinking of removing this config script all together and adding the "Slot" flag to the test configuration json directly through "Import-Configuration.ps1", similar to that of the Gallery functional tests... thoughts here?

tests/NuGet.Services.AzureSearch.FunctionalTests/Config/Prod.json
```diff
@@ -1,12 +1,14 @@
 ﻿{
```
joelverhagen: Should we even have these configs here? IMO these should be private.

shishirx34: no idea why these were added in the first place, @loic-sharma - fine to get rid of all these config files and simply have one sample "config.json"/"dev.json" which can define the config parameters for easy local run/debug?

shishirx34: deleting int and prod.json configs. Keeping the Dev.json for local debugging since its used as the default configuration name.

tests/Scripts/Configure-AzureSearchFunctionalTests.ps1
joelverhagen: Yeah, I think we should our patterns. Gallery's method seems good enough.

tests/NuGet.Services.AzureSearch.FunctionalTests/Config/Prod.json
loic-sharma: Yes, these are fine to delete.

tests/NuGet.Services.AzureSearch.FunctionalTests/Support/RelevancyFactAttribute.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Xunit;
+
+namespace NuGet.Services.AzureSearch.FunctionalTests
+{
+    public class RelevancyFactAttribute : FactAttribute
```
agr: Keep in mind that if you'd want to add another variant of it descending from `TheoryAttribute`, you'd need to specify test discoverer explicitly (like we do in [NuGet.Jobs](https://github.com/NuGet/NuGet.Jobs/blob/a48031f29aec0e3afd5efc7db425649869567029/tests/TestUtil/AdminOnlyTheoryAttribute.cs#L15)), otherwise it would fail to run the tests properly.
Direct descendants of `FactAttribute` does not seem to suffer from it, but I would still specify it here, too, for consistency.

agr: Please, add documentation.

tests/BasicSearchTests.FunctionalTests.Core/EnvironmentSettings.cs
```diff
@@ -61,6 +62,19 @@ public static string ConfigurationName
             }
         }
 
+        public static string Slot
+        {
+            get
+            {
+                if (string.IsNullOrEmpty(_slot))
```
agr: This could have been an instance of `Lazy<string>`, then you'd need not to write the code to check if it was initialized.

shishirx34:
@joelverhagen - ready with the changes. tested the configs against staging and prod for dev and different regions, looks good.

tests/BasicSearchTests.FunctionalTests.Core/BaseFunctionalTests.cs
```diff
@@ -21,6 +21,7 @@ public BaseFunctionalTests()
         public BaseFunctionalTests(string baseUrl)
         {
             // Arrange
+            Console.WriteLine($"Running tests against: {baseUrl}");
```
joelverhagen: If we want tracing we should use `ITestOutputHelper`

tests/NuGet.Services.AzureSearch.FunctionalTests/Support/AzureSearchConfiguration.cs
```diff
@@ -0,0 +1,109 @@
...
+using System.Threading.Tasks;
+using BasicSearchTests.FunctionalTests.Core;
+using Newtonsoft.Json;
+
+namespace NuGet.Services.AzureSearch.FunctionalTests
+{
+    public class AzureSearchConfiguration
+    {
+        private static AzureSearchConfiguration _configuration = null;
+        private static SemaphoreSlim _semaphore = new SemaphoreSlim(1);
```
joelverhagen: `readonly`

PR #559, opened 2019-06-15
Remove redundant PackageHasSignatureValidator

scottbommarito:
This validation is done more thoroughly by `PackageIsRepositorySignedValidator`


scottbommarito:
Discussed with @dtivel.

We should validate the content of the catalog--even if we can tell if the package is signed from the flat-container, we can't guarantee the catalog contains that information correctly as well.

PR #560, opened 2019-06-15
Run validations anyway with null catalog entries

scottbommarito:
Fix for https://github.com/NuGet/NuGetGallery/issues/7260

The catalog entries are only used to determine whether or not validations should run and are not required to run validations. If the queue message does not have a catalog entry, we can just run validations anyway and then throw if the state is inconsistent.


src/Ng/Jobs/MonitoringProcessorJob.cs
```diff
@@ -227,37 +219,6 @@ private async Task ProcessPackagesAsync(CancellationToken token)
             await _statusService.UpdateAsync(status, token);
         }
 
-        private async Task<IEnumerable<CatalogIndexEntry>> FetchCatalogIndexEntriesFromRegistrationAsync(
```
scottbommarito: Fetching the catalog entry from registration is a decent mitigation to when we don't have the latest catalog entry but it could lead to correctness issues.

For example, if the package was processed by flat-container but skipped by registration, fetching the entries from registration could lead to us getting the wrong catalog entry and then skipping running validations.

Given we now no longer require catalog entries to run validations, I felt that removing this mitigation was better than leaving it in.

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/ValidationContext.cs
```diff
@@ -60,11 +60,6 @@ public class ValidationContext
             CancellationToken token,
             ILogger<ValidationContext> logger)
         {
-            if (entries == null)
```
joelverhagen: We should leave a comment here about what cases entries will be non and non-null

scottbommarito: Added to the docs of `ValidationContext.Entries` itself.

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/Validator.cs
```diff
@@ -65,8 +65,18 @@ public async Task<ValidationResult> ValidateAsync(ValidationContext context)
         /// <summary>
         /// Checks that the current batch of catalog entries contains the entry that was created from the current state of the database.
         /// </summary>
+        /// <remarks>
+        /// Our validations depend on the fact that the database and V3 are expected to have the same version of a package.
+        /// If the catalog entry we're running validations on, which is supposed to represent the current state of V3, is less recent than the database, then we shouldn't run validations.
+        /// </remarks>
         protected virtual async Task<bool> ShouldRunAsync(ValidationContext context)
         {
+            if (context.Entries == null)
```
scottbommarito: Looked into adding unit tests around this. Apparently there are none currently and adding some is not trivial. I would like to punt this for now.

PR #561, opened 2019-06-15
Support flat container icon URL in Azure Search

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/7261

Another PR will happen when @agr finished the DB schema for icons. For now, only Catalog2AzureSearch will populate icon URL for packages with icon files.

Similar to PR https://github.com/NuGet/NuGet.Services.Metadata/pull/556.


src/NuGet.Services.AzureSearch/DocumentUtilities.cs
```diff
@@ -116,13 +117,19 @@ public static string GetSearchFilterString(SearchFilters searchFilters)
...
+            // https://github.com/NuGet/NuGetGallery/issues/7261
+            // https://github.com/nuget/nugetgallery/issues/7064
         }
 
         public static void PopulateMetadata(
             IBaseMetadataDocument document,
             string normalizedVersion,
             PackageDetailsCatalogLeaf leaf,
-            Uri galleryBaseUrl)
+            Uri galleryBaseUrl,
```
loic-sharma: Consider making this class non-static and passing these configs through constructor. You can make `FlatContainerPackagePathProvider` a field that's created by dependency injection too.

joelverhagen: Done. Left `FlatContainerPackagePathProvider` as a non-field to respect the option's reload semantics.

tests/NuGet.Services.AzureSearch.Tests/HijackDocumentBuilderFacts.cs
```diff
@@ -395,12 +395,24 @@ public void SetsLicenseUrlToGalleryWhenPackageHasLicenseFile()
 
                 Assert.Equal(Data.GalleryLicenseUrl, document.LicenseUrl);
             }
+
+            [Fact]
+            public void SetsIconUrlToFlatContainerWhenPackageHasIconFile()
```
agr: This seem to only test the situation when we have both iconUrl and iconFile specified. How about other combinations?

joelverhagen: just `iconUrl` is already tested. I will add one for just `iconFile`

PR #562, opened 2019-06-15
 Parallelize the catalog leaf download differently than batches

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/7265

USSC and SEA have the fix. USNC and EA do not.

instanceName | name | value (seconds)
-- | -- | --
Catalog2AzureSearch-ussc-a | ProcessBatchSeconds | 53.10917
Catalog2AzureSearch-usnc-a | ProcessBatchSeconds | 341.2235
Catalog2AzureSearch-sea-a | ProcessBatchSeconds | 131.307
Catalog2AzureSearch-ea-a | ProcessBatchSeconds | 2,089.71

Also add some sane defaults to the Azure Search configs.

### DIFF
--- a/src/NuGet.Services.AzureSearch/Catalog2AzureSearch/AzureSearchCollectorLogic.cs
+++ b/src/NuGet.Services.AzureSearch/Catalog2AzureSearch/AzureSearchCollectorLogic.cs
@@ -36,6 +36,20 @@ namespace NuGet.Services.AzureSearch.Catalog2AzureSearch
             _options = options ?? throw new ArgumentNullException(nameof(options));
             _telemetryService = telemetryService ?? throw new ArgumentNullException(nameof(telemetryService));
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+            if (_options.Value.MaxConcurrentBatches <= 0)
+            {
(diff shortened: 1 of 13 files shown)


src/NuGet.Services.AzureSearch/Catalog2AzureSearch/AzureSearchCollectorLogic.cs
```diff
@@ -36,6 +36,20 @@ public class AzureSearchCollectorLogic : ICommitCollectorLogic
             _options = options ?? throw new ArgumentNullException(nameof(options));
             _telemetryService = telemetryService ?? throw new ArgumentNullException(nameof(telemetryService));
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+            if (_options.Value.MaxConcurrentBatches <= 0)
+            {
+                throw new ArgumentException(
```
loic-sharma: Super nit: `ArgumentOutOfRangeException`? Same for line 49

joelverhagen: Sounds good.

PR #563, opened 2019-06-18
Fix PackageRegistrationDeprecationMetadata's PackageDeprecationItem constructor

scottbommarito:
Found a minor bug with it after deploying to dev.
Added unit tests to prevent from happening in the future.


src/NuGet.Services.Metadata.Catalog.Monitoring/Model/PackageRegistrationDeprecationMetadata.cs
```diff
@@ -25,7 +25,10 @@ public PackageRegistrationDeprecationMetadata(PackageDeprecationItem deprecation
         {
             Reasons = deprecation.Reasons;
             Message = deprecation.Message;
-            AlternatePackage = new PackageRegistrationAlternatePackageMetadata(deprecation);
+            if (deprecation.AlternatePackageId != null || deprecation.AlternatePackageRange != null)
+            {
+                AlternatePackage = new PackageRegistrationAlternatePackageMetadata(deprecation);
```
loic-sharma: Consider renaming `AlternatePackage` to `AlternatePackageOrNull` to make it clear this may be null.

scottbommarito: It has to be this because this class is deserialized from JSON, and the JSON it is deserialized from has `AlternatePackage` as the property name.

PR #564, opened 2019-06-18
Use the global JSON serializer settings in Azure Search Service

joelverhagen:
Currently `"Debug": null` shows up on all of the responses. This is because the `Json` method uses the default serializer settings, not the ones we set on app start.

Addresses https://github.com/NuGet/Engineering/issues/2513

### DIFF
--- a/src/NuGet.Services.SearchService/App_Start/WebApiConfig.cs
+++ b/src/NuGet.Services.SearchService/App_Start/WebApiConfig.cs
@@ -45,8 +45,7 @@ namespace NuGet.Services.SearchService
         public static void Register(HttpConfiguration config)
         {
             config.Formatters.Remove(config.Formatters.XmlFormatter);
-            config.Formatters.JsonFormatter.SerializerSettings.NullValueHandling = NullValueHandling.Ignore;
-            config.Formatters.JsonFormatter.SerializerSettings.Converters.Add(new StringEnumConverter());
+            SetSerializerSettings(config.Formatters.JsonFormatter.SerializerSettings);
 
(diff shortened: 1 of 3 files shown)


src/NuGet.Services.SearchService/App_Start/WebApiConfig.cs
```diff
@@ -154,6 +159,7 @@ private static IConfigurationRoot GetConfigurationRoot()
         {
             const string prefix = "APPSETTING_";
             var jsonFile = Path.Combine(HostingEnvironment.MapPath("~/"), @"Settings\local.json");
+            jsonFile = @"C:\Users\jver\Desktop\local-dev.json";
```
joelverhagen: WIll remove

src/NuGet.Services.SearchService/Controllers/SearchController.cs
```diff
@@ -197,7 +197,7 @@ private async Task<IHttpActionResult> ExecuteSearchAsync<TResponse>(Func<Task<TR
         {
             try
             {
-                return Json(await searchAction());
+                return Json(await searchAction(), Configuration.Formatters.JsonFormatter.SerializerSettings);
```
loic-sharma: All `Json` calls will need to pass these `SerializerSettings`? 😞

joelverhagen: https://github.com/aspnet/AspNetWebStack/blob/ebef5b7d821b64ed5c48765d0caf6ce8a9bcfaf5/src/System.Web.Http/ApiController.cs#L441-L444

PR #565, opened 2019-06-18
Work around Azure Search SDK bug and return 503

joelverhagen:
Address https://github.com/NuGet/Engineering/issues/2511

@loic-sharma, what do you think about using this filter for `InvalidSearchRequestException`?

### DIFF
--- a/src/NuGet.Services.AzureSearch/Wrappers/DocumentsOperationsWrapper.cs
+++ b/src/NuGet.Services.AzureSearch/Wrappers/DocumentsOperationsWrapper.cs
@@ -5,36 +5,78 @@ using System;
 using System.Threading.Tasks;
 using Microsoft.Azure.Search;
 using Microsoft.Azure.Search.Models;
+using Microsoft.Extensions.Logging;
 
 namespace NuGet.Services.AzureSearch.Wrappers
 {
     public class DocumentsOperationsWrapper : IDocumentsOperationsWrapper
     {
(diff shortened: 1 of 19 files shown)


loic-sharma:
> @loic-sharma, what do you think about using this filter for InvalidSearchRequestException?

Sounds like a great idea! I'd double check that the `ASP.NET Applications \ Errors Unhandled During Execution/sec` perf counter isn't affected by this approach, and I'd add a note on the `InvalidSearchRequestException` type explaining the filter's behavior.

src/NuGet.Services.AzureSearch/Wrappers/DocumentsOperationsWrapper.cs
```diff
@@ -5,36 +5,77 @@
...
+                            "Retrying.",
+                            name,
+                            currentAttempt);
+                        continue;
+                    }
+
+                    // There is a bug where an inner RetryDelegatingHandler fails with NullReferenceException when the
+                    // service is under load and attempting a search. Throw a clearer exception for now so we can
+                    // understand the frequency in logs.
+                    // https://github.com/Azure/azure-sdk-for-net/issues/3224
```
loic-sharma: Should we also add a link to our issue? https://github.com/NuGet/Engineering/issues/2511

joelverhagen: Done.

PR #566, opened 2019-06-19
RegistrationDeprecationValidator should not fail when reasons are out of order

scottbommarito:
SPARQL appears to reshuffle the order of arrays, which causes them to be in different orders in the DB and in registration.

As a result, if you choose "Legacy" and "CriticalBugs", it will reshuffle it into "CriticalBugs" and "Legacy" which will fail the validation.


tests/NgTests/Validation/RegistrationDeprecationValidatorTestData.cs
```diff
@@ -73,6 +73,33 @@ public override IEnumerable<Func<PackageRegistrationIndexMetadata>> CreateIndexe
...
+            {
+                // Multiple reasons can be in different orders
+                foreach (var reasons in GetPossibleDeprecationReasonCombinations().Where(c => c.Count() > 1))
+                {
+                    yield return () => Tuple.Create(
+                        new PackageRegistrationIndexMetadata
+                        {
+                            Deprecation = new PackageRegistrationDeprecationMetadata
+                            {
+                                Reasons = reasons.OrderBy(r => r)
```
loic-sharma: Consider materializing this with a `ToList`. Same comment for line 95

joelverhagen:
We should probably document that the order of the reasons in undefined.

PR #567, opened 2019-06-19
MonitoringProcessorJob should only compare catalog entries if both non-null

scottbommarito:
Minor bug with `MonitoringProcessorJob` that I found while running validations on DEV.

`MonitoringProcessorJob` compares the list of catalog entries in the previous validation run with the catalog entries in the queue message to determine if validations should run. Now that we added the support for validations to be run with a null list of catalog entries, the list of catalog entries in the previous validation can be null, so we should not compare catalog entries and always run the validation in that case.

Ultimately this is not a huge deal, because when an exception is thrown in this part of the code, the validation is eventually requeued with a null `ValidationResult`, in which case it skips the comparison, and it will pass/fail as intended.

I would like to add tests to this class but there are none right now and doing so would be a big undertaking.


loic-sharma:
> I would like to add tests to this class but there are none right now and doing so would be a big undertaking.

Oof... Could we add minimal tests?

scottbommarito:
@loic-sharma - I think adding any sort of useful tests would involve major refactoring of the class, primarily to split up the logic into more testable places. I would prefer not to do it in this PR.

PR #569, opened 2019-06-21
MonitoringProcessor should keep processing queue messages until cancelled

scottbommarito:
I observed the job deadlocking on DEV today and although I wasn't able to find the exact cause, I identified this as a possible mitigation.

Basically, the job spins up 64 threads to process queue messages, each of which completes when there are no messages left in the queue. As a result, if any one of those threads hangs when the queue is empty, the job stops handling messages entirely.

This change makes it so that each thread waits 30 seconds and then continues handling messages when the queue is empty instead of completing. This way, if the queue is empty and one thread hangs, the other threads will be unaffected. Additionally, to avoid the process from getting stale, I added a "queue loop" cancellation token that cancels after 24 hours so that the job loop refreshes and updates configuration.


src/Ng/Jobs/MonitoringProcessorJob.cs
```diff
@@ -27,11 +26,16 @@ namespace Ng.Jobs
...
     {
+        private const int DefaultQueueLoopDurationHours = 24;
+        private const int DefaultEmptyQueueDelaySeconds = 30;
+
         private PackageValidator _packageValidator;
         private IStorageQueue<PackageValidatorContext> _queue;
         private IPackageMonitoringStatusService _statusService;
         private IMonitoringNotificationService _notificationService;
         private CollectorHttpClient _client;
+        private TimeSpan _queueLoopDuration;
```
xavierdecoster: The default value is suffixed with the unit `Hours`. I'd propose to also add this suffix to this private field. --> `_queueLoopDurationHours`.
Same for `Seconds` and `_emptyQueueDelay` --> `_emptyQueueDelaySeconds'

scottbommarito: This is a `TimeSpan`, which means it could be used to get the duration in any value--seconds, hours, minutes, etc. For that reason, I kept the name of this field without any unit, and made the default value const and configuration parameter named with the unit (e.g. `DefaultQueueLoopDurationHours` and `-queueLoopDurationHours`) because they are actually parsed in a way specific to the unit.

scottbommarito:
Despite seeing multiple deadlocks yesterday before this fix within a couple hours, the job has now ran fine without any issues for over 12 hours. I believe this fix has mitigated the issue.

src/Ng/Jobs/MonitoringProcessorJob.cs
```diff
@@ -115,20 +129,68 @@ private static void SetUserAgentString()
...
+                            "Failed to fetch last message or no messages left in queue. Waiting {EmptyQueueDelaySeconds} seconds before polling again.",
+                            _emptyQueueDelay.TotalSeconds);
 
-            Logger.LogInformation("No messages left in queue.");
+                        await Task.Delay(_emptyQueueDelay, queueLoopCancellationToken);
+                    }
+                }
+                catch (Exception e)
+                {
+                    Logger.LogCritical(
```
joelverhagen: There should be a delay in the exception path IMO

scottbommarito: Sounds good.

src/Ng/Jobs/MonitoringProcessorJob.cs
xavierdecoster: Oh, right! 👍

PR #577, opened 2019-07-02
Make LastCommitTimestamp sortable

joelverhagen:
Progress on https://github.com/nuget/nugetgallery/issues/7226

### DIFF
--- a/src/NuGet.Services.AzureSearch/Models/CommittedDocument.cs
+++ b/src/NuGet.Services.AzureSearch/Models/CommittedDocument.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.Azure.Search;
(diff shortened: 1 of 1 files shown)


src/NuGet.Services.AzureSearch/Models/CommittedDocument.cs
```diff
@@ -24,6 +25,7 @@ public abstract class CommittedDocument : KeyedDocument, ICommittedDocument
         public bool? LastUpdatedFromCatalog { get; set; }
 
         [JsonProperty(NullValueHandling = NullValueHandling.Include)]
+        [IsSortable]
```
loic-sharma: Super nit: What do you think of putting `IsSortable` before `JsonProperty`? I think it looks a little nicer to have the shorter attribute first.

joelverhagen: Sure, seems fine.

PR #582, opened 2019-07-09
Add V2 functional tests for azure search

shishirx34:
Addresses: https://github.com/NuGet/NuGetGallery/issues/7233

I verified these tests against all the envs.


tests/NuGet.Services.AzureSearch.FunctionalTests/BasicTests/V2SearchProtocolTests.cs
```diff
@@ -0,0 +1,266 @@
...
+        }
+
+        [Fact]
+        public async Task ShouldGetResultsForEmptyString()
+        {
+            // Act
+            var result = await V2SearchAsync(new V2SearchBuilder { Query = "" });
+
+            // Assert
+            Assert.True(result.TotalHits.HasValue && result.TotalHits.Value > 0, "No results found, should find atleast some results for empty string query.");
```
loic-sharma:
```suggestion
            Assert.True(result.TotalHits.HasValue && result.TotalHits.Value > 0, "No results found, should find at least some results for empty string query.");
```

tests/NuGet.Services.AzureSearch.FunctionalTests/BasicTests/V2SearchProtocolTests.cs
```diff
@@ -0,0 +1,266 @@
...
+
+            Assert.NotNull(resultsWithoutSkip);
+            Assert.NotNull(resultsWithSkip);
+            Assert.True(resultsWithoutSkip.Data.Count > 1);
+            Assert.True(resultsWithSkip.Data.Count > 1);
+            Assert.True(resultsWithoutSkip.Data.Count == resultsWithSkip.Data.Count + skip);
+            var packageIDListThatShouldBeSkipped = resultsWithoutSkip.Data
+                .Select(x => x.PackageRegistration.Id)
+                .Take(skip);
+            var skippedResultPackageIdList = resultsWithSkip.Data
```
loic-sharma: Nit: I would remove the `skippedResultPackageIdList` variable and do the select as you're creating `commonResults`. I don't feel strongly about this, feel free to ignore

tests/NuGet.Services.AzureSearch.FunctionalTests/BasicTests/V2SearchProtocolTests.cs
```diff
@@ -0,0 +1,266 @@
...
+            hasPrereleaseVersions = resultsWithoutPrerelease
+                .Data
+                .Any(x => IsPrerelease(x.Version));
+            Assert.False(hasPrereleaseVersions);
+        }
+
+        [Fact]
+        public async Task ResultsHonorSemverLevel()
+        {
+            var searchTerm = "packageId:BaseTestPackage.SemVer2";
```
loic-sharma: I like that this is simple and low overhead, but this package may be deleted/unlisted without us realizing this will break this functional test. To mitigate this concern, what do you think of renaming this package to something like `TestPackageDoNotRemove.SemVer2`, and then making that package's description a link to this test?

tests/NuGet.Services.AzureSearch.FunctionalTests/BasicTests/V2SearchProtocolTests.cs
```diff
@@ -0,0 +1,266 @@
...
+        {
+            var results = await V2SearchAsync(new V2SearchBuilder { Query = "json", SortBy = orderBy });
+
+            Assert.NotNull(results);
+            Assert.True(results.Data.Count > 1);
+
+            int count = results.Data.Count < 10 ? results.Data.Count : 10;
+            var topResults = results
+                .Data
+                .Select(x => x.GetType().GetProperty(property).GetValue(x, null))
```
loic-sharma: Reflection breaks refactoring tools. Could you instead accept a delegate that gets the results property value? Something like:

```
[Theory]
[InlineData("lastEdited", GetLastEditedPackageData)]
...
public async Task ResultsAreOrderedBySpecifiedParameter(string orderBy, Func<V2Package, object> accessor, bool reverse = false)
{
...
}

...
private object GetLastEditedPackageData(V2Package data) => data.LastEdited;
```

tests/BasicSearchTests.FunctionalTests.Core/Constants.cs
```diff
@@ -13,6 +13,8 @@ public static class Constants
         public const string TestPackageSummary = "";
         public const string TestPackageAuthor = "clayco";
         public const string TestPackageCopyright = "Copyright 2013";
-        public const string NonExistentSearchString = "RandomCharacters234asdfa433";
+        public const string NonExistentSearchString = "asadfasdfsadfwerfasdf23423rafdsf";
```
agr: What if I go and create a package with such ID?

tests/NuGet.Services.AzureSearch.FunctionalTests/BasicTests/V2SearchProtocolTests.cs
loic-sharma: Also, please follow the `Contants.TestPackageId` pattern

tests/NuGet.Services.AzureSearch.FunctionalTests/BasicTests/V2SearchProtocolTests.cs
```diff
@@ -0,0 +1,266 @@
...
+        }
+
+        [Fact]
+        public async Task ResultsMatchIdQueryFieldSearch()
+        {
+            var results = await V2SearchAsync(new V2SearchBuilder { Query = "packageId:" + Constants.TestPackageId });
+
+            Assert.NotNull(results);
+            Assert.True(results.Data.Count == 1);
+            Assert.Equal(Constants.TestPackageId, results.Data.ToList()[0].PackageRegistration.Id, StringComparer.OrdinalIgnoreCase);
```
loic-sharma: Can you use LINQ here?

```suggestion
            Assert.Equal(Constants.TestPackageId, results.Data.First().PackageRegistration.Id, StringComparer.OrdinalIgnoreCase);
```

tests/BasicSearchTests.FunctionalTests.Core/Constants.cs
joelverhagen: For this case I typically use a double hyphen or a double dot:
http://regexstorm.net/tester?p=%5e%5cw%2b%28%5b_.-%5d%5cw%2b%29*%24&i=newtonsoft--json&o=in
https://github.com/NuGet/NuGet.Client/blob/503981224facaff322f0a68c399861fea6369327/src/NuGet.Core/NuGet.Packaging/PackageCreation/Utility/PackageIdValidator.cs#L14

tests/NuGet.Services.AzureSearch.FunctionalTests/Support/NuGetSearchFunctionalTestBase.cs
```diff
@@ -2,10 +2,13 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using BasicSearchTests.FunctionalTests.Core;
+using BasicSearchTests.FunctionalTests.Core.Models;
+using BasicSearchTests.FunctionalTests.Core.TestSupport;
 using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net.Http;
```
loic-sharma: Nit: please sort `using`s with `System.*` first :)

tests/NuGet.Services.AzureSearch.FunctionalTests/BasicTests/V2SearchProtocolTests.cs
```diff
@@ -0,0 +1,266 @@
...
+            var resultsWithoutTestTags = tagResults
+                .Any(x => !x.ToLower().Contains(Constants.TestPackageTag.ToLower()));
+
+            Assert.False(resultsWithoutTestTags);
+        }
+
+        public static bool IsPrerelease(string original)
+        {
+            NuGetVersion nugetVersion;
+            if (!NuGetVersion.TryParse(original, out nugetVersion))
```
loic-sharma: Nit: you can simplify these two lines into

```suggestion
            if (!NuGetVersion.TryParse(original, out var nugetVersion))
```

tests/NuGet.Services.AzureSearch.FunctionalTests/BasicTests/V2SearchProtocolTests.cs
shishirx34: We do use `BaseTestPackage` for our tests, hence the similar name. I would leave it as is. If it does end up breaking in future I would create one, but I think we've been fine for now.

tests/BasicSearchTests.FunctionalTests.Core/Constants.cs
shishirx34: then it would break the test. We've been using the previous search string `RandomCharachters...` for our regular search tests, which didn't return any results. However, same doesn't work with Azure search as it creates tokens like `Random`, `Charachters`... thereby returning some results. I can't think of any search string that can't be broken by creating a package with such a search query unless counting for special characters.

shishirx34: Ah, just saw @joelverhagen's comment. Will look into it.

shishirx34: @joelverhagen - that regex is for the ID though. using hyphens and dots won't work with Azure search anymore since those are stop words thereby creating tokens that could actually match the search term.

shishirx34: had offline chat with @joelverhagen - decided to leave this as is since that would reduce the scope of the test to that of only packageIds

tests/NuGet.Services.AzureSearch.FunctionalTests/BasicTests/V2SearchProtocolTests.cs
```diff
@@ -0,0 +1,272 @@
...
+        {
+            var results = await V2SearchAsync(new V2SearchBuilder { Query = "json", SortBy = orderBy });
+
+            Assert.NotNull(results);
+            Assert.True(results.Data.Count > 1);
+
+            int count = results.Data.Count < 10 ? results.Data.Count : 10;
+            var topResults = results
+                .Data
+                .Select(x => GetPropertyValue(x))
```
loic-sharma:
```suggestion
                .Select(GetPropertyValue)
```

tests/NuGet.Services.AzureSearch.FunctionalTests/BasicTests/V2SearchProtocolTests.cs
```diff
@@ -0,0 +1,272 @@
...
+        }
+
+        [Fact]
+        public async Task ResultsHonorSemverLevel()
+        {
+            var searchTerm = "packageId:" + Constants.TestPackageIdSemVer2;
+            var results = await V2SearchAsync(new V2SearchBuilder { Query = searchTerm, IncludeSemVer2 = true });
+
+            Assert.NotNull(results);
+            Assert.True(results.Data.Count >= 1);
```
joelverhagen: nit: `Assert.True` and `Assert.False` should have helpful messages

tests/NuGet.Services.AzureSearch.FunctionalTests/BasicTests/V2SearchProtocolTests.cs
```diff
@@ -0,0 +1,272 @@
...
+            }
+        }
+
+        [Fact]
+        public async Task ResultsMatchIdQueryFieldSearch()
+        {
+            var results = await V2SearchAsync(new V2SearchBuilder { Query = "packageId:" + Constants.TestPackageId });
+
+            Assert.NotNull(results);
+            Assert.True(results.Data.Count == 1);
```
joelverhagen: Use `Equal` or `Single`

tests/NuGet.Services.AzureSearch.FunctionalTests/BasicTests/V2SearchProtocolTests.cs
```diff
@@ -0,0 +1,272 @@
...
+            Assert.Equal(Constants.TestPackageId, results.Data.First().PackageRegistration.Id, StringComparer.OrdinalIgnoreCase);
+        }
+
+        [Fact]
+        public async Task ResultsMatchAuthorQueryFieldSearch()
+        {
+            var results = await V2SearchAsync(new V2SearchBuilder { Query = "author:" + Constants.TestPackageAuthor });
```
joelverhagen: In the test output are the URLs traced? I find it useful in E2E tests to be able to click a URL of the failure and see the bad results right immediately

tests/NuGet.Services.AzureSearch.FunctionalTests/BasicTests/V2SearchProtocolTests.cs
```diff
@@ -0,0 +1,272 @@
...
+            if (!NuGetVersion.TryParse(original, out var nugetVersion))
+            {
+                return false;
+            }
+
+            return nugetVersion.IsPrerelease;
+        }
+
+        public static bool IsSemVer2(string original)
+        {
```
joelverhagen: This is actually not strictly true. A package can be semver2 if one of its dependency version ranges has a semver2 version.

This is not necessary for a test but please call it out in a comment

tests/NuGet.Services.AzureSearch.FunctionalTests/BasicTests/V2SearchProtocolTests.cs
```diff
@@ -0,0 +1,272 @@
...
+            }
+
+            return nugetVersion.IsSemVer2;
+        }
+
+        public static IEnumerable<object[]> TakeResults
+        {
+            get
+            {
+                for (int i = 1; i <= 10; i++)
```
joelverhagen: nit you could do `Enumerable.Range(1, 10).Select(i => new object[] { i })`

tests/NuGet.Services.AzureSearch.FunctionalTests/NuGet.Services.AzureSearch.FunctionalTests.csproj
```diff
@@ -73,6 +74,9 @@
     <PackageReference Include="NuGet.Services.Configuration">
       <Version>2.43.0</Version>
     </PackageReference>
+    <PackageReference Include="NuGet.Versioning">
```
joelverhagen: Can you check the product code solution and use the NuGet.* version used in those graphs? Having latest here is irrelevant since the version logic almost never changes.

tests/NuGet.Services.AzureSearch.FunctionalTests/Support/NuGetSearchFunctionalTestBase.cs
```diff
@@ -63,6 +65,19 @@ protected async Task<IReadOnlyList<string>> SearchAsync(string query, bool inclu
             }
         }
 
+        protected async Task<V2SearchResult> V2SearchAsync(V2SearchBuilder searchBuilder)
+        {
+            using (var response = await Client.GetAsync(searchBuilder.RequestUri))
+            {
+                response.EnsureSuccessStatusCode();
+
+                var json = await response.Content.ReadAsStringAsync();
```
joelverhagen: I think `ITestOutputHelper` should be passed in here to log the URL and maybe even the response body

joelverhagen: I do something like this for E2E
https://github.com/NuGet/NuGet.Services.EndToEnd/blob/f6895c206c14a5092998cdca5b932fa46c73f19c/src/NuGet.Services.EndToEnd/Support/Clients/SimpleHttpClient.cs#L46-L66

PR #586, opened 2019-07-10
Don't clear catalog commit information in search documents

joelverhagen:
This is necessary because the Auxiliary2AzureSearch job could potentially clear all last commit timestamps on all documents which would lead to an alert being fired on missing last commit timestamp in /search/diag. As a side effect, this also leads to better traceability on the changes made to search documents.


tests/NuGet.Services.AzureSearch.Tests/HijackDocumentBuilderFacts.cs
```diff
@@ -159,11 +159,11 @@ public async Task SetsExpectedProperties()
       ],
       ""title"": ""Windows Azure Storage"",
       ""tokenizedPackageId"": ""WindowsAzure.Storage"",
+      ""lastCommitTimestamp"": null,
```
joelverhagen: The only times we should be setting these values to null is when the record is initialized from DB.

PR #587, opened 2019-07-10
Add better logging for package specific document actions

joelverhagen:


src/NuGet.Services.AzureSearch/Catalog2AzureSearch/CatalogIndexActionBuilder.cs
```diff
@@ -312,30 +329,42 @@ private bool IsUpdateLatest(SearchIndexChangeType changeType)
             NuGetVersion version,
             HijackDocumentChanges changes)
         {
+            IndexAction<KeyedDocument> indexAction;
+
             if (changes.Delete)
             {
```
cristinamanum: Would a switch work better ?

joelverhagen: This is a series of bools and in another method we call a method in the `if` so I don't think it would help much.

PR #588, opened 2019-07-10
[Azure Search] Update relevancy tests

loic-sharma:
I've split tests into two categories:

1. "First results" - These tests that the expected results are first
2. "Top results" - These tests that the expected results are found

I ensured that our tests align with the most frequently selected results, as per our search telemetry.

Addresses https://github.com/nuget/nugetgallery/issues/6980


joelverhagen:
Great idea on these two buckets 🥇

tests/NuGet.Services.AzureSearch.FunctionalTests/Relevancy/V3RelevancyFunctionalTests.cs
```diff
@@ -14,65 +16,74 @@ public V3RelevancyFunctionalTests(CommonFixture fixture, ITestOutputHelper testO
...
         }
 
-        [RelevancyFact]
-        public async Task MicrosoftExtensions()
+        [RelevancyTheory]
+        [MemberData(nameof(EnsureTopResultsData))]
+        public async Task EnsureTopResults(string searchTerm, string[] expectedTopResults)
         {
-            var results = await SearchAsync("Microsoft.Extensions");
+            var results = await SearchAsync(searchTerm);
```
joelverhagen: i think the number of results returned should be clarified here. For example if 100 is returned, this is not a very strong test.

joelverhagen: Would prefer a parameter to `SearchAsync` but a comment can work too (but may be out of date). Perhaps an assertion too like asserting the count is less than or equal to something.

tests/NuGet.Services.AzureSearch.FunctionalTests/Relevancy/V3RelevancyFunctionalTests.cs
```diff
@@ -14,65 +16,74 @@ public V3RelevancyFunctionalTests(CommonFixture fixture, ITestOutputHelper testO
...
+            // TODO: Reduce exact match boost and update "json.net", "entity framework", "redis", and "csv" tests.
+            // See: https://github.com/NuGet/NuGetGallery/issues/7330
+            yield return new object[] { "newtonsoft.json", new[] { "newtonsoft.json" } };
+            yield return new object[] { "newtonsoft", new[] { "newtonsoft.json" } };
+            yield return new object[] { "json.net", new[] { "json.net", "newtonsoft.json" } };
+            yield return new object[] { "json", new[] { "json", "newtonsoft.json" } };
 
-            Assert.NotEmpty(results);
-            Assert.Contains("log4net", results);
+            yield return new object[] { "tags:\"aws-sdk-v3\"", new[] { "awssdk.core", "awssdk.s3" } };
```
shishirx34: did you run this test against dev? I do not see a package "awssdk.s3" in search(at least in gallery)

shishirx34: oh are these tests only meant for prod? forgot.

tests/NuGet.Services.AzureSearch.FunctionalTests/Relevancy/V3RelevancyFunctionalTests.cs
loic-sharma: Added a parameter and explicitly chose the number of results in the test

PR #589, opened 2019-07-11
Measure the durations for /search/diag queries

joelverhagen:


src/NuGet.Services.AzureSearch/Catalog2AzureSearch/CatalogIndexActionBuilder.cs
```diff
@@ -258,8 +258,8 @@ private async Task<IndexChanges> GetIndexChangesAsync(Context context)
                     indexAction = IndexAction.Merge<KeyedDocument>(_search.UpdateVersionListAndOwnersFromCatalog(
                        context.PackageId,
                        searchFilters,
-                       lastCommitTimestamp: context.LatestCommitTimestamp,
-                       lastCommitId: context.LatestCommitId,
+                       lastCommitTimestamp: context.LastCommitTimestamp,
```
joelverhagen: Aligned to `Last` for clarity. `Last` is used in search document field names and status response JSON so it is preferred.

PR #590, opened 2019-07-11
[Azure Search] Remove the freshness boost

loic-sharma:
:warning: This requires an index rebuild to take effect

I verified this by running the new relevancy functional tests plus my search scoring tool. None of the tests I have were affected by this change.

Addresses https://github.com/nuget/nugetgallery/issues/7332


joelverhagen:
Before merging this change to master, we should go through a sampling of SxS queries to make sure we're still good. This can be done at the same time as some other evaluation if you want.

loic-sharma:
I ran the search side-by-side feedback.

Queries that are improved:
* "micro orm" ("Dapper went from 10th to 2nd)
* "common mark" (CommonMark.NET went from 2nd to 1st)
* "adal" (Microsoft.IdentityModel.Clients.ActiveDirectory went from 4th to 2nd)
* "mysql" (pomelo.entityframeworkcore.mysql is third as expected 
* "tags:linqpad-samples" (Octokit went from 5th to 1st)

Queries that regressed:
* "aad" (Microsoft.IdentityModel.Clients.ActiveDirectory went from 2nd to 5th)
* "fxcop analyzers" (Microsoft.CodeAnalysis.FxCopAnalyzers went from 2nd to 3rd)

... TODO, stopped at "xamarin"

PR #592, opened 2019-07-12
[Azure Search 2] Add DownloadSetComparer to detect download count changes

joelverhagen:
While comparing, fine places where download counts decrease and emit telemetry. This is a strange case currently causing some user pain and this will give us more data.

Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/591
Will help us detect https://github.com/NuGet/NuGetGallery/issues/7258 is one more place

### DIFF
--- a/src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/DownloadSetComparer.cs
+++ b/src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/DownloadSetComparer.cs
@@ -1,0 +1,105 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using Microsoft.Extensions.Logging;
+using NuGet.Services.AzureSearch.AuxiliaryFiles;
+
+namespace NuGet.Services.AzureSearch.Auxiliary2AzureSearch
+{
+    public class DownloadSetComparer : IDownloadSetComparer
+    {
+        private readonly IAzureSearchTelemetryService _telemetryService;
+        private readonly ILogger<DownloadSetComparer> _logger;
+
+        public DownloadSetComparer(
+            IAzureSearchTelemetryService telemetryService,
+            ILogger<DownloadSetComparer> logger)
+        {
+            _telemetryService = telemetryService ?? throw new ArgumentNullException(nameof(telemetryService));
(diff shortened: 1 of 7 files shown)


src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/DownloadSetComparer.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        public SortedDictionary<string, long> Compare(
+            DownloadData oldData,
+            DownloadData newData)
+        {
+            var stopwatch = Stopwatch.StartNew();
+
+            // We use a very simplistic algorithm here. Find the union of both ID sets and compare each download count.
+            var uniqueIds = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            foreach (var id in oldData.Keys.Concat(newData.Keys))
```
loic-sharma: What're your thoughts on instead passing this LINQ `Concat` enumerable to the constructor of the `HashSet`?

loic-sharma: Same comment for line 83

tests/NuGet.Services.AzureSearch.Tests/Auxiliary2AzureSearch/DownloadSetComparerFacts.cs
```diff
@@ -0,0 +1,192 @@
...
+                        true,
+                        7,
+                        true,
+                        false,
+                        0),
+                    Times.Once);
+            }
+
+            [Fact]
+            public void DetectsDecreaseDueToMissingId()
```
loic-sharma: Should you test the case where V1 increases by X, but V2 decreases by X, resulting in an empty delta but decrease telemetry? It's an weird/obscure corner case, feel free to ignore

src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/DownloadSetComparer.cs
joelverhagen: Yeah, great idea! I forgot about that overload.

tests/NuGet.Services.AzureSearch.Tests/Auxiliary2AzureSearch/DownloadSetComparerFacts.cs
joelverhagen: No this is a good thing to test. I'll add that.

PR #593, opened 2019-07-15
[Azure Search 3] Read DowloadData from downloads.v1.json

joelverhagen:
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/592

I wanted to leave the existing `download.v1.json` data structure `Downloads` as-is to ensure existing search is unaffected. Instead I added an API to read this file into the more convenient `DownloadData` class I added which is flexible enough for comparison and maintaining original case of IDs.

### DIFF
--- a/src/NuGet.Services.AzureSearch/AuxiliaryFiles/AuxiliaryFileClient.cs
+++ b/src/NuGet.Services.AzureSearch/AuxiliaryFiles/AuxiliaryFileClient.cs
@@ -41,6 +41,27 @@ namespace NuGet.Services.AzureSearch.AuxiliaryFiles
 
         private ICloudBlobContainer Container => _lazyContainer.Value;
 
+        public async Task<DownloadData> LoadDownloadDataAsync()
+        {
+            var result = await LoadAuxiliaryFileAsync(
+                _options.Value.AuxiliaryDataStorageDownloadsPath,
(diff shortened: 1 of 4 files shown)


src/NuGet.Services.AzureSearch/AuxiliaryFiles/AuxiliaryFileClient.cs
```diff
@@ -41,6 +41,27 @@ public class AuxiliaryFileClient : IAuxiliaryFileClient
...
+                _options.Value.AuxiliaryDataStorageDownloadsPath,
+                etag: null,
+                loadData: loader =>
+                {
+                    var downloadData = new DownloadData();
+
+                    Downloads.Load(
+                        name: null,
+                        loader: loader,
+                        addCount: (id, version, count) => downloadData.SetDownloadCount(id, version, count));
```
loic-sharma: Can this be simplified?

```suggestion
                        addCount: downloadData.SetDownloadCount;
```

joelverhagen: No, the `Download` type operates with `int` and `DownloadData` uses `long`. But I will change this method to use `long` so this can work.

PR #595, opened 2019-07-15
[Azure Search 5] Implement Auxiliary2AzureSearchCommand

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/6458
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/594

This is the main command logic. Note that I clean out invalid package IDs (138 exist in PROD), versions (0 exist in PROD), and non-normalized versions (0 exist in PROD).

### DIFF
--- a/src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/Auxiliary2AzureSearchCommand.cs
+++ b/src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/Auxiliary2AzureSearchCommand.cs
@@ -1,17 +1,315 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using NuGet.Packaging;
+using NuGet.Services.AzureSearch.AuxiliaryFiles;
+using NuGet.Services.AzureSearch.Wrappers;
+using NuGet.Services.Metadata.Catalog.Helpers;
+using NuGet.Versioning;
 
 namespace NuGet.Services.AzureSearch.Auxiliary2AzureSearch
 {
     public class Auxiliary2AzureSearchCommand
     {
+        /// <summary>
+        /// A package ID can result in one document per search filter if the there is a version that applies to each
+        /// of the filters. The simplest such case is a prerelease, SemVer 1.0.0 package version like 1.0.0-beta. This
+        /// version applies to all package filters.
+        /// </summary>
+        private static readonly int MaxDocumentsPerId = Enum.GetValues(typeof(SearchFilters)).Length;
+
+        private readonly IAuxiliaryFileClient _auxiliaryFileClient;
+        private readonly IDownloadDataClient _downloadDataClient;
+        private readonly IDownloadSetComparer _downloadSetComparer;
+        private readonly ISearchDocumentBuilder _searchDocumentBuilder;
+        private readonly ISearchIndexActionBuilder _indexActionBuilder;
+        private readonly Func<IBatchPusher> _batchPusherFactory;
+        private readonly ISystemTime _systemTime;
+        private readonly IOptionsSnapshot<Auxiliary2AzureSearchConfiguration> _options;
+        private readonly IAzureSearchTelemetryService _telemetryService;
+        private readonly ILogger<Auxiliary2AzureSearchCommand> _logger;
+
+        public Auxiliary2AzureSearchCommand(
+            IAuxiliaryFileClient auxiliaryFileClient,
+            IDownloadDataClient downloadDataClient,
+            IDownloadSetComparer downloadSetComparer,
+            ISearchDocumentBuilder searchDocumentBuilder,
+            ISearchIndexActionBuilder indexActionBuilder,
+            Func<IBatchPusher> batchPusherFactory,
+            ISystemTime systemTime,
+            IOptionsSnapshot<Auxiliary2AzureSearchConfiguration> options,
+            IAzureSearchTelemetryService telemetryService,
+            ILogger<Auxiliary2AzureSearchCommand> logger)
+        {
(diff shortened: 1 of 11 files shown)


src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/Auxiliary2AzureSearchCommand.cs
```diff
@@ -1,17 +1,287 @@
...
         public async Task ExecuteAsync()
         {
-            // TODO: implement the job
-            // https://github.com/NuGet/NuGetGallery/issues/6447
-            await Task.Yield();
+            var stopwatch = Stopwatch.StartNew();
+            var success = false;
+            try
+            {
+                _logger.LogInformation("Fetching old download count data from blob storage.");
```
loic-sharma: Do you think it's worth adding a comment that `newData` comes from statistics pipeline and `oldResult` is the last indexed download counts? I was a little confused when I first saw this

src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/Auxiliary2AzureSearchCommand.cs
```diff
@@ -1,17 +1,287 @@
...
+                            continue;
+                        }
+
+                        Guard.Assert(indexActions.Hijack.Count == 0, "There should be no hijack index changes.");
+
+                        indexActionsToPush.Add(new IdAndValue<IndexActions>(id, indexActions));
+                        Interlocked.Add(ref documentsToPush, indexActions.Search.Count);
+                    }
+                },
+                _options.Value.MaxConcurrentVersionListWriters);
```
loic-sharma: `MaxConcurrentVersionListWriters` makes it sound like only `MaxConcurrentVersionListWriters` tasks may access the version concurrently, but it's actually `MaxConcurrentVersionListWriters *  MaxConcurrentBatches`. What do you think of renaming this variable to something like `MaxConcurrentVersionListWritersPerBatch`? (I know this is an annoying change since you have to update NuGetDeployment too, sorry :( )

src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/Auxiliary2AzureSearchCommand.cs
joelverhagen: Yes, that's a good idea.

src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/Auxiliary2AzureSearchCommand.cs
joelverhagen: Well it varies actually. In Catalog2AzureSearch there is only one thread using `IBatchPusher` so `MaxConcurrentVersionListWriters` is correctly named. `MaxConcurrentBatches` is used in that case to process package IDs in parallel and yield index changes to be pushed later by a single guy.

In the other jobs you're right it's more `MaxConcurrentVersionListWritersPerBatch`. Early on I thought about coming up with separate parallelism controls per job so that they would be named more clearly but this became not worth IMO the effort since the `BatchPusher` implementation was used everywhere and needed a single definition of `MaxConcurrentVersionListWriters` so the code could be shared.

In short I feel that modification of parallelism values should be done with full cognizance of the effects on the particular job and the name is only the very first step. Hopefully it's "close enough" to get rid of most of the confusion.

Perhaps a code comment on this property in `AzureSearchJobConfiguration` with this sort of detail would be enough?

src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/Auxiliary2AzureSearchCommand.cs
```diff
@@ -1,17 +1,287 @@
...
+
+                // If all of the IDs to process plus the current ID were to need a document for each search filter and
+                // that number plus the current known documents to push would make the batch larger than the maximum
+                // batch size, produce index actions for the IDs that we have collected so far.
+                if (((idsToProcess.Count + 1) * SearchFilterCount) + documentsToPush > _options.Value.AzureSearchBatchSize)
+                {
+                    documentsToPush = await ProcessIdsAsync(idsToProcess, indexActionsToPush, documentsToPush, changes);
+                }
+
+                idsToProcess.Add(idToBatch);
```
loic-sharma: This code is clever and clean, but it's hard to digest as its reversed from the "human" flow:

1. Grab enough ids to process
2. Create index actions from pending ids
3. Push index actions

Is the "human" flow significantly more complex in terms of code? If not, what do you think of aligning with the "human" flow?

loic-sharma: (I bring this up b/c these 20 lines from 132-149 took me like 10 minutes to understand)

src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/Auxiliary2AzureSearchCommand.cs
joelverhagen: Oh another thing to add is that in practice I think Auxiliary2AzureSearch will have `MaxConcurrentBatches` set to 1. We don't want this job to cause too much load on the search index.

loic-sharma: > Perhaps a code comment on this property in `AzureSearchJobConfiguration` with this sort of detail would be enough?

That works for me

src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/Auxiliary2AzureSearchCommand.cs
joelverhagen: > "human" flow

What does that make me? 🤖

joelverhagen: Cleaned it up and added comments. Let me know what you think.

loic-sharma: Looks great, thanks!

PR #596, opened 2019-07-15
[Azure Search 6] Add JobOutcome to be more specific in telemetry

joelverhagen:
Related https://github.com/NuGet/NuGetGallery/issues/6458.
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/595

### DIFF
--- a/src/NuGet.Services.AzureSearch/AzureSearchTelemetryService.cs
+++ b/src/NuGet.Services.AzureSearch/AzureSearchTelemetryService.cs
@@ -78,14 +78,14 @@ namespace NuGet.Services.AzureSearch
                 });
         }
 
-        public void TrackOwners2AzureSearchCompleted(bool success, TimeSpan elapsed)
+        public void TrackOwners2AzureSearchCompleted(JobOutcome outcome, TimeSpan elapsed)
         {
             _telemetryClient.TrackMetric(
                 Prefix + "Owners2AzureSearchCompletedSeconds",
                 elapsed.TotalSeconds,
                 new Dictionary<string, string>
                 {
-                    { "Success", success.ToString() },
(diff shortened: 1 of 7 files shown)


src/NuGet.Services.AzureSearch/AzureSearchTelemetryService.cs
```diff
@@ -346,14 +346,14 @@ public void TrackDownloadSetComparison(int oldCount, int newCount, int changeCou
...
-        public void TrackAuxiliary2AzureSearchCompleted(bool success, TimeSpan elapsed)
+        public void TrackAuxiliary2AzureSearchCompleted(JobOutcome outcome, TimeSpan elapsed)
         {
             _telemetryClient.TrackMetric(
                 Prefix + "Auxiliary2AzureSearchCompletedSeconds",
                 elapsed.TotalSeconds,
                 new Dictionary<string, string>
                 {
-                    { "Success", success.ToString() },
+                    { "Outcome", outcome.ToString() },
```
shishirx34: why not add the string value instead of the enum in the metrics? Its more readable, otherwise one has to know the translation of enums to understand if the outcome succeeded or failed.

joelverhagen: This will result in a readable string. Example: https://dotnetfiddle.net/aynBJL
```
Outcome: NoOp
```

shishirx34: ah, my bad.. default prints string. `ToString("d")` is not the default.

https://docs.microsoft.com/en-us/dotnet/api/system.enum.tostring?view=netframework-4.8

PR #599, opened 2019-07-17
Don't update download counts if there are too many decreases

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/6447

There was a strange case I saw in DEV where there were no download counts in `downloads.v1.json`. I think this is because we do a server-side copy but I'm not sure.


ryuyu:
It varies quite a bit.
Lowest i've seen so far when it is non-zero is 130 ids and 151 versions.
Highest so far is 2775 ids and 3235 versions.

PR #602, opened 2019-07-19
Optimize packageid:{ID} case in V3 search since we get a lot of those

joelverhagen:
This cuts the average query time in nearly a half.
Progress on https://github.com/NuGet/NuGetGallery/issues/6466

Basically we use a document lookup API instead of a full blown search. This is much faster.

## Before

Samples | Average | Min | Max | Std. Dev. | Error % | Throughput | Received KB/sec | Sent KB/sec | Avg. Bytes
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
10000 | 152 | 141 | 363 | 13.23 | 0.00% | 64.61701 | 318.63 | 12.2 | 5049.4

## After

Samples | Average | Min | Max | Std. Dev. | Error % | Throughput | Received KB/sec | Sent KB/sec | Avg. Bytes
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
10000 | 79 | 73 | 280 | 9.75 | 0.00% | 123.8405 | 610.66 | 23.38 | 5049.4

### DIFF
--- a/src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/Auxiliary2AzureSearchCommand.cs
+++ b/src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/Auxiliary2AzureSearchCommand.cs
@@ -271,7 +271,7 @@ namespace NuGet.Services.AzureSearch.Auxiliary2AzureSearch
 
             foreach (var id in data.Keys.ToList())
             {
-                var isValidId = PackageIdValidator.IsValidPackageId(id);
+                var isValidId = id.Length <= PackageIdValidator.MaxPackageIdLength && PackageIdValidator.IsValidPackageId(id);
                 if (!isValidId)
                 {
                     invalidIdCount++;
--- a/src/NuGet.Services.AzureSearch/SearchService/Models/ApiType.cs
+++ b/src/NuGet.Services.AzureSearch/SearchService/Models/ApiType.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Services.AzureSearch.SearchService
+{
+    public enum ApiType
+    {
+        /// <summary>
+        /// The data for the user was fetched using Azure Search's "get document by key" API. The .NET API is called "Get" and
+        /// "GetAsync" but REST API is called "lookup".
+        /// https://docs.microsoft.com/en-us/rest/api/searchservice/lookup-document
+        /// </summary>
+        Get,
+
+        /// <summary>
+        /// The data for the user was fetched using Azure Search's "search documents" API.
+        /// https://docs.microsoft.com/en-us/rest/api/searchservice/search-documents
+        /// </summary>
+        Search,
+    }
+}
--- a/src/NuGet.Services.AzureSearch/SearchService/ParsedQuery.cs
+++ b/src/NuGet.Services.AzureSearch/SearchService/ParsedQuery.cs
@@ -1,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace NuGet.Services.AzureSearch.SearchService
+{
+    public class ParsedQuery
+    {
+        public ParsedQuery(string text, string packageId)
+        {
+            Text = text ?? throw new ArgumentNullException(nameof(text));
+            PackageId = packageId;
+        }
+
+        /// <summary>
+        /// The text that will be provided to Azure Search. This is a Lucene query, not the query provided by the user.
+        /// </summary>
+        public string Text { get; }
+
+        public string PackageId { get; }
+    }
+}
--- a/tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
+++ b/tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
@@ -11,13 +11,18 @@ namespace NuGet.Services.AzureSearch.SearchService
     {
         public class V2Search : FactsBase
         {
+            /// <summary>
+            /// 100 characters is the maximum length for package IDs.
+            /// </summary>
+            private const string MaxId = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
+
             [Theory]
             [MemberData(nameof(CommonAzureSearchQueryData))]
             public void GeneratesAzureSearchQuery(string input, string expected)
             {
                 var actual = _target.V2Search(new V2SearchRequest { Query = input });
 
-                Assert.Equal(expected, actual);
+                Assert.Equal(expected, actual.Text);
             }
 
             [Theory]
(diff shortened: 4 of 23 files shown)


src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/Auxiliary2AzureSearchCommand.cs
```diff
@@ -271,7 +271,7 @@ private void CleanDownloadData(DownloadData data)
 
             foreach (var id in data.Keys.ToList())
             {
-                var isValidId = PackageIdValidator.IsValidPackageId(id);
+                var isValidId = id.Length <= PackageIdValidator.MaxPackageIdLength && PackageIdValidator.IsValidPackageId(id);
```
loic-sharma: Should we update `IsValidPackageId` directly?

src/NuGet.Services.AzureSearch/SearchService/Models/ApiType.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Services.AzureSearch.SearchService
+{
+    public enum ApiType
+    {
+        /// <summary>
+        /// The data for the user was fetched using the "get document by key" API. The .NET API is called "Get" and
```
loic-sharma: It's not super clear you're referring to Azure Search until the link. How about something like:

```suggestion
        /// The data for the user was fetched using Azure Search's "get document by key" API. The .NET API is called "Get" and
```

Same comment for `Search`'s doc

src/NuGet.Services.AzureSearch/SearchService/ParsedQuery.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Services.AzureSearch.SearchService
+{
+    public class ParsedQuery
+    {
+        public ParsedQuery(string text, string packageId)
```
loic-sharma: Should these have null checks?

src/NuGet.Services.AzureSearch/SearchService/ParsedQuery.cs
```diff
@@ -0,0 +1,17 @@
...
+        {
+            Text = text;
+            PackageId = packageId;
+        }
+
+        public string Text { get; }
```
loic-sharma: I would add a comment. Considering the type name, one could assume this is the inputted NuGet query, not the outputted Azure Search query.

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
```diff
@@ -11,13 +11,15 @@ public class SearchTextBuilderFacts
     {
         public class V2Search : FactsBase
         {
+            private const string MaxId = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
```
loic-sharma: This might be clearer with something like:

```suggestion
            private const string MaxId = new string('a', 123);
```

I don't feel strongly about this, feel free to keep it as is if you prefer it

src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/Auxiliary2AzureSearchCommand.cs
joelverhagen: Client code. Not worth since we have unit test coverage.

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
```diff
@@ -31,7 +33,39 @@ public void WhenLuceneQuery_TreatsLeadingIdAsPackageId(bool luceneQuery, string
...
 
-                Assert.Equal(expected, actual);
+                Assert.Equal(expected, actual.Text);
+            }
+
+            [Theory]
+            [InlineData("id:hello", "hello")]
+            [InlineData("packageid:hello", "hello")]
+            [InlineData("packageid:" + MaxId, MaxId)]
+            public void ExtractsPackageIdWhenThereIsOnlyOneClause(string query, string expected)
```
loic-sharma: Should we test if the package id is wrapped in quotes?

src/NuGet.Services.AzureSearch/SearchService/ParsedQuery.cs
joelverhagen: For text, sure.
`packageId` no.

src/NuGet.Services.AzureSearch/SearchService/ParsedQuery.cs
joelverhagen: Actually "query" is how we refer to the user input and the Azure Search SDK accepts "text". But a comment sounds good.

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
joelverhagen: I chose this so that I can use `[InlineData(...)]` which requires a const.

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
joelverhagen: Yeah, good idea.

src/NuGet.Services.AzureSearch/SearchService/ParsedQuery.cs
loic-sharma: For `text`, looks like `packageId` can be null

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
joelverhagen: I left a comment saying the length.

src/NuGet.Services.AzureSearch/SearchService/ParsedQuery.cs
loic-sharma: Ah I missed your comment b/c I didn't reload the page :)

joelverhagen: Oh, I got that backwards in my words. 😨 

`text` cannot be null.
`packageId` can be null.

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
```diff
@@ -31,7 +36,42 @@ public void WhenLuceneQuery_TreatsLeadingIdAsPackageId(bool luceneQuery, string
...
+                Assert.Equal(expected, actual.PackageId);
+            }
+
+            [Theory]
+            [InlineData("id:hello version:1.0.0")]
+            [InlineData("packageid:foo--bar")] // Invalid package ID
+            [InlineData("packageid:" + MaxId + "a")]
+            [InlineData("packageid:hello a")]
+            [InlineData("packageid:\"hello a\"")]
+            [InlineData("hello")]
```
loic-sharma: Should we test if there are multiple package id terms?

src/NuGet.Services.AzureSearch/SearchService/ParsedQuery.cs
loic-sharma: I retracted my upvote

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
joelverhagen: Yes. Added that and more.

PR #604, opened 2019-07-22
Introduce IndexOperationBuilder to encapsulate query optimizations

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/6466.

There is some low hanging fruit for optimizing the most common Azure Search queries:

1. `packageid:` + `ignoreFilter=false` should use document lookup against search index (https://github.com/NuGet/NuGet.Services.Metadata/pull/602)
   - Many non-hijack requests use this pattern, presumably to get download count or latest stable version.
1. `packageid:` + `version:` + `ignoreFilter=true` should use document lookup against hijack index
   - This is the `Packages(Id=,Version=)` hijack case.
1. (maybe) `packageid:` + `ignoreFilter=true`
   - This is the `FindPackagesById()` hijack case.

This PR introduces a new class which will encapsulate this sort of logic. It will parse the query, inspect the groupings, and generate the parameters for the appropriate operation. For "lookup document" it will generate the document key and for "search" it will generate the search text and search parameters.

The performance has not regressed from the previous implementation.

Samples | Average | Min | Max | Std. Dev. | Error % | Throughput | Received KB/sec | Sent KB/sec | Avg. Bytes
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
10000 | 62 | 57 | 402 | 10.44 | 0.00% | 158.6345 | 783.91 | 29.95 | 5060.2

### DIFF
--- a/src/NuGet.Services.AzureSearch/SearchService/IIndexOperationBuilder.cs
+++ b/src/NuGet.Services.AzureSearch/SearchService/IIndexOperationBuilder.cs
@@ -1,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Services.AzureSearch.SearchService
+{
+    /// <summary>
+    /// This interface encapsulates the selection of Azure Search operation as well as the generation of parameters for
+    /// the selected operation. Query optimizations such as looking up a document by key instead of doing a full Lucene
+    /// query are decided here.
+    /// </summary>
+    public interface IIndexOperationBuilder
+    {
+        IndexOperation Autocomplete(AutocompleteRequest request);
+        IndexOperation V2SearchWithHijackIndex(V2SearchRequest request);
+        IndexOperation V2SearchWithSearchIndex(V2SearchRequest request);
+        IndexOperation V3Search(V3SearchRequest request);
+    }
+}
--- a/src/NuGet.Services.AzureSearch/SearchService/IndexOperation.cs
+++ b/src/NuGet.Services.AzureSearch/SearchService/IndexOperation.cs
@@ -1,0 +1,63 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Azure.Search.Models;
+
+namespace NuGet.Services.AzureSearch.SearchService
+{
+    public class IndexOperation
+    {
+        private IndexOperation(
+            IndexOperationType type,
+            string documentKey,
+            string searchText,
+            SearchParameters searchParameters)
+        {
+            Type = type;
+            DocumentKey = documentKey;
+            SearchText = searchText;
+            SearchParameters = searchParameters;
+        }
+
+        /// <summary>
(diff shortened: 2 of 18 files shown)


src/NuGet.Services.AzureSearch/SearchService/IIndexOperationBuilder.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace NuGet.Services.AzureSearch.SearchService
+{
+    public interface IIndexOperationBuilder
```
loic-sharma: >IIndexOperationBuilder [](start = 21, length = 22)

Consider adding a comment here (this type name is rather abstract if you don't know Azure Search)

src/NuGet.Services.AzureSearch/SearchService/IndexOperation.cs
```diff
@@ -0,0 +1,45 @@
...
+            SearchParameters searchParameters)
+        {
+            Type = type;
+            DocumentKey = documentKey;
+            SearchText = searchText;
+            SearchParameters = searchParameters;
+        }
+
+        public IndexOperationType Type { get; }
+        public string DocumentKey { get; }
```
loic-sharma: >DocumentKey [](start = 22, length = 11)

Consider adding a comment like "null if Type is XYZ". Same comment for SearchText and SearchParameters

loic-sharma:
>         ParsedQuery V2Search(V2SearchRequest request);

Maybe rename `V2Search` and `V3Search` to `ParseV2Search` and `ParseV3Search`?

---
Refers to: src/NuGet.Services.AzureSearch/SearchService/ISearchTextBuilder.cs:18 in 9ea7896. [](commit_id = 9ea78969a29dff41d96f7f16b0f8b6a58fa2e30f, deletion_comment = False)

src/NuGet.Services.AzureSearch/SearchService/ParsedQuery.cs
```diff
@@ -2,22 +2,21 @@
...
+            Grouping = grouping ?? throw new ArgumentNullException(nameof(grouping));
         }
 
-        /// <summary>
-        /// The text that will be provided to Azure Search. This is a Lucene query, not the query provided by the user.
-        /// </summary>
-        public string Text { get; }
-
-        public string PackageId { get; }
+        public Dictionary<QueryField, HashSet<string>> Grouping { get; }
```
loic-sharma: What did we land on for immutability for the `Grouping`?

joelverhagen: Mutable. Cases where `packageid:` becomes a filter would should remove `packageid:` from the search text building.

src/NuGet.Services.AzureSearch/SearchService/IndexOperation.cs
joelverhagen: Done.

src/NuGet.Services.AzureSearch/SearchService/IIndexOperationBuilder.cs
joelverhagen: Done.

PR #606, opened 2019-07-22
Use Azure Search $filter for packageid:{id} ignoreFilter=true case

joelverhagen:
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/605.
Progress on https://github.com/NuGet/NuGetGallery/issues/6466.


joelverhagen:
This did not yield the performance gains I hoped for.

PR #609, opened 2019-07-24
[Azure search] Add filter for excluding certain packages from default empty search results

shishirx34:
This PR addresses https://github.com/NuGet/NuGetGallery/issues/7027 
This contains:
1. A new filterable field in the index for excluding the packages by default.
2. Db2AzureSearch change to populate this field from a set list in our auxiliary container. This introduces a new file `ExcludedPackagesList.v1.json`. I will follow up with a config change PR later. 
3. Search service change to set this filter on empty search queries. 

Added tests too. I might add more tests where missing. I have compared the results with the old search in prod and the new results look good.


src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/Auxiliary2AzureSearchConfiguration.cs
```diff
@@ -12,6 +12,7 @@ public class Auxiliary2AzureSearchConfiguration : AzureSearchJobConfiguration, I
         public string AuxiliaryDataStorageContainer { get; set; }
         public string AuxiliaryDataStorageDownloadsPath { get; set; }
         public string AuxiliaryDataStorageVerifiedPackagesPath { get; set; }
+        public string AuxiliaryDataStorageExcludedPackagesListPath { get; }
```
joelverhagen: Having "List" in the name is too verbose. For example, VerifiedPackages is not VerifiedPackagesList. I think in the blob name and in the property names it should be "ExcludedPackages"

src/NuGet.Services.AzureSearch/Db2AzureSearch/NewPackageRegistration.cs
```diff
@@ -9,25 +9,28 @@ namespace NuGet.Services.AzureSearch.Db2AzureSearch
...
     /// </summary>
     public class NewPackageRegistration
     {
         public NewPackageRegistration(
             string packageId,
             long totalDownloadCount,
             string[] owners,
-            IReadOnlyList<Package> packages)
+            IReadOnlyList<Package> packages,
+            bool isExcludedByDefault = false)
```
joelverhagen: nit: no defaults here. the callers are minimal so it's best to be specific.

I followed this pattern everywhere in the Azure Search models because of the null / true / false case we spoke about yesterday.

src/NuGet.Services.AzureSearch/Db2AzureSearch/PackageEntityIndexActionBuilder.cs
```diff
@@ -150,7 +150,8 @@ private static VersionListChange GetVersionListChange(Package x)
                 latestFlags.LatestVersionInfo.FullVersion,
                 package,
                 owners,
-                packageRegistration.TotalDownloadCount));
+                packageRegistration.TotalDownloadCount,
+                isExcludedByDefault: packageRegistration.IsExcludedByDefault));
```
joelverhagen: >isExcludedByDefault [](start = 16, length = 19)

nit: no need to name the parameter since the property passed is the same name

src/NuGet.Services.AzureSearch/SearchService/IndexFields.cs
```diff
@@ -36,6 +36,7 @@ public static class Search
             public static readonly string TotalDownloadCount = Name(nameof(SearchDocument.Full.TotalDownloadCount));
             public static readonly string DownloadScore = Name(nameof(SearchDocument.Full.DownloadScore));
             public static readonly string Versions = Name(nameof(SearchDocument.UpdateLatest.Versions));
+            public static readonly string IsExcludedByDefault = Name(nameof(SearchDocument.Full.IsExcludedByDefault));
```
joelverhagen: nit: lex order

src/NuGet.Services.AzureSearch/SearchService/SearchParametersBuilder.cs
```diff
@@ -37,7 +37,7 @@ public SearchParameters LastCommitTimestamp()
             };
         }
 
-        public SearchParameters V2Search(V2SearchRequest request)
+        public SearchParameters V2Search(V2SearchRequest request, string text)
```
joelverhagen: I think it would be clearer to make the parameter here `isDefaultSearch` or `excludePackagesHiddenByDefault` and move the logic up to `IndexOperationBuilder`. The idea with `IndexOperationBuilder` is that he knows the big picture and he calls into `SearchParametersBuilder` with most decisions already made. Said another way, `SearchParametersBuilder` shouldn't know anything about the search text.

src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
```diff
@@ -12,7 +12,7 @@ namespace NuGet.Services.AzureSearch.SearchService
 {
     public partial class SearchTextBuilder : ISearchTextBuilder
     {
-        private const string MatchAllDocumentsQuery = "*";
+        public static string MatchAllDocumentsQuery = "*";
```
joelverhagen: This should not be public I think. Instead `IndexOperationBuilder` should detect an empty `ParsedQuery.Grouping` for this.

src/NuGet.Services.AzureSearch/ISearchDocumentBuilder.cs
```diff
@@ -27,6 +27,11 @@ public interface ISearchDocumentBuilder
             SearchFilters searchFilters,
             long totalDownloadCount);
 
+        SearchDocument.UpdateIsExcludedByDefault UpdateIsExcludedByDefault(
```
joelverhagen: This is unused right? I think we only have `FullFromDb` knowing about this property today so this method shouldn't be needed.

When Auxiliary2AzureSearch wants to keep this information up to date, we will probably expand the `UpdateDownloadCount` to do both instead of having a separate document update.

src/NuGet.Services.AzureSearch/AzureSearchTelemetryService.cs
```diff
@@ -346,6 +346,27 @@ public void TrackDownloadSetComparison(int oldCount, int newCount, int changeCou
...
+            _telemetryClient.TrackMetric(
+                Prefix + "ReadLatestIndexedExcludedPackagesListSeconds",
+                elapsed.TotalSeconds,
+                new Dictionary<string, string>
+                {
+                    { "PackageIdCount", packageIdCount.ToString() },
+                });
+        }
+
+        public IDisposable TrackReplaceLatestExcludedPackagesList(int packageIdCount)
```
joelverhagen: Unused?

src/NuGet.Services.AzureSearch/Db2AzureSearch/Db2AzureSearchCommand.cs
```diff
@@ -133,17 +140,29 @@ private async Task InitializeAsync()
...
+
+            try
+            {
+                var storageResult = await _auxiliaryFileClient.LoadExcludedPackagesListAsync(etag: null);
+                _excludedPackagesList = storageResult.Data ?? new HashSet<string>();
+            }
+            catch (StorageException ex) when(ex.RequestInformation?.HttpStatusCode == (int)HttpStatusCode.NotFound)
+            {
+                _logger.LogWarning($"Excluded packages list not found in the storage. No packages will be excluded.");
+                _excludedPackagesList = new HashSet<string>();
```
joelverhagen: >_excludedPackagesList [](start = 16, length = 21)

We should fail in this case

src/NuGet.Services.AzureSearch/Db2AzureSearch/Db2AzureSearchCommand.cs
```diff
@@ -133,17 +140,29 @@ private async Task InitializeAsync()
             await _blobContainerBuilder.CreateAsync(containerDeleted);
             await _indexBuilder.CreateSearchIndexAsync();
             await _indexBuilder.CreateHijackIndexAsync();
+
+            try
+            {
+                var storageResult = await _auxiliaryFileClient.LoadExcludedPackagesListAsync(etag: null);
+                _excludedPackagesList = storageResult.Data ?? new HashSet<string>();
```
joelverhagen: >_excludedPackagesList [](start = 16, length = 21)

This shouldn't be a field. It can be just be read in `ExecuteAsync` and passed as a parameter.

src/NuGet.Services.AzureSearch/Db2AzureSearch/NewPackageRegistrationProducer.cs
```diff
@@ -34,6 +34,7 @@ public class NewPackageRegistrationProducer : INewPackageRegistrationProducer
 
         public async Task ProduceWorkAsync(
             ConcurrentBag<NewPackageRegistration> allWork,
+            HashSet<string> excludedPackagesList,
             CancellationToken cancellationToken)
         {
```
joelverhagen: Consider a `Guard.Assert` that makes sure the comparer is OrdinalIgnoreCase

tests/NuGet.Services.AzureSearch.Tests/SearchDocumentBuilderFacts.cs
```diff
@@ -556,11 +557,35 @@ public void UsesIdWhenMissingForSortableTitle(string title)
...
                 Assert.Equal(Data.PackageId.ToLowerInvariant(), document.SortableTitle);
             }
 
+            [Theory]
+            [InlineData(true)]
+            [InlineData(false)]
+            public void SetsIsExcludedByDefaultPropertyCorrectly(bool shouldBeExcluded)
+            {
+                var package = Data.PackageEntity;
+                package.Title = "title";
```
joelverhagen: Shouldn't need to capture `Data.PackageEntity` in a local variable or set `Title`

src/NuGet.Services.AzureSearch/AzureSearchTelemetryService.cs
```diff
@@ -346,6 +346,27 @@ public void TrackDownloadSetComparison(int oldCount, int newCount, int changeCou
                 });
         }
 
+        public void TrackReadLatestIndexedExcludedPackagesList(int packageIdCount, TimeSpan elapsed)
```
joelverhagen: >TrackReadLatestIndexedExcludedPackagesList [](start = 20, length = 42)

This should be called in AuxiliaryFileClient right?

shishirx34: actually not needed. will get rid of these. AuxiliaryFileClient tracks the auxiliary file downloads separately. This was before I was independently downloading aux file.

src/NuGet.Services.AzureSearch/Db2AzureSearch/Db2AzureSearchCommand.cs
shishirx34: I felt earlier we shouldn't.. but now I think that may be you are right, we should fail. Since if we were to miss adding exclusion list our prod results will be horrible.

src/NuGet.Services.AzureSearch/Db2AzureSearch/Db2AzureSearchCommand.cs
```diff
@@ -100,7 +105,11 @@ private async Task ExecuteAsync(CancellationToken token)
                 // Push all package package data to the Azure Search indexes and write the version list blobs.
                 var allOwners = new ConcurrentBag<IdAndValue<IReadOnlyList<string>>>();
                 var allDownloads = new ConcurrentBag<DownloadRecord>();
-                await PushAllPackageRegistrationsAsync(cancelledCts, produceWorkCts, allOwners, allDownloads);
+
+                var storageResult = await _auxiliaryFileClient.LoadExcludedPackagesAsync(etag: null);
+                var excludedPackages = storageResult.Data ?? new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
joelverhagen: Can `LoadExcludedPackagesAsync` really return null data?

src/NuGet.Services.AzureSearch/Models/SearchDocument.cs
```diff
@@ -92,6 +95,17 @@ public class UpdateDownloadCount : UpdatedDocument, IDownloadCount
             public double? DownloadScore { get; set; }
         }
 
+        /// <summary>
+        /// Used when updating just the fields related to the default search exclusion of a document. Note that this model does
+        /// not need any analyzer or other Azure Search attributes since it is not used for index creation. The
+        /// <see cref="Full"/> and its parent classes handle this.
+        /// </summary>
+        [SerializePropertyNamesAsCamelCase]
+        public class UpdateIsExcludedByDefault : UpdatedDocument, IIsExcludedByDefault
```
joelverhagen: I don't think we need this document right? `Full` is enough? I don't think you even need the new interface

src/NuGet.Services.AzureSearch/SearchService/IndexOperationBuilder.cs
```diff
@@ -63,14 +63,14 @@ public IndexOperation V2SearchWithHijackIndex(V2SearchRequest request)
             }
 
             var text = _textBuilder.Build(parsed);
-            var parameters = _parametersBuilder.V2Search(request);
+            var parameters = _parametersBuilder.V2Search(request, parsed.IsEmptyQuery());
```
joelverhagen: @loic-sharma I might have led @shishirx34 astray here -- do you think he should check for `*` for an empty grouping?

src/NuGet.Services.AzureSearch/Models/SearchDocument.cs
shishirx34: I had this for future integration with auxiliary2azuresearch, will get rid of it. I think the interface is useful for explicitly defining the `PopulateIsExcludedByDefault` method. We could merge this with DownloadCount updates, but would avoid doing that until we are ready to add this updation logic aux2azs.

src/NuGet.Services.AzureSearch/Db2AzureSearch/Db2AzureSearchCommand.cs
shishirx34: not really. habituated with null checks 😆

src/NuGet.Services.AzureSearch/SearchService/IndexOperationBuilder.cs
shishirx34: looks like `SearchTextBuilder.Build` does return `*` when the `AzureSearchQueryBuilder` ends up being empty string, the check is after the empty grouping check. Not sure which scenarios this could be, but there is indeed a path that can be missed by simply checking the `parsed.IsEmptyQuery` vs `*` check for text.

https://github.com/NuGet/NuGet.Services.Metadata/blob/dev/src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs#L107

https://github.com/NuGet/NuGet.Services.Metadata/blob/dev/src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs#L150

joelverhagen: I think invalid scoped terms `foo:newtonsoft.json` would yield this. I guess we should do `== ThatConstantForAsterisk` instead of `IsEmptyQuery()`

joelverhagen: Sorry about the confusion -- maybe just dbl check with Loic to be sure. My brain is a little scattered right now.

src/NuGet.Services.AzureSearch/SearchService/IndexOperationBuilder.cs
```diff
@@ -33,7 +33,7 @@ public IndexOperation V3Search(V3SearchRequest request)
             }
 
             var text = _textBuilder.Build(parsed);
-            var parameters = _parametersBuilder.V3Search(request);
+            var parameters = _parametersBuilder.V3Search(request, IsEmptySearchQuery(text));
```
joelverhagen: Oh yeah, this seems best. It's acting the actual query we send to Azure Search and the text information doesn't leak into the parameter builder.

src/NuGet.Services.AzureSearch/SearchService/ParsedQuery.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
```
joelverhagen: Unnecessary?

src/NuGet.Services.AzureSearch/Db2AzureSearch/NewPackageRegistrationProducer.cs
```diff
@@ -34,6 +34,7 @@ public class NewPackageRegistrationProducer : INewPackageRegistrationProducer
 
         public async Task ProduceWorkAsync(
             ConcurrentBag<NewPackageRegistration> allWork,
+            HashSet<string> excludedPackages,
             CancellationToken cancellationToken)
```
loic-sharma: Instead of accepting the `excludedPackages` in this method, consider calling `IAuxiliaryFileClient.LoadExcludedPackagesAsync` in this method to get the excluded packages.

src/NuGet.Services.AzureSearch/SearchService/IndexOperationBuilder.cs
loic-sharma: He should check for `*` due to the tags corner case: searching for `tags:,;|` will result in an non-empty grouping but an empty search query.

src/NuGet.Services.AzureSearch/SearchService/ISearchParametersBuilder.cs
```diff
@@ -8,9 +8,9 @@ namespace NuGet.Services.AzureSearch.SearchService
     public interface ISearchParametersBuilder
     {
         SearchParameters LastCommitTimestamp();
-        SearchParameters V2Search(V2SearchRequest request);
-        SearchParameters V3Search(V3SearchRequest request);
-        SearchParameters Autocomplete(AutocompleteRequest request);
+        SearchParameters V2Search(V2SearchRequest request, bool excludePackagesHiddenByDefault);
```
loic-sharma: These `excludePackagesHiddenByDefault` parameter names are a tad long, is there anything shorter? Maybe `isDefaultSearch` or something?

PR #610, opened 2019-07-25
Refresh secrets in a background thread in Azure Search Service

joelverhagen:
Depends on https://github.com/NuGet/ServerCommon/pull/304.
Address https://github.com/NuGet/NuGetGallery/issues/7337.

This makes Azure Search Service refresh the KeyVault secret cache on a background thread instead of caching them forever. I have seen Azure app server processes live for over 19 days we shouldn't just depend on process restart.

### DIFF
--- a/src/NuGet.Services.AzureSearch/NuGet.Services.AzureSearch.csproj
+++ b/src/NuGet.Services.AzureSearch/NuGet.Services.AzureSearch.csproj
@@ -144,6 +144,7 @@
     <Compile Include="SearchService\IndexOperation.cs" />
     <Compile Include="SearchService\IndexOperationBuilder.cs" />
     <Compile Include="SearchService\IndexOperationType.cs" />
+    <Compile Include="SearchService\ISecretRefresher.cs" />
     <Compile Include="SearchService\Models\DebugDocumentResult.cs" />
     <Compile Include="SearchService\IAuxiliaryDataCache.cs" />
     <Compile Include="AuxiliaryFiles\IAuxiliaryFileClient.cs" />
@@ -186,6 +187,7 @@
     <Compile Include="SearchService\ParsedQuery.cs" />
     <Compile Include="SearchService\SearchTextBuilder.cs" />
     <Compile Include="AzureSearchTelemetryService.cs" />
+    <Compile Include="SearchService\SecretRefresher.cs" />
     <Compile Include="VersionList\Guard.cs" />
     <Compile Include="VersionList\HijackIndexChange.cs" />
     <Compile Include="VersionList\HijackIndexChangeType.cs" />
@@ -234,15 +236,12 @@
       <Version>5.0.3</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Services.Validation.Common.Job">
-      <Version>4.1.0-dev-2701693</Version>
+      <Version>4.1.0-dev-2884457</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Build.Tasks.Pack">
       <Version>4.8.0</Version>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
-    </PackageReference>
-    <PackageReference Include="NuGetGallery.Core">
-      <Version>4.4.5-master-2699519</Version>
     </PackageReference>
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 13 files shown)


src/NuGet.Services.AzureSearch/NuGet.Services.AzureSearch.csproj
```diff
@@ -231,6 +233,9 @@
     <PackageReference Include="Microsoft.Azure.Search">
       <Version>5.0.3</Version>
     </PackageReference>
+    <PackageReference Include="NuGet.Services.KeyVault">
```
joelverhagen: TODO: pull this through NuGet.Jobs.

src/NuGet.Services.AzureSearch/SearchService/ISecretRefresher.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.AzureSearch.SearchService
+{
+    /// <summary>
+    /// Unused to update the secrets in the background periodically.
```
loic-sharma:
```suggestion
    /// Used to update the secrets in the background periodically.
```

src/NuGet.Services.AzureSearch/SearchService/SecretRefresher.cs
```diff
@@ -0,0 +1,64 @@
...
+                    delay = _options.Value.SecretRefreshFrequency;
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogError(0, ex, "An exception was thrown while refreshing the secrets.");
+
+                    delay = _options.Value.SecretRefreshFailureRetryFrequency;
+                }
+
+                if (token.IsCancellationRequested)
```
joelverhagen: What do y'all think about recording the "last refresh time" and reporting it on `/search/diag` for monitoring?

joelverhagen: Considerations:

1. I don't think it's harmful to tell this information publicly. It is _related_ to secrets through and could be used to know that secrets haven't been loaded in a long time.
1. I want to put this on `/search/diag` instead of emitting a metric so that a monitoring runner can report a metric only about instances that are in rotation. This is a lot harder to from metrics emitted from a service.

src/NuGet.Services.SearchService/App_Start/WebApiConfig.cs
```diff
@@ -175,17 +168,35 @@ private static IConfigurationRoot GetConfigurationRoot()
...
+            }
 
             // Reload the configuration with secret injection enabled. This is was is used by the application.
             var injectedBuilder = new ConfigurationBuilder()
                 .AddInjectedJsonFile(jsonFile, secretInjector)
                 .AddInjectedEnvironmentVariables(prefix, secretInjector);
             var injectedConfiguration = injectedBuilder.Build();
 
-            return injectedConfiguration;
+            // No disable all secrets loads from a non-refresh path. Refresh will be called periodically from a
```
scottbommarito: `No disable` -> `Disable`

src/NuGet.Services.AzureSearch/SearchService/SecretRefresher.cs
shishirx34: May be report this as "LastServiceRefreshTime" on /search/diag, so its not obvious that this is about _secrets_

joelverhagen: Open source, but sure sounds fine!

src/NuGet.Services.AzureSearch/SearchService/SecretRefresher.cs
```diff
@@ -0,0 +1,64 @@
...
+                {
+                    return;
```
cristinamanum: This state seems to be an error state because the refresh action should not be cancelled. Should it be monitored?

skofman1:
Please don't merge before AME migration is completed. Otherwise this functionality will break.

src/NuGet.Services.AzureSearch/SearchService/SecretRefresher.cs
joelverhagen: This token is returned by the background worker system. It will be cancelled when the app is shutting down.

PR #613, opened 2019-07-26
[Azure Search] Support queries with long terms

loic-sharma:
This requires an index rebuild to take effect. I rebuilt my index using PROD data, and tested the side-by-side feedback and the top queries as per our Gallery browser telemetry.

The queries that improved:

* entityframeworkcore

The queries that regressed: N/A

Addresses https://github.com/NuGet/NuGetGallery/issues/7380


src/NuGet.Services.AzureSearch/Analysis/TruncateCustomTokenFilter.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// Truncates tokens to 300 characters or less. This is necessary as Azure Search's
+    /// <see cref="TruncateTokenFilter"/> defaults to 10 characters.
+    /// </summary>
+    public static class TruncateCustomTokenFilter
+    {
+        public const string Name = "nuget_truncate_filter";
+
+        public static TruncateTokenFilter Instance = new TruncateTokenFilter(
+            Name,
+            length: 300);
```
scottbommarito: Any reason you chose 300?

src/NuGet.Services.AzureSearch/Analysis/TruncateCustomTokenFilter.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Azure.Search.Models;
+
+namespace NuGet.Services.AzureSearch
+{
+    /// <summary>
+    /// Truncates tokens to 300 characters or less. This is necessary as Azure Search's
+    /// <see cref="TruncateTokenFilter"/> defaults to 10 characters.
```
shishirx34: The REST API [documentation](https://docs.microsoft.com/en-us/azure/search/index-add-custom-analyzers#TokenFilters) states. Is there another place that states 10 as the default?

``
truncate | TruncateTokenFilter | Truncates the terms into a specific length.

Options 

length (type: int) - Default: 300, maximum: 300. Required.
``

src/NuGet.Services.AzureSearch/Analysis/TruncateCustomTokenFilter.cs
loic-sharma: Two reasons:

1. It's the max allowed length for a token 
2. It's the documented default value of Azure Search's `TruncateTokenFilter` length (see [this](https://docs.microsoft.com/en-us/azure/search/index-add-custom-analyzers))... but their documentation is wrong 😢

shishirx34: should be reported. :(

src/NuGet.Services.AzureSearch/Analysis/TruncateCustomTokenFilter.cs
loic-sharma: The `TruncateTokenFilter` passes `null` down to the API by default. The value of 10 isn't documented anywhere, it seems to be a bug (I've emailed the Azure Search folks about it)

src/NuGet.Services.AzureSearch/Analysis/TruncateCustomTokenFilter.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// </summary>
+    public static class TruncateCustomTokenFilter
```
joelverhagen: Integration test?

src/NuGet.Services.AzureSearch/Analysis/TruncateCustomTokenFilter.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        public const string Name = "nuget_truncate_filter";
```
joelverhagen: We can add this to the integration test suite, right?

loic-sharma: Yes, in fact, I believe the existing integration tests break because of this change. Thanks for the reminder :)

src/NuGet.Services.AzureSearch/Analysis/TruncateCustomTokenFilter.cs
loic-sharma: I've sent the Azure Search team an email about it

src/NuGet.Services.AzureSearch/Analysis/TruncateCustomTokenFilter.cs
loic-sharma: Added

src/NuGet.Services.AzureSearch/Analysis/TruncateCustomTokenFilter.cs
loic-sharma: Added

PR #616, opened 2019-07-31
[Azure] Boost results that match all tokens in single term queries

loic-sharma:
⚠️ This MUST be merged after https://github.com/NuGet/NuGet.Services.Metadata/pull/617 ⚠️ 

Queries that improved:
* `system.drawing`
* `socket.io`
* `System.Net.Http.Formatting`
* `Npgsql.EntityFrameworkCore.PostgreSQL` - 2nd and 3rd results are more relevant now
* `system.web.mvc` - Results are relevant now, but most relevant result is 3rd

Queries that could've improved but didn't:
* `windows.extensions` - The most relevant result, `System.Windows.Extensions`, has 1.5K downloads
* `System.Web` - There are many results that match with high download counts. Results aren't spectacular

Queries that regressed:
* `microsoft.aspnetcore`- The most relevant result, `microsoft.aspnetcore`, is now 4th

Addresses https://github.com/NuGet/NuGetGallery/issues/7391


tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchTextBuilderFacts.cs
```diff
@@ -232,7 +232,10 @@ public static IEnumerable<object[]> CommonAzureSearchQueryData()
                     { "title:foo bar", "+title:foo bar" },
                     { "title:foo unknown:bar", "title:foo" },
 
-                    // If there are non-field-scoped terms and no field-scoped terms, at least of one the non-field-scoped terms is required.
+                    // If there is a single unscoped term, favor results that match all token fragments after splitting on periods.
+                    { "Foo.Bar", "Foo.Bar (+Foo +Bar)^2" },
```
joelverhagen: Tests for leading dots `.Foo`, trailing dots `Foo.`, double dots `Foo..Bar`, just dots `..`, quotes around `"Foo.Bar"`

joelverhagen: Looks like we need the "dots but multiple terms" case `Foo.Bar Baz`. Maybe i missed it

loic-sharma: Good catch, I've added more tests!

PR #618, opened 2019-08-01
[Azure Search]Add V3 Search Functional tests

shishirx34:
Addresses https://github.com/NuGet/NuGetGallery/issues/2798 

I verified these against all the three environments.

### DIFF
--- a/tests/BasicSearchTests.FunctionalTests.Core/TestSupport/V3SearchBuilder.cs
+++ b/tests/BasicSearchTests.FunctionalTests.Core/TestSupport/V3SearchBuilder.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Collections.Specialized;
+using System.Web;
 
 namespace BasicSearchTests.FunctionalTests.Core.TestSupport
 {
@@ -17,18 +18,18 @@ namespace BasicSearchTests.FunctionalTests.Core.TestSupport
 
         protected override NameValueCollection GetQueryString()
         {
-            var queryString = System.Web.HttpUtility.ParseQueryString(string.Empty);
+            var queryString = HttpUtility.ParseQueryString(string.Empty);
             queryString["q"] = Query;
             queryString["prerelease"] = Prerelease.ToString();
 
             if (Skip.HasValue)
             {
-                queryString["Skip"] = Skip.ToString();
+                queryString["skip"] = Skip.ToString();
             }
 
             if (Take.HasValue)
             {
-                queryString["Take"] = Take.ToString();
+                queryString["take"] = Take.ToString();
             }
 
             if (IncludeSemVer2)
--- a/tests/BasicSearchTests.FunctionalTests.Core/TestSupport/V3SearchBuilder.cs
+++ b/tests/BasicSearchTests.FunctionalTests.Core/TestSupport/V3SearchBuilder.cs
@@ -1,12 +1,42 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
+using System.Collections.Specialized;
 
 namespace BasicSearchTests.FunctionalTests.Core.TestSupport
 {
     public class V3SearchBuilder : QueryBuilder
     {
+        public int? Skip { get; set; }
+
+        public int? Take { get; set; }
+
+        public bool IncludeSemVer2 { get; set; }
+
         public V3SearchBuilder() : base("/query?") { }
+
+        protected override NameValueCollection GetQueryString()
+        {
+            var queryString = System.Web.HttpUtility.ParseQueryString(string.Empty);
+            queryString["q"] = Query;
+            queryString["prerelease"] = Prerelease.ToString();
+
+            if (Skip.HasValue)
+            {
+                queryString["Skip"] = Skip.ToString();
+            }
+
+            if (Take.HasValue)
+            {
+                queryString["Take"] = Take.ToString();
+            }
+
+            if (IncludeSemVer2)
+            {
+                queryString["semVerLevel"] = "2.0.0";
+            }
+
+            return queryString;
+        }
     }
 }
--- a/tests/NuGet.Services.AzureSearch.FunctionalTests/BasicTests/V3SearchProtocolTests.cs
+++ b/tests/NuGet.Services.AzureSearch.FunctionalTests/BasicTests/V3SearchProtocolTests.cs
@@ -1,0 +1,322 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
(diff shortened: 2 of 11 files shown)


tests/BasicSearchTests.FunctionalTests.Core/TestSupport/V3SearchBuilder.cs
```diff
@@ -1,12 +1,42 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
+using System.Collections.Specialized;
 
 namespace BasicSearchTests.FunctionalTests.Core.TestSupport
 {
     public class V3SearchBuilder : QueryBuilder
     {
```
loic-sharma: A V3 search client already exists in the Azure Search functional tests, can we converge these? See: https://github.com/NuGet/NuGet.Services.Metadata/blob/master/tests/NuGet.Services.AzureSearch.FunctionalTests/Support/NuGetSearchFunctionalTestBase.cs#L43

loic-sharma: Idea: if/when you converge these client libraries, you could create a new `NuGet.Protocol.Search` library that follows the [`NuGet.Protocol.Catalog`](https://github.com/NuGet/NuGet.Services.Metadata/tree/master/src/NuGet.Protocol.Catalog) pattern. I would move the search service [request/response models](https://github.com/NuGet/NuGet.Services.Metadata/tree/master/src/NuGet.Services.AzureSearch/SearchService/Models) to this new project and have your client library implement [`ISearchService`](https://github.com/NuGet/NuGet.Services.Metadata/blob/master/src/NuGet.Services.AzureSearch/SearchService/ISearchService.cs).

For example: [`SearchClient.SearchAsync`](https://github.com/loic-sharma/BaGet/blob/controller-refactor/src/BaGet.Protocol/Search/SearchClient.cs#L45) accepts a [`SearchRequest`](https://github.com/loic-sharma/BaGet/blob/controller-refactor/src/BaGet.Protocol/Search/SearchRequest.cs) and returns a [`SearchResponse`](https://github.com/loic-sharma/BaGet/blob/controller-refactor/src/BaGet.Protocol/Search/SearchResponse.cs). Notice that the client and service reuse the exact same interfaces and models.

shishirx34: For the first one the PR already does the refactoring for existing V3 search client, see: https://github.com/NuGet/NuGet.Services.Metadata/pull/618/files#diff-066ba730cfa64b581d1b955f35f6085fR60 
I did the bare minimum here so as to not modify existing relevancy tests and break it given that those can be fickle.

For second, I like the idea. But, I would rather not bloat up this PR with those refactorings.

tests/BasicSearchTests.FunctionalTests.Core/TestSupport/V3SearchBuilder.cs
```diff
@@ -1,12 +1,42 @@
...
+
+        protected override NameValueCollection GetQueryString()
+        {
+            var queryString = System.Web.HttpUtility.ParseQueryString(string.Empty);
+            queryString["q"] = Query;
+            queryString["prerelease"] = Prerelease.ToString();
+
+            if (Skip.HasValue)
+            {
+                queryString["Skip"] = Skip.ToString();
```
joelverhagen: nit: lower `skip`

tests/BasicSearchTests.FunctionalTests.Core/TestSupport/V3SearchBuilder.cs
```diff
@@ -1,12 +1,42 @@
...
+            }
+
+            if (Take.HasValue)
+            {
+                queryString["Take"] = Take.ToString();
```
joelverhagen: nit: lower `take`

tests/NuGet.Services.AzureSearch.FunctionalTests/BasicTests/V3SearchProtocolTests.cs
```diff
@@ -0,0 +1,322 @@
...
+            Assert.True(results.Data.Count == take, $"The search result did not return the expected {take} results");
+        }
+
+        [Fact]
+        public async Task SkipDoesSkipThePackagesInResult()
+        {
+            var searchTerm = "json";
+            var skip = 5;
+            var resultsWithoutSkip = await V3SearchAsync(new V3SearchBuilder { Query = searchTerm, Take = 10 });
+            var resultsWithSkip = await V3SearchAsync(new V3SearchBuilder { Query = searchTerm, Skip = skip, Take = 10 - skip });
```
joelverhagen: There is a small race condition here, results can change over time but I think it's acceptable

tests/NuGet.Services.AzureSearch.FunctionalTests/BasicTests/V3SearchProtocolTests.cs
```diff
@@ -0,0 +1,322 @@
...
+            hasPrereleaseVersions = resultsWithoutPrerelease
+                .Data
+                .Any(x => TestUtilities.IsPrerelease(x.Version));
+            Assert.False(hasPrereleaseVersions, $"The search query returned results with prerelease versions when queried for Prerelease = false");
+        }
+
+        [Theory]
+        [InlineData("packageid:" + Constants.TestPackageId_Unlisted)]
+        [InlineData("packageid:" + Constants.TestPackageId_Unlisted + " version:" + Constants.TestPackageVersion_Unlisted)]
+        public async Task HidesUnlistedPackagesByDefault(string query)
```
joelverhagen: Thanks for porting these ones over too!

tests/BasicSearchTests.FunctionalTests.Core/TestSupport/V3SearchBuilder.cs
```diff
@@ -1,12 +1,42 @@
...
+
+        public int? Take { get; set; }
+
+        public bool IncludeSemVer2 { get; set; }
+
         public V3SearchBuilder() : base("/query?") { }
+
+        protected override NameValueCollection GetQueryString()
+        {
+            var queryString = System.Web.HttpUtility.ParseQueryString(string.Empty);
```
loic-sharma: Can we add a `using` for this namespace or is there a name collision?

shishirx34: should be fine to be `using`.. I copied it from someplace, didn't notice.

tests/NuGet.Services.AzureSearch.FunctionalTests/BasicTests/V3SearchProtocolTests.cs
shishirx34: you mean if the search ranking changes between line 101 and 102? That is an extremely miniscule possibility :)

shishirx34: talked offline. The possibility of mismatch would be in cases where the search score is tied for the results, in such cases its possible the results might get swapped at the threshold. This is very unlikely to happen with azure search.

PR #623, opened 2019-08-09
[Hotfix] Remove invalid package IDs and versions from old data

joelverhagen:
Db2AzureSearch initializes the downloads.v2.json with all of the data found in the downloads.v1.json.

This means that the first time Auxiliary2AzureSearch runs, it can encounter invalid package IDs in the "old" data set. It already filters out package IDs in the new data set. I made it filter it out in the old data set too.

Another fix would have been cleaning the data in Db2AzureSearch but I chose not to do this since it would require re-running Db2AzureSearch (index rebuild process). This is not worth the effort when I can just fix it here. As an added benefit, it means that in the off change that we change our package ID validation rules, the next Auxiliary2AzureSearch run will apply the rules to both download data sets.

Address https://github.com/NuGet/Engineering/issues/2639

### DIFF
--- a/src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/Auxiliary2AzureSearchCommand.cs
+++ b/src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/Auxiliary2AzureSearchCommand.cs
@@ -90,6 +90,9 @@ namespace NuGet.Services.AzureSearch.Auxiliary2AzureSearch
                 _logger.LogInformation("Fetching new download count data from blob storage.");
                 var newData = await _auxiliaryFileClient.LoadDownloadDataAsync();
 
+                _logger.LogInformation("Removing invalid IDs and versions from the old data.");
+                CleanDownloadData(oldResult.Result);
+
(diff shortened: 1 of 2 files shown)


src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/Auxiliary2AzureSearchCommand.cs
```diff
@@ -90,6 +90,9 @@ public async Task ExecuteAsync()
                 _logger.LogInformation("Fetching new download count data from blob storage.");
                 var newData = await _auxiliaryFileClient.LoadDownloadDataAsync();
 
+                _logger.LogInformation("Removing invalid IDs and versions from the old data.");
+                CleanDownloadData(oldResult.Result);
```
shishirx34: how does the invalid data make it into these files?

joelverhagen: It comes from stats DB.
https://github.com/NuGet/Engineering/issues/2638

PR #627, opened 2019-08-13
Add embedded icon URLs to Azure Search from DB (package entity)

scottbommarito:
https://github.com/nuget/nugetgallery/issues/7283


src/NuGet.Services.AzureSearch/BaseDocumentBuilder.cs
```diff
@@ -175,16 +179,21 @@ public BaseDocumentBuilder(IOptionsSnapshot<AzureSearchJobConfiguration> options
...
-                document.IconUrl = new Uri(_options.Value.ParseFlatContainerBaseUrl(), iconPath).AbsoluteUri;
+                SetIconUrlFromFlatContainer(document);
             }
             else
             {
                 document.IconUrl = leaf.IconUrl;
             }
         }
 
+        private void SetIconUrlFromFlatContainer(IBaseMetadataDocument document)
```
zhhyu: Maybe name it as "Get" method to keep the same convention as other items?

scottbommarito: The primary purpose of this method is to set document's icon URL (see `document.IconUrl = ...`), although it does have to fetch the icon URL as part of the method.

zhhyu: Do we need this functionality for other V3 jobs, like catalog2registration? If so, maybe a helper class/function is better rather than a private method here?

src/Catalog/NuGet.Services.Metadata.Catalog.csproj
```diff
@@ -266,7 +266,7 @@
       <Version>1.0.8.3533</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Services.Entities">
-      <Version>4.4.5-master-2699519</Version>
+      <Version>4.4.5-dev-2939401</Version>
```
joelverhagen: Today, the later version of NuGetGallery.Core brings this stuff directly to the NuGet.Services.AzureSearch project. We should take that path to ensure a) these older entry points remain unaffected and b) the rest of the NuGetGallery.Core code that uses NuGet.Services.Entities is consistent. In short, NuGet.Services.AzureSearch has a NuGetGallery.Core older than this NuGet.Services.Entities. This should not happen.

tests/NuGet.Services.AzureSearch.Tests/BaseDocumentBuilderFacts.cs
```diff
@@ -160,6 +161,83 @@ public void AllowNullDependencyGroups()
...
+
+                Assert.Equal(
+                    $"{Data.FlatContainerBaseUrl}{Data.FlatContainerContainerName}/{Data.PackageId.ToLowerInvariant()}/{Data.NormalizedVersion.ToLowerInvariant()}/icon",
+                    full.IconUrl);
+            }
+
+            [Fact]
+            public void IfLeafDoesNotHaveIconFile_UsesIconUrl()
+            {
+                var iconUrl = "iconUrl";
```
zhhyu: Nit change: why do you define a local variable for "IconUrl" here but for the test above, you directly set "IconUrl" at the constructor.

src/NuGet.Services.AzureSearch/BaseDocumentBuilder.cs
zhhyu: Maybe define the "Get" function and use "document.IconUrl = GetIconUrlFromFlatContainer(***)" to align with other set-up? Just nit changes.

tests/NuGet.Services.AzureSearch.Tests/BaseDocumentBuilderFacts.cs
```diff
@@ -160,6 +161,83 @@ public void AllowNullDependencyGroups()
...
+                    IconFile = "iconFile",
+                    IconUrl = "iconUrl"
+                };
+
+                var full = new HijackDocument.Full();
+
+                Target.PopulateMetadata(full, Data.NormalizedVersion, leaf);
+
+                Assert.Equal(
+                    $"{Data.FlatContainerBaseUrl}{Data.FlatContainerContainerName}/{Data.PackageId.ToLowerInvariant()}/{Data.NormalizedVersion.ToLowerInvariant()}/icon",
```
joelverhagen: nit: all of the inputs for the expected URL are in `Data`. You can maybe make an `FlatContainerIconUrl` property (or better name) to define it once for both assertions in your change set

src/Catalog/NuGet.Services.Metadata.Catalog.csproj
joelverhagen: To be clear this means updating NuGetGallery.Core in jobs and pulling it in through NuGet.Services.AzureSearch's Validation.Common.Job dependency.

joelverhagen: Actually I think the NuGet.Jobs update will get done with https://github.com/NuGet/NuGet.Jobs/pull/796, if you want me to push that through first.

joelverhagen: I merged it. You should be able to use Validation.Common.Job from this build:
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2942767

scottbommarito: I was planning on updating these versions when a master version of `NuGet.Services.Entities` was released, but you're saying I need to update the `NuGetGallery.Core` references as well?

scottbommarito: Okay will do!

joelverhagen: Since NuGetGallery.Core is used in this repository by Azure Search, it is not guaranteed that the entities changes you pull into the graph are compatible with NuGetGallery.Core. Perhaps these changes are... but it's not worth threading the needle like that when you can move them both forward.

src/NuGet.Services.AzureSearch/BaseDocumentBuilder.cs
scottbommarito: I was thinking of doing that, but given that we're passing the `document` object to the function anyway, I felt that there wasn't too much point in passing it back to do the assignment.

loic-sharma: This file has the `Get*` pattern to set a property already, I agree with @zhhyu that style consistency would be preferable.

src/Catalog/NuGet.Services.Metadata.Catalog.csproj
joelverhagen: We don't need this change now.

src/NuGet.Indexing/NuGet.Indexing.csproj
```diff
@@ -190,7 +190,7 @@
       <Version>2.2.2</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Services.Entities">
-      <Version>4.4.5-master-2699519</Version>
+      <Version>4.4.5-dev-2939401</Version>
```
joelverhagen: We don't need this change now.

src/Catalog/NuGet.Services.Metadata.Catalog.csproj
scottbommarito: Moved the `NuGetGallery.Core` version in `NuGet.Services.AzureSearch` forward. Is that all you needed me to do?

To be honest, I think we should maybe consider having some root project with like these so we can push versions of dependencies forward without worrying about conflicts like these.

tests/NuGet.Services.AzureSearch.Tests/BaseDocumentBuilderFacts.cs
scottbommarito: done!

src/NuGet.Indexing/NuGet.Indexing.csproj
scottbommarito: Discussed offline--it appears that due to the fact that `4.4.5-master-2699519` is more recent than `4.4.5-dev-2939401` according to SemVer, these changes were necessary to force NuGet to resolve to the `4.4.5-dev-2939401` properly.

src/NuGet.Services.AzureSearch/BaseDocumentBuilder.cs
```diff
@@ -175,16 +179,21 @@ public BaseDocumentBuilder(IOptionsSnapshot<AzureSearchJobConfiguration> options
...
+        {
+            var provider = new FlatContainerPackagePathProvider(_options.Value.FlatContainerContainerName);
```
skofman1: shouldn't this be initialized in the constructor?

scottbommarito: Not sure--it was not previously, and I didn't want to change the behavior. Potentially, the `_options.Value.FlatContainerContainerName` could change and then require this to be reinitialized, but I don't know if that's possible.

@joelverhagen @loic-sharma @shishirx34 - is that a possibility?

joelverhagen: You could since configs never change at runtime. But IIRC there are some option interfaces in Microsoft.Extensions.Options/Configuration that can give different values over time so I personally try to "new up" classes like these in method bodies we we can have fewer "caches"

Ideally `FlatContainerPackagePathProvider` would itself accept a config option in ctor and leave this kind of decision as its own implementation detail.

scottbommarito: Given that, I'd rather leave this as-is.

PR #632, opened 2019-08-15
[Azure Search 1] Read auxiliary data from local storage in Azure Search Service

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/2635

The goal of this PR is to make Azure Search Service read auxiliary data from the local (region-wise) storage account instead of from the shared (USNC) auxiliary data storage account. These files are kept up to date in the local storage account via Auxiliary2AzureSearch.

Some significant model changes were necessary so that the `AuxiliaryFileMetadata` is returned for the downloads.v2.json file and the verified packages file. This metadata is like "last modified" and "file size" blob details which are displayed on the `/search/diag` page.

This change requires a configuration update for Azure Search Service so that it has the connection string for the local storage account. I have a PR out for that in NuGetDeployment internal repo.

There will be subsequent PRs: 

1. Move from `SortedDictionary` to `Dictionary` in `DownloadData`
2. Allow the auxiliary reloader to control the lifetime of deduped string instances (big memory win)

### DIFF
--- a/src/NuGet.Services.AzureSearch/AuxiliaryFiles/DownloadDataClient.cs
+++ b/src/NuGet.Services.AzureSearch/AuxiliaryFiles/DownloadDataClient.cs
@@ -41,7 +41,7 @@ namespace NuGet.Services.AzureSearch.AuxiliaryFiles
 
         private ICloudBlobContainer Container => _lazyContainer.Value;
 
-        public async Task<ResultAndAccessCondition<DownloadData>> ReadLatestIndexedAsync()
+        public async Task<AuxiliaryFileResult<DownloadData>> ReadLatestIndexedAsync(IAccessCondition accessCondition)
         {
             var stopwatch = Stopwatch.StartNew();
             var blobName = GetLatestIndexedBlobName();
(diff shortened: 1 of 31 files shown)


src/NuGet.Services.AzureSearch/AuxiliaryFiles/AccessConditionExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.WindowsAzure.Storage;
+using NuGetGallery;
+
+namespace NuGet.Services.AzureSearch.AuxiliaryFiles
+{
+    internal static class SimpleCloudBlobExtensions
```
loic-sharma: >SimpleCloudBlobExtensions [](start = 26, length = 25)

Nit: This type's name doesn't match this file's name.

src/NuGet.Services.AzureSearch/AuxiliaryFiles/DownloadDataClient.cs
```diff
@@ -41,36 +41,45 @@ public class DownloadDataClient : IDownloadDataClient
...
-            IAccessCondition accessCondition;
+            AuxiliaryFileMetadata metadata;
             try
             {
-                using (var stream = await blobReference.OpenReadAsync(AccessCondition.GenerateEmptyCondition()))
+                using (var stream = await blobReference.OpenReadAsync(accessCondition))
                 {
-                    accessCondition = AccessConditionWrapper.GenerateIfMatchCondition(blobReference.ETag);
                     ReadStream(stream, downloads.SetDownloadCount);
+                    notModified = false;
```
loic-sharma: >notModified = false; [](start = 20, length = 20)

This `notModified = false;` hurts my brain a little (or maybe it's lack of sleep?), what do you think of renaming everything to `modified` to avoid the double negation?

loic-sharma: Maybe add comments on `AuxiliaryFileResult`'s properties if you do this change

---
In reply to: [314494827](https://github.com/NuGet/NuGet.Services.Metadata/pull/632#discussion_r314494827) [](ancestors = 314494827)

PR #634, opened 2019-08-15
[Azure Search 3] Dedupe auxiliary data strings in Azure Search Service

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/2635
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/633

This saves 87.0% of string allocations due to versions found in the `DownloadData` type from my last test. I am emitting a counter to measure this over time.

The auxiliary file cache used in the Azure Search Service holds on to the string cache as a instance field but since the cache is a singleton in DI, this cache lives forever and will "leak" package IDs that are removed from the data file. This is existing behavior from the auxiliary data file reader but was done with `string.Intern` which _cannot_ be purged. This leak is acceptable because the number of deleted package IDs is very low. In general the number of package IDs in the download data file will always go up since deletes do not propagate to the statistics DB.

We're still ahead memory wise from the previous implementation because we filter out invalid package IDs.

### DIFF
--- a/src/NuGet.Services.AzureSearch/AuxiliaryFiles/StringCache.cs
+++ b/src/NuGet.Services.AzureSearch/AuxiliaryFiles/StringCache.cs
@@ -1,0 +1,78 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Concurrent;
+using System.Threading;
+
+namespace NuGet.Services.AzureSearch.AuxiliaryFiles
+{
+    public class StringCache
+    {
+        /// <summary>
+        /// Maintain a lookup of strings for de-duping. We maintain the original case for de-duping purposes by using
+        /// the default string comparer. As of July of 2019 in PROD, maintaining original case of version
+        /// string adds less than 0.3% extra strings. De-duping version strings in a case-sensitive manner removes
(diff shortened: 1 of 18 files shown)


src/NuGet.Services.AzureSearch/AuxiliaryFiles/StringCache.cs
```diff
@@ -0,0 +1,57 @@
...
+                Interlocked.Increment(ref _hitCount);
+            }
+
+            return output;
+        }
+
+        /// <summary>
+        /// Resets <see cref="RequestCount"/> and <see cref="HitCount"/> back to zero.
+        /// </summary>
+        public void ResetCounts()
```
loic-sharma: What's this for? The only usage I could find was from a test.

joelverhagen: I should call this after emitting telemetry. I forgot to. Will fix.

tests/NuGet.Services.AzureSearch.Tests/SearchService/AuxiliaryDataCacheFacts.cs
```diff
@@ -121,6 +122,48 @@ public async Task InitializesAgainWhenAlreadyInitialized()
                 Assert.True(_target.Initialized);
                 VerifyReadWithETag();
             }
+
+            [Fact]
+            public async Task ResetsStringCacheCounts()
```
loic-sharma: 💯

PR #636, opened 2019-08-20
[Hotfix] Don't allow skip value over 100k

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/7466.

Tested:

- nuget.exe list (reduced max to 200)
- VS PM UI against V3 (reduced max to 200)
- VS PM UI against V2 (reduced max to 200)
- gallery (skipped page 5000 then next'ed forward twice)
- ran Azure Search functional tests

### DIFF
--- a/tests/NuGet.Services.AzureSearch.Tests/SearchService/AzureSearchServiceFacts.cs
+++ b/tests/NuGet.Services.AzureSearch.Tests/SearchService/AzureSearchServiceFacts.cs
@@ -17,6 +17,26 @@ namespace NuGet.Services.AzureSearch.SearchService
         public class V2SearchAsync : BaseFacts
         {
             [Fact]
+            public async Task SearchIndexAndEmptyOperation()
+            {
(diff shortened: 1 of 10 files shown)


tests/NuGet.Services.AzureSearch.Tests/SearchService/AzureSearchServiceFacts.cs
```diff
@@ -16,6 +16,23 @@ public class AzureSearchServiceFacts
     {
         public class V2SearchAsync : BaseFacts
         {
+            [Fact]
+            public async Task SearchIndexAndEmptyOperation()
```
loic-sharma: Do you think it's worth testing that no search/hijack operations were performed? I don't feel strongly about adding this, feel free to skip

joelverhagen: Sure, sounds good.

PR #638, opened 2019-08-21
Revert "[Azure] Boost results that match all tokens in single term queries"

joelverhagen:
Reverts NuGet/NuGet.Services.Metadata#616


loic-sharma:
😭

PR #640, opened 2019-08-22
Don't allow retries or wrap exceptions when pushing to Azure Search

joelverhagen:
Help the debugging of https://github.com/NuGet/Engineering/issues/2664

This broke the expectation of `BatchPusher`:
https://github.com/NuGet/NuGet.Services.Metadata/blob/c6e3f85d41549d24b7c19d7c21ff740598965a1d/src/NuGet.Services.AzureSearch/BatchPusher.cs#L217-L230

This made debugging harder because the individual failure messages were not logged since they not part of the `Exception` string. `BatchPusher` handles the `IndexBatchException` exception and gives useful logging.


joelverhagen:
Also fixed a build warning about web.config in Azure Search.

PR #648, opened 2019-09-05
Default search should include cases when isExcludedByDefault is null

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/7506.

I tried `isExcludedByDefault ne true` but that did not perform as well as this approach.


joelverhagen:
> I wouldn't be opposed to this fix either.

Tracked here: https://github.com/NuGet/NuGetGallery/issues/7384

PR #651, opened 2019-09-05
[SDL] Catalog2Dnx uses SHA512 rather than MD5

zhhyu:
Update the corresponding "AreSynchronized" method to use SHA512 for source/destination blob check of Catalog2Dnx. Test coverage is not easy to achieve because of the dependency on the blob itself and SDK. 

Verify it locally.


src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -389,7 +390,12 @@ public override async Task<bool> AreSynchronized(Uri firstResourceUri, Uri secon
             {
                 if (await source.ExistsAsync())
                 {
-                    return !string.IsNullOrEmpty(source.Properties.ContentMD5) && source.Properties.ContentMD5 == destination.Properties.ContentMD5;
+                    var sourceBlobMetadata = source.Metadata;
```
loic-sharma: Is this code path tested?

zhhyu: Yeah, the "ExistsAsync()" will fetch the properties and metadata so we can get the metadata later as it is. Verify it locally.
The [document](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blob-container-properties-metadata) is here.

src/Catalog/NuGet.Services.Metadata.Catalog.csproj
```diff
@@ -300,6 +300,9 @@
     <PackageReference Include="NuGet.Services.Logging">
       <Version>2.25.0-master-30088</Version>
     </PackageReference>
+    <PackageReference Include="NuGetGallery.Core">
```
skofman1: why do we need this dependency? If it's only for CoreConstants.Sha512HashAlgorithmId lets just copy the constant. Extra dependencies create maintenance overhead.

zhhyu: Yeah, you are right. The only usage of this dependency is "CoreConstants" for the key name. If you think hard coded key name is fine, I will use it here.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -389,7 +390,37 @@ public override async Task<bool> AreSynchronized(Uri firstResourceUri, Uri secon
...
+                    if (sourceBlobMetadata == null || destinationBlobMetadata == null)
+                    {
+                        return false;
+                    }
+
+                    var sourceBlobHasSha512Hash = sourceBlobMetadata.TryGetValue(Sha512HashAlgorithmId, out var sourceBlobSha512Hash);
+                    var destinationBlobHasSha512Hash = destinationBlobMetadata.TryGetValue(Sha512HashAlgorithmId, out var destinationBlobSha512Hash);
+                    if (!sourceBlobHasSha512Hash)
+                    {
+                        Trace.WriteLine(string.Format("The source blob ({0}) doesn't have the Sha512 hash.", source.Uri.ToString()));
```
skofman1: make the unexpected behaviors Warning level

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -389,7 +390,37 @@ public override async Task<bool> AreSynchronized(Uri firstResourceUri, Uri secon
...
+                    }
+                    if (!destinationBlobHasSha512Hash)
+                    {
+                        Trace.WriteLine(string.Format("The destination blob ({0}) doesn't have the Sha512 hash.", destination.Uri.ToString()));
```
joelverhagen: nit: `SHA512` instead of `Sha512`

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -389,7 +390,37 @@ public override async Task<bool> AreSynchronized(Uri firstResourceUri, Uri secon
...
-                    return !string.IsNullOrEmpty(source.Properties.ContentMD5) && source.Properties.ContentMD5 == destination.Properties.ContentMD5;
+                    var sourceBlobMetadata = source.Metadata;
+                    var destinationBlobMetadata = destination.Metadata;
+                    if (sourceBlobMetadata == null || destinationBlobMetadata == null)
+                    {
+                        return false;
+                    }
+
+                    var sourceBlobHasSha512Hash = sourceBlobMetadata.TryGetValue(Sha512HashAlgorithmId, out var sourceBlobSha512Hash);
+                    var destinationBlobHasSha512Hash = destinationBlobMetadata.TryGetValue(Sha512HashAlgorithmId, out var destinationBlobSha512Hash);
```
joelverhagen: Could we log the two hashes?

src/Catalog/Persistence/AzureStorage.cs
cristinamanum: nit : you can return false here and at line 405. In that case you would not need the `if (sourceBlobHasSha512Hash && destinationBlobHasSha512Hash)` branch below

zhhyu: Yeah, the reason that I prefer not to return earlier is that if both of source and destination have no SHA512 hash, we want to log both of them rather than just the source one and return immediately.

src/Catalog/Persistence/AzureStorage.cs
zhhyu: Log when the two hashes are not equal.

zhhyu:
Some efforts to provide some-level decoupling and improve the test coverage on the legacy AzureStorage. Test it locally. Please have another review. Thanks! @skofman1 @joelverhagen @cristinamanum 
One thing could be improved is that we can introduce "Gallery.Core" and enrich the "ICloudBlobClient" and the corresponding wrapper, where the Azure search depends on it currently.

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -379,21 +380,57 @@ public override Uri GetUri(string name)
...
             // For interacting with the source, we just use the same blob request options as the destination blob.
             ApplyBlobRequestOptions(source);
 
-            if (await destination.ExistsAsync())
+            return await AreSynchronized(new AzureCloudBlockBlob(source), new AzureCloudBlockBlob(destination));
+        }
+
+        public async Task<bool> AreSynchronized(ICloudBlockBlob sourceBlockBlob, ICloudBlockBlob destinationBlockBlob)
+        {
+            if (await destinationBlockBlob.ExistsAsync(CancellationToken.None))
```
skofman1: This code can be much more readable if it's restructured a bit:

bool destinationExists = await destinationBlockBlob.ExistsAsync(CancellationToken.None);
bool sourceExists = await sourceBlockBlob.ExistsAsync(CancellationToken.None);

if (!destinationExists || !sourceExists)
{
return destinationExists && sourceExists;
}

 var sourceBlobMetadata = await sourceBlockBlob.GetMetadataAsync(CancellationToken.None);
 var destinationBlobMetadata = await destinationBlockBlob.GetMetadataAsync(CancellationToken.None);

 if (sourceBlobMetadata == null || destinationBlobMetadata == null)
{
       return false;
}

if (!sourceBlobHasSha512Hash)
 {
                        Trace.TraceWarning(string.Format("The source blob ({0}) doesn't have the SHA512 hash.", sourceBlockBlob.Uri.ToString()));
}
if (!destinationBlobHasSha512Hash)
{
                       Trace.TraceWarning(string.Format("The destination blob ({0}) doesn't have the SHA512 hash.", destinationBlockBlob.Uri.ToString()));
  }

if (sourceBlobHasSha512Hash && destinationBlobHasSha512Hash)
{
         if (sourceBlobSha512Hash == destinationBlobSha512Hash)
          {
                   Trace.WriteLine(string.Format("The source blob ({0}) and destination blob ({1}) have the same SHA512 hash and are synchronized.", sourceBlockBlob.Uri.ToString(), destinationBlockBlob.Uri.ToString()));
                    return true;
           }

          Trace.TraceWarning(string.Format("The source blob ({0}) and destination blob ({1}) have the different SHA512 hash and are not synchronized. " +
                            "The source blob hash is {2} while the destination blob hash is {3}",
                            sourceBlockBlob.Uri.ToString(), destinationBlockBlob.Uri.ToString(), sourceBlobSha512Hash, destinationBlobSha512Hash));
                    }
else
{
                    return false;
}

src/Catalog/Persistence/AzureStorage.cs
```diff
@@ -379,21 +380,57 @@ public override Uri GetUri(string name)
...
+                    if (sourceBlobHasSha512Hash && destinationBlobHasSha512Hash)
+                    {
+                        if (sourceBlobSha512Hash == destinationBlobSha512Hash)
+                        {
+                            Trace.WriteLine(string.Format("The source blob ({0}) and destination blob ({1}) have the same SHA512 hash and are synchronized.",
+                                sourceBlockBlob.Uri.ToString(), destinationBlockBlob.Uri.ToString()));
+                            return true;
+                        }
+
+                        Trace.TraceWarning(string.Format("The source blob ({0}) and destination blob ({1}) have the different SHA512 hash and are not synchronized. " +
```
loic-sharma: >TraceWarning [](start = 30, length = 12)

Should this be a `TraceWarning`? Blob mutations seem like a normal operation, considering making this a `WriteLine`

zhhyu: Yeah, because it's an unexpected behavior, so we make it as a warning.

tests/CatalogMetadataTests/AzureStorageFacts.cs
```diff
@@ -0,0 +1,150 @@
...
+        [Theory]
+        [InlineData(true, true, "SHA512Value1", true, true, "SHA512Value1", true)]
+        [InlineData(true, true, "SHA512Value1", true, true, "SHA512Value2", false)]
+        [InlineData(false, false, null, true, true, "SHA512Value1", true)]
+        [InlineData(true, true, "SHA512Value1", false, false, null, false)]
+        [InlineData(false, false, null, false, false, null, true)]
+        [InlineData(true, false, null, true, true, "SHA512Value1", false)]
+        [InlineData(true, true, "SHA512Value1", true, false, null, false)]
+        [InlineData(true, false, null, true, false, null, false)]
+        public async void ValidateAreSynchronizedmethod(bool isSourceBlobExisted,
```
loic-sharma: >isSourceBlobExisted [](start = 61, length = 19)

What do you think of the name `sourceBlobExists`? Same comment for `isDestinationBlobExisted`

zhhyu: Ok

src/Catalog/Persistence/AzureStorage.cs
loic-sharma: Ah okay. Maybe add a comment since that's not evident

src/Catalog/Persistence/AzureStorage.cs
zhhyu: This will break the current logic. If destination exists, but source doesn't exist, it should return true. If both of them don't exist, it should return true.
Decide to keep the current one.

PR #652, opened 2019-09-11
Mega-boost the package with ID matching the query containing symbols

joelverhagen:
This is the first step in addressing the exact match search issue: https://github.com/NuGet/NuGetGallery/issues/7463.

Mega-boost the package with ID matching the query containing symbols. Possible symbols in package IDs are dot (`.`), underscore (`_`), and hyphen (`-`).

Our current plan of action is:

1. Roll this change out on a fast track (no A/B testing).
   - We have high confidence in this change from our offline NDCG metric, user reports, and analysis of today's query patterns
1. A/B test the exact match on all package IDs (not just IDs with symbols).
   - We are being more cautious on this case since we had a lot of feedback during the preview stage saying some cases like "json" or "entity" should not go first.
1. Work on a UI to clearly label exact matches.
1. Reconsider exact match on all package IDs if not done yet.


src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
```diff
@@ -143,6 +145,17 @@ public string Build(ParsedQuery parsed)
                 }
             }
 
+            // Handle the exact match case. If the search query is a single unscoped term is is also a valid package
```
zhhyu: is is

joelverhagen: Fixed.

src/NuGet.Services.AzureSearch/SearchService/AzureSearchTextBuilder.cs
```diff
@@ -105,6 +105,25 @@ public void AppendBoostIfMatchAllTerms(IReadOnlyList<string> terms, float boost)
                 _result.Append(boost);
             }
 
+            /// <summary>
+            /// Append a clouse to boost an exact matched package ID.
```
shishirx34: nit: clause

PR #654, opened 2019-09-13
Add ServiceClientTracingLogger to log Azure Search requests and responses

joelverhagen:
Related to https://github.com/NuGet/Engineering/issues/2664.

### DIFF
--- a/src/NuGet.Jobs.Auxiliary2AzureSearch/Job.cs
+++ b/src/NuGet.Jobs.Auxiliary2AzureSearch/Job.cs
@@ -1,9 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Net;
-using System.Threading.Tasks;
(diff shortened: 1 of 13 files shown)


src/NuGet.Services.AzureSearch/ICommand.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
+namespace NuGet.Services.AzureSearch
+{
+    public interface ICommand
```
scottbommarito: nit: Do you think it would be better if this was `IAzureSearchCommand` instead of just `ICommand`? I thought this was [System.Windows.Input.ICommand](https://docs.microsoft.com/en-us/dotnet/api/system.windows.input.icommand?view=netframework-4.8) at first.

joelverhagen: Sure, sounds good!

PR #655, opened 2019-09-13
Custom JsonConverter to be able to read certain funny catalog leaves.

agr:
Currently, we can't read the following catalog leaf using the `CatalogClient`:
https://api.nuget.org/v3/catalog0/data/2018.03.11.05.06.09/fluentconsoleapplication.0.1.0.json

This fixes the issue.


src/NuGet.Protocol.Catalog/Serialization/BareStringOrArrayConcatenatingConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// </summary>
+    public class BareStringOrArrayConcatenatingConverter : JsonConverter
+    {
+        public override bool CanConvert(Type objectType)
+        {
+            return objectType == typeof(string);
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
```
joelverhagen: If a registration page has one of these, how does client handle it? I think we should try to mimic the behavior of client if it permissive here.

scottbommarito: https://api.nuget.org/v3/registration3-gz-semver2/fluentconsoleapplication/index.json

Registration doesn't have `releaseNotes`, apparently.

scottbommarito:
Not a huge fan of this change. Our protocol intends for this field to be a string and not an array, and by introducing this converter instead of fixing our data, we are essentially changing our protocol. We should fix up any blobs with this issue and make sure the bug doesn't reproduce still. My understanding is that this was caused by when a nuspec contains an element twice, and we've since fixed the bug.

src/NuGet.Protocol.Catalog/Serialization/BareStringOrArrayConcatenatingConverter.cs
joelverhagen: Oh, my bad. What about if `description` had this problem? My point is array concatenation seems like an arbitrary decision and we should try to be consistent with anything we already have. 

Also, how does client or client APIs handle such a .nuspec on disk?

scottbommarito: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Protocol/Model/PackageSearchMetadata.cs

Looks like it would fail in the same way that this throws.

agr: Updated to pick the first.

agr:
I scanned the whole catalog to figure out the extent of the issue, it turned out that leaf was the only one with such issues in release notes. We decided to fix the data instead.

PR #656, opened 2019-09-17
Icon ingestion job

agr:
`Catalog2IconJob` is the entry point for the job. It instantiates the `IconsCollector` which organizes processing of catalog leaves in catalog page size batches. It retrieves the catalog leaf and delegates its processing to `CatalogLeafDataProcessor`.
`CatalogLeafDataProcessor` when the package is IconUrl package tries to read the IconUrl content (and handles the errors and caching). When the package is Icon file, it does the same as catalog2dnx currently does. It also deletes icons when catalog leaf indicates the package was deleted.
Creation of the icon files in flat container is handled by `IconProcessor`.


src/Ng/Jobs/Catalog2IconJob.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        protected override void Init(IDictionary<string, string> arguments, CancellationToken cancellationToken)
+        {
+            ServicePointManager.DefaultConnectionLimit = DegreeOfParallelism;
+
+            var verbose = arguments.GetOrDefault(Arguments.Verbose, false);
+            var packageStorageBase = arguments.GetOrThrow<string>(Arguments.ContentBaseAddress);
+            var auxStorageFactory = CreateAuxStorageFactory(arguments, verbose);
+            var targetStorageFactory = CreateTargetStorageFactory(arguments, verbose);
+            var packageStorage = new AzureStorage(new Uri(packageStorageBase), TimeSpan.FromMinutes(15), TimeSpan.FromMinutes(10), false, false, true, null);
```
skofman1: named params please

src/Catalog/Icons/CatalogLeafDataProcessor.cs
```diff
@@ -0,0 +1,268 @@
...
+        private readonly ILogger<CatalogLeafDataProcessor> _logger;
+
+        public CatalogLeafDataProcessor(
+            IAzureStorage packageStorage,
+            IIconProcessor iconProcessor,
+            IExternalIconContentProvider externalIconContentProvider,
+            IIconCopyResultCache iconCopyResultCache,
+            ITelemetryService telemetryService,
+            ILogger<CatalogLeafDataProcessor> logger
+            )
```
skofman1: nit:')'  on next row

src/Catalog/Icons/CatalogLeafDataProcessor.cs
```diff
@@ -0,0 +1,268 @@
...
+            _iconCopyResultCache = iconCopyResultCache ?? throw new ArgumentNullException(nameof(iconCopyResultCache));
+            _telemetryService = telemetryService ?? throw new ArgumentNullException(nameof(telemetryService));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public async Task ProcessPackageDelete(Storage storage, CatalogCommitItem item, CancellationToken cancellationToken)
+        {
+            var targetStoragePath = GetTargetStorageIconPath(item);
+            await _iconProcessor.DeleteIcon(storage, targetStoragePath, cancellationToken, item.PackageIdentity.Id, item.PackageIdentity.Version.ToNormalizedString());
+            // it would be nice to remove the icon copy result from cache for this item, but we don't have a an icon URL here,
```
skofman1: >a [](start = 107, length = 1)

a

src/Catalog/Icons/IconCopyResultCache.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            if (ExternalIconCopyResults.TryGetValue(iconUrl, out var result))
+            {
+                return result;
+            }
+
+            return null;
+        }
+
+        public void StoreCachedResult(Uri iconUrl, ExternalIconCopyResult newItem)
```
skofman1: AddCachedResult

src/Catalog/Icons/IconCopyResultCache.cs
```diff
@@ -0,0 +1,87 @@
...
+            using (var contentStream = content.GetContentStream())
+            using (var reader = new StreamReader(contentStream))
+            {
+                var serializer = new JsonSerializer();
+                var dictionary = (Dictionary<Uri, ExternalIconCopyResult>)serializer.Deserialize(reader, typeof(Dictionary<Uri, ExternalIconCopyResult>));
+                ExternalIconCopyResults = new ConcurrentDictionary<Uri, ExternalIconCopyResult>(dictionary);
+            }
+        }
+
+        public async Task StoreIconUrlCacheAsync(CancellationToken cancellationToken)
```
skofman1: SaveAsync

src/Catalog/Icons/IconCopyResultCache.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        private readonly IStorage _auxStorage;
+
+        public IconCopyResultCache(
+            IStorage auxStorage)
+        {
+            _auxStorage = auxStorage ?? throw new ArgumentNullException(nameof(auxStorage));
+        }
+
+        public async Task InitializeIconUrlCacheAsync(CancellationToken cancellationToken)
```
skofman1: Generally, the method names here can be simplified: Initialize, SaveAsync, Get, Set

src/Catalog/Helpers/Retry.cs
```diff
@@ -0,0 +1,60 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.Metadata.Catalog.Helpers
+{
+    public class Retry
```
cristinamanum: Can Polly be used? If not maybe add comments why not.

src/Catalog/Helpers/Retry.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        public static async Task IncrementalAsync(
```
cristinamanum: Comments for the methods please.

src/Catalog/Helpers/Retry.cs
```diff
@@ -0,0 +1,60 @@
...
+            TimeSpan initialWaitInterval,
+            TimeSpan waitIncrement)
+        {
+            for (int currentRetry = 0; currentRetry < maxRetries; ++currentRetry)
+            {
+                try
+                {
+                    await runLogicAsync();
+                    return;
+                }
```
cristinamanum: Duplicate code between the two methods. 
Maybe invoke return `await IncrementalAsync(runLogicAsync, shouldRetryOnException, x=>false, maxRetries, initialWaitInterval, waitIncrement) `

Or `await IncrementalAsync(runLogicAsync, shouldRetryOnException, null, maxRetries, initialWaitInterval, waitIncrement) `and null check in the method below.

src/Catalog/Icons/CatalogLeafDataProcessor.cs
```diff
@@ -0,0 +1,273 @@
...
+            _packageStorage = packageStorage ?? throw new ArgumentNullException(nameof(packageStorage));
+            _iconProcessor = iconProcessor ?? throw new ArgumentNullException(nameof(iconProcessor));
+            _externalIconContentProvider = externalIconContentProvider ?? throw new ArgumentNullException(nameof(externalIconContentProvider));
+            _iconCopyResultCache = iconCopyResultCache ?? throw new ArgumentNullException(nameof(iconCopyResultCache));
+            _telemetryService = telemetryService ?? throw new ArgumentNullException(nameof(telemetryService));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public async Task ProcessPackageDelete(Storage storage, CatalogCommitItem item, CancellationToken cancellationToken)
+        {
```
cristinamanum: ProcessPackageDeleteAsync instead of ProcessPackageDelete?

src/Catalog/Icons/CatalogLeafDataProcessor.cs
```diff
@@ -0,0 +1,273 @@
...
+            _packageStorage = packageStorage ?? throw new ArgumentNullException(nameof(packageStorage));
+            _iconProcessor = iconProcessor ?? throw new ArgumentNullException(nameof(iconProcessor));
+            _externalIconContentProvider = externalIconContentProvider ?? throw new ArgumentNullException(nameof(externalIconContentProvider));
+            _iconCopyResultCache = iconCopyResultCache ?? throw new ArgumentNullException(nameof(iconCopyResultCache));
+            _telemetryService = telemetryService ?? throw new ArgumentNullException(nameof(telemetryService));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public async Task ProcessPackageDelete(Storage storage, CatalogCommitItem item, CancellationToken cancellationToken)
```
cristinamanum: Does this method deleted the icon or the entire package? If only the icon maybe the method name can be changed to reflect this?

src/Catalog/Helpers/Retry.cs
agr: This will not work. One can't do `result = await runLogicAsync()` if `runLogicAsync` returns just `Task`. Also, I wouldn't even be able to call the overload that accepts `Func<Task<T>>` as the first argument, because `Func<Task>` cannot be cast to `Func<Task<T>>`.

src/Catalog/Helpers/Retry.cs
agr: This follows the [pattern](https://github.com/NuGet/NuGet.Services.Metadata/blob/master/src/NuGet.Indexing/Retry.cs) we already have in NuGet.Services.Metadata. Pulling an external library for a tool that we'll run for a little while and discard seems a bit excessive.

src/Catalog/Icons/CatalogLeafDataProcessor.cs
agr: How about `ProcessPackageDeleteLeafAsync`? The class is the leaf processor, so method emphasizes that it processes leaves.

src/Catalog/Icons/CatalogLeafDataProcessor.cs
agr: It processes the catalog leaf that indicated that the package was deleted. Renamed.

src/Ng/Jobs/Catalog2IconJob.cs
```diff
@@ -0,0 +1,90 @@
...
+                LoggerFactory.CreateLogger<CatalogLeafDataProcessor>());
+
+            _collector = new IconsCollector(
+                new Uri(source),
+                TelemetryService,
+                targetStorageFactory,
+                catalogClient,
+                leafProcessor,
+                iconCopyResultCache,
+                () => httpMessageHandler,
```
agr: This causes crashes when job reaches the end of the catalog, because base collector disposes of HttpClient at that point, which in turn disposes of MessageHandler.

src/Catalog/Helpers/Retry.cs
agr: I tried pulling in Polly, but unfortunately it depends on netstandard pulling great deal of System.* libraries into the output directory and causes a lot of new warnings.

src/Catalog/Icons/CatalogLeafDataProcessor.cs
```diff
@@ -0,0 +1,274 @@
...
+            if (!IsValidIconUrl(iconUrl))
+            {
+                _logger.LogInformation("Invalid icon URL {IconUrl}", iconUrl);
+                return;
+            }
+            var cachedResult = _iconCopyResultCache.Get(iconUrl);
+            if (cachedResult != null)
+            {
+                if (cachedResult.IsCopySucceeded)
+                {
```
loic-sharma: This method is quite complex. Consider splitting copying from cached result and copying externally:

```
if (cachedResult != null && await TryCopyFromCachedResultAsync(cachedResult))
{
   return;
}

// Copy externally..
```

src/Catalog/Icons/CatalogLeafDataProcessor.cs
```diff
@@ -0,0 +1,274 @@
...
+                            async () => await destinationStorage.CopyAsync(storageUrl, destinationStorage, destinationUrl, null, cancellationToken),
+                            e => { _logger.LogWarning(0, e, "Exception while copying from cache {StorageUrl}", storageUrl); return true; },
+                            MaxBlobStorageCopyAttempts,
+                            initialWaitInterval: TimeSpan.FromSeconds(5),
+                            waitIncrement: TimeSpan.FromSeconds(1));
+                    }
+                    catch (Exception e)
+                    {
+                        _logger.LogWarning(0, e, "Copy from cache failed after {NumRetries} attempts. Falling back to copy from external URL. {StorageUrl}",
+                            MaxBlobStorageCopyAttempts,
```
loic-sharma: This case seems unlikely, does this happen? Just curious

src/Catalog/Icons/HttpResponseMessageProvider.cs
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace NuGet.Services.Metadata.Catalog.Icons
+{
+    public class HttpResponseMessageProvider : IHttpResponseMessageProvider
```
loic-sharma: >IHttpResponseMessageProvider [](start = 47, length = 28)

Thoughts on naming these `IHttpClient` and `HttpClientWrapper`? A little shorter

src/Catalog/Icons/CatalogLeafDataProcessor.cs
agr: There is no proper cache cleanup logic when package is deleted, so we get here when there was a package delete that removes the icon. I haven't seen any other failures that got us here.

src/Catalog/Icons/HttpResponseMessageProvider.cs
agr: That would assume some sort of feature parity with `HttpClient`, which it absolutely does not have.

loic-sharma: Yes, we're cutting corners by not adding all `HttpClient` methods to `IHttpClient`, but does it matter? Is there a reason why one couldn't add missing `HttpClient` methods to `HttpResponseMessageProvider` in the future?

We use this pattern already for blob storage ([example](https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery.Core/Services/CloudBlobContainerWrapper.cs)).

src/Ng/Catalog2icon.nuspec
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0"?>
+<package >
+  <metadata>
+    <id>Catalog2icon</id>
```
joelverhagen: nit: Catalog2Icon

src/Catalog/Icons/CatalogLeafDataProcessor.cs
```diff
@@ -0,0 +1,278 @@
...
+                    iconUrl,
+                    item.PackageIdentity.Id,
+                    item.PackageIdentity.Version);
+            }
+            return true;
+        }
+
+        private async Task ProcessEmbeddedIconAsync(IStorage destinationStorage, CatalogCommitItem item, string iconFile, CancellationToken cancellationToken)
+        {
+            var packageFilename = PackageUtility.GetPackageFileName(item.PackageIdentity.Id, item.PackageIdentity.Version.ToNormalizedString()).ToLower();
```
joelverhagen: >ToLower [](start = 144, length = 7)

ToLowerInvariant

src/Catalog/Icons/IconProcessor.cs
```diff
@@ -0,0 +1,274 @@
...
+            }
+
+            var contentType = DetermineContentType(iconData, onlyGallerySupported: false);
+            if (string.IsNullOrWhiteSpace(contentType))
+            {
+                _logger.LogInformation("Failed to determine image type.");
+                return null;
+            }
+            _logger.LogInformation("Content type for {PackageId} {PackageVersion} {ContentType}", packageId, normalizedPackageVersion, contentType);
+            var content = new ByteArrayStorageContent(iconData, contentType, DefaultCacheControl);
```
joelverhagen: >DefaultCacheControl [](start = 77, length = 19)

Shouldn't this value be shared with Catalog2Dnx?

src/Catalog/Icons/IconsCollector.cs
```diff
@@ -0,0 +1,121 @@
...
+            JToken context,
+            DateTime commitTimeStamp,
+            bool isLastBatch,
+            CancellationToken cancellationToken)
+        {
+            await _iconCopyResultCache.InitializeAsync(cancellationToken);
+
+            var filteredItems = items
+                .GroupBy(i => i.PackageIdentity)                          // if we have multiple commits for the same package (id AND version)
+                .Select(g => g.OrderBy(i => i.CommitTimeStamp).ToList()); // group them together for processing in order
```
joelverhagen: Can you leave a comment about why you process all leafs per package? Why not just the latest? I think we talked offline about this but it is worth commenting on.

src/Ng/Jobs/Catalog2IconJob.cs
```diff
@@ -0,0 +1,97 @@
...
+            var source = arguments.GetOrThrow<string>(Arguments.Source);
+            var iconProcessor = new IconProcessor(TelemetryService, LoggerFactory.CreateLogger<IconProcessor>());
+            var httpHandlerFactory = CommandHelpers.GetHttpMessageHandlerFactory(TelemetryService, verbose);
+            var httpMessageHandler = httpHandlerFactory();
+            var httpClient = new HttpClient(httpMessageHandler);
+            var simpleHttpClient = new SimpleHttpClient(httpClient, LoggerFactory.CreateLogger<SimpleHttpClient>());
+            var catalogClient = new CatalogClient(simpleHttpClient, LoggerFactory.CreateLogger<CatalogClient>());
+            var httpResponseProvider = new HttpClientWrapper(httpClient);
+            var externalIconProvider = new ExternalIconContentProvider(httpResponseProvider, LoggerFactory.CreateLogger<ExternalIconContentProvider>());
+            var iconCopyResultCache = new IconCopyResultCache(auxStorageFactory.Create());
```
joelverhagen: Man I wish there was a technology to do all of this `new` for us.... :)

src/Catalog/Icons/IconCopyResultCache.cs
```diff
@@ -0,0 +1,87 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using NuGet.Services.Metadata.Catalog.Persistence;
+
+namespace NuGet.Services.Metadata.Catalog.Icons
+{
+    public class IconCopyResultCache : IIconCopyResultCache, IIconCopyResultCachePersistence
```
joelverhagen: UTs for this ?

src/Catalog/Icons/IconProcessor.cs
```diff
@@ -0,0 +1,274 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using NuGet.Common;
+using NuGet.Services.Metadata.Catalog.Persistence;
+
+namespace NuGet.Services.Metadata.Catalog.Icons
+{
+    public class IconProcessor : IIconProcessor
+    {
+        private const string DefaultCacheControl = "max-age=120";
```
joelverhagen: >DefaultCacheControl [](start = 29, length = 19)

Seems like a low cache time but I guess we can override it in CDN (or do we already??(

src/Catalog/Icons/IconsCollector.cs
```diff
@@ -0,0 +1,121 @@
...
+                        if (item.IsPackageDetails)
+                        {
+                            PackageDetailsCatalogLeaf leaf;
+                            try
+                            {
+                                leaf = await _catalogClient.GetPackageDetailsLeafAsync(item.Uri.AbsoluteUri);
+                            }
+                            catch (Exception e)
+                            {
+                                _logger.LogError(0, e, "Error while trying to retrieve catalog leaf {LeafUrl}", item.Uri.AbsoluteUri);
```
joelverhagen: >LogError [](start = 40, length = 8)

Won't this log twice? One here and one at the entry point? If we do this how many times does the exception appear in:

- traces
- exceptions

src/Catalog/Icons/IconProcessor.cs
```diff
@@ -0,0 +1,274 @@
...
+            {
+                return "image/png";
+            }
+
+            if (ArrayStartsWith(imageData, JpegHeader))
+            {
+                return "image/jpeg";
+            }
+
+            if (onlyGallerySupported)
```
joelverhagen: >onlyGallerySupported [](start = 16, length = 20)

Does this become application/octet-stream eventually? Is this consistent with catalog2dnx

src/Catalog/Icons/IconProcessor.cs
```diff
@@ -0,0 +1,274 @@
...
+            string normalizedPackageVersion)
+        {
+            var iconPath = PathUtility.StripLeadingDirectorySeparators(iconFilename);
+            var destinationUri = destinationStorage.ResolveUri(destinationStoragePath);
+
+            await ExtractAndStoreIconAsync(packageStream, iconPath, destinationStorage, destinationUri, cancellationToken, packageId, normalizedPackageVersion);
+            return destinationUri;
+        }
+
+        private async Task ExtractAndStoreIconAsync(
```
joelverhagen: >ExtractAndStoreIconAsync [](start = 27, length = 24)

The fact that this does not share code with Catalog2Dnx is concerning. Couldn't the implementations diverge easily?

src/Ng/Jobs/Catalog2IconJob.cs
agr: We need a base ng job imbued with such futuristic tech. Until then: out of scope :)

src/Catalog/Icons/IconProcessor.cs
agr: We use the same for nuspec and nupkg in flat container.

src/Catalog/Icons/IconsCollector.cs
agr: Yes, it will. But this one has proper logging scope and leaf URL in the message. The one at the entry point is utterly useless for this type of failure as it has none and we have 100 parallel tasks running, so even guessing the URL from logs does not work.

src/Catalog/Icons/IconProcessor.cs
agr: I want to pull this code to catalog2dnx eventually, so we have proper content type everywhere.

src/Catalog/Icons/IconProcessor.cs
agr: Added issue to track converging those: https://github.com/NuGet/Engineering/issues/2748

src/Catalog/Icons/IconCopyResultCache.cs
agr: added

src/Catalog/Icons/IconCopyResultCache.cs
```diff
@@ -37,29 +37,34 @@ public async Task InitializeAsync(CancellationToken cancellationToken)
...
+                _externalIconCopyResults = new ConcurrentDictionary<Uri, ExternalIconCopyResult>(dictionary);
             }
         }
 
         public async Task SaveAsync(CancellationToken cancellationToken)
         {
             var cacheUrl = _auxStorage.ResolveUri(CacheFilename);
-            var serialized = JsonConvert.SerializeObject(ExternalIconCopyResults);
+            var serialized = JsonConvert.SerializeObject(_externalIconCopyResults);
             var content = new StringStorageContent(serialized, contentType: "text/json");
```
joelverhagen: application/json

PR #658, opened 2019-09-18
Add search endpoint monitoring to ensure that a package version is in search

joelverhagen:
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/657.
Progress on https://github.com/NuGet/NuGetGallery/issues/7227.

### DIFF
--- a/src/Ng/Arguments.cs
+++ b/src/Ng/Arguments.cs
@@ -144,6 +144,20 @@ namespace Ng
         public const string FlatContainerCursorUri = "flatContainerCursorUri";
 
         /// <summary>
+        /// The argument prefix for the cursor of a <see cref="SearchEndpoint"/>. There are multiple search endpoints
+        /// so a parameter matching this prefix represents the cursor for a single instance. The suffix (after the prefix)
+        /// is the instance identifier, e.g. "usnc-a".
+        /// </summary>
+        public const string SearchCursorUriPrefix = "searchCursorUri-";
+
+        /// <summary>
+        /// The argument prefix for the base URL of a <see cref="SearchEndpoint"/>. There should be the same number of
+        /// <see cref="SearchBaseUriPrefix"/> parameters passed as <see cref="SearchCursorUriPrefix"/> with the same
+        /// set of suffixes.
+        /// </summary>
+        public const string SearchBaseUriPrefix = "searchBaseUri-";
+
+        /// <summary>
         /// The folder in which <see cref="PackageMonitoringStatus"/>es are saved by the <see cref="PackageMonitoringStatusService"/>.
         /// Defaults to <see cref="PackageStatusFolderDefault"/>.
         /// </summary>
--- a/src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/Search/SearchHasVersionValidator.cs
+++ b/src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/Search/SearchHasVersionValidator.cs
@@ -1,0 +1,76 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using NuGet.Protocol;
+using NuGet.Protocol.Core.Types;
+
+namespace NuGet.Services.Metadata.Catalog.Monitoring
+{
+    public class SearchHasVersionValidator : Validator<SearchEndpoint>
+    {
+        public SearchHasVersionValidator(
+            SearchEndpoint endpoint,
+            ValidatorConfiguration config,
+            ILogger<Validator> logger)
+            : base(endpoint, config, logger)
+        {
+        }
+
+        protected override async Task RunInternalAsync(ValidationContext context)
+        {
+            var searchVisible = await IsVisibleInSearchAsync(context);
+            var databaseState = await GetDatabaseStateAsync(context);
+            var databaseVisible = databaseState == DatabaseState.Listed;
+            if (databaseVisible != searchVisible)
+            {
+                const string listedString = "listed";
+                const string unlistedString = "unlisted";
+
+                throw new MetadataInconsistencyException(
+                    $"Database shows {databaseState.ToString().ToLowerInvariant()}" +
+                    $" but search shows {(searchVisible ? listedString : unlistedString)}.");
+            }
+        }
+
+        private async Task<bool> IsVisibleInSearchAsync(ValidationContext context)
+        {
+            var searchPage = await context.GetSearchPageForIdAsync(Endpoint.BaseUri);
+            var searchItem = searchPage.SingleOrDefault();
+            bool searchListed;
+            if (searchItem != null)
+            {
+                var searchVersions = await searchItem.GetVersionsAsync();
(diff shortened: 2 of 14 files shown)


src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/Search/SearchHasVersionValidator.cs
```diff
@@ -0,0 +1,100 @@
...
+            {
+                searchListed = false;
+            }
+
+            return searchListed;
+        }
+
+        private async Task<IEnumerable<IPackageSearchMetadata>> GetSearchPageForIdAsync(ValidationContext context)
+        {
+            return await (Task<IEnumerable<IPackageSearchMetadata>>)context.Cache.GetOrAdd(
```
scottbommarito: I'm not a huge fan of the free-form "string to object" thing. Having it relatively unstructured like this is both harder to use and opens it up to confusing keys in the future.

Specifically, I'd rather see a "base uri to query endpoint to results" cache just for search results, e.g. `IDictionary<Uri, IDictionary<string, Task<IEnumerable<IPackageSearchMetadata>>>>` or at least a `IDictionary<Uri, Task<IEnumerable<IPackageSearchMetadata>>>` if you are opposed to the multi-level nesting. This could allow you to share a lot of the logic necessary to construct queries across all of the search validations.

For example, you could add something like this to `ValidationContext`:
```
private IDictionary<Uri, IDictionary<string, Task<IEnumerable<IPackageSearchMetadata>>>> _searchQueryCache = new Dictionary<Uri, IDictionary<string, Task<IEnumerable<IPackageSearchMetadata>>>>();

public Task<IEnumerable<IPackageSearchMetadata>> GetPackageFromSearch(Uri baseUri, string endpoint)
{
    if (_searchQueryCache.TryGetValue(baseUri, out var queryCache) && queryCache.TryGetValue(query, out var result))
    {
        return result;
    }
    
    var queryUri = new Uri(Endpoint.BaseUri, endpoint);
    Logger.LogInformation(
        "Searching for package {Id} on {Uri}.",
        Package.Id,
        queryUri.AbsoluteUri);

    var rawSearchResource = new RawSearchResourceV3(context.HttpSourceResource.HttpSource, new[] { queryUri });
    var packageSearchResource = new PackageSearchResourceV3(rawSearchResource);
    return packageSearchResource.SearchAsync(
        searchTerm: $"packageid:{Package.Id}",
        filter: new SearchFilter(includePrerelease: true),
        skip: 0,
        take: 1,
        log: CommonLogger,
        cancellationToken: CancellationToken);
}
```

src/Ng/Arguments.cs
```diff
@@ -143,6 +143,20 @@ public static class Arguments
         /// </summary>
         public const string FlatContainerCursorUri = "flatContainerCursorUri";
 
+        /// <summary>
+        /// The argument prefix for the cursor of a <see cref="SearchEndpoint"/>. There are multiple search endpoints
+        /// so a parameter matching this prefix represents the cursor for a single instance. The suffix (after the prefix)
+        /// is the instance identifier, e.g. "usnc-a".
+        /// </summary>
+        public const string SearchCursorUriPrefix = "searchCursorUri-";
```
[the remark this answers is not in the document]
joelverhagen: Switching to JSON config is out of scope.

I think a semi-colon separated string is less readable than multiple parameters.

What alternative do you suggest?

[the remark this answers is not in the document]
scottbommarito: I think prefix is the best unless you switch to JSON config. This is fine, but I'd rather have JSON config.

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/Search/SearchHasVersionValidator.cs
joelverhagen: Okay, I was thinking the knowledge should live in the validator but I can do this instead.

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/ValidationContext.cs
```diff
@@ -77,37 +86,57 @@ public class ValidationContext
...
+                    var queryUri = new Uri(baseUri, "/query");
+                    _logger.LogInformation(
+                        "Searching for package {Id} on {Uri}.",
+                        Package.Id,
+                        queryUri.AbsoluteUri);
+
+                    var rawSearchResource = new RawSearchResourceV3(_httpSourceResource.HttpSource, new[] { queryUri });
+                    var packageSearchResource = new PackageSearchResourceV3(rawSearchResource);
+                    return packageSearchResource.SearchAsync(
+                        searchTerm: $"packageid:{Package.Id}",
```
loic-sharma: This may kick of the search request multiple times. Should this be wrapped by a `Lazy` to mitigate this?

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/ValidationContext.cs
```diff
@@ -77,37 +86,57 @@ public class ValidationContext
...
+                () => _databasePackageRegistrationMetadataResource.GetLeafAsync(Package, _commonLogger, CancellationToken));
             _v3Leaf = new Lazy<Task<PackageRegistrationLeafMetadata>>(
-                () => _v3PackageRegistrationMetadataResource.GetLeafAsync(Package, commonLogger, CancellationToken));
+                () => _v3PackageRegistrationMetadataResource.GetLeafAsync(Package, _commonLogger, CancellationToken));
 
             _timestampMetadataDatabase = new Lazy<Task<PackageTimestampMetadata>>(
                 () => _databasetimestampMetadataResource.GetAsync(this));
         }
 
+        public async Task<IEnumerable<IPackageSearchMetadata>> GetSearchPageForIdAsync(Uri baseUri)
```
loic-sharma: >baseUri [](start = 91, length = 7)

It's a tiny bit unclear what `baseUri` is. What do you think of a name like `searchBaseUri`?

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/Search/SearchHasVersionValidator.cs
```diff
@@ -0,0 +1,76 @@
...
+                    $"Database shows {databaseState.ToString().ToLowerInvariant()}" +
+                    $" but search shows {(searchVisible ? listedString : unlistedString)}.");
+            }
+        }
+
+        private async Task<bool> IsVisibleInSearchAsync(ValidationContext context)
+        {
+            var searchPage = await context.GetSearchPageForIdAsync(Endpoint.BaseUri);
+            var searchItem = searchPage.FirstOrDefault();
+            bool searchListed;
```
loic-sharma: Nitpick: do you think it's worth checking that there's only a single result?

joelverhagen: I sent `take=1` so I think `SingleOrDefault` will be enough.

src/NuGet.Services.Metadata.Catalog.Monitoring/Validation/Test/ValidationContext.cs
joelverhagen: Good catch. TIL.

PR #668, opened 2019-10-01
Support autocomplete on tokenized package ID

joelverhagen:
There are some top autocomplete queries that match the beginning of tokens but don't match the first token. Examples:
`entity` -> Microsoft.EntityFrameworkCore
`sqlite` -> Microsoft.Data.Sqlite
`http` -> System.Net.Http

Take a look at the autocomplete relevancy tests I added for more examples.

In short, I looked at the top 600 autocomplete queries with of length greater than 1 and looked for patterns. By in large, most queries start at the beginning of the package ID which the existing Azure Search implementation covered. However, there were several (e.g. "entityfra" or "aspnet") where the most popular matches don't have this as a prefix on the whole ID but rather a prefix of a token. This reality is what motivated the change.

Also note that I added the exact match mega boost for ALL packages IDs. The rationale here is that the user is certainly searching for a package ID (not a tag or text in the description) so a search like "JSON" is certainly looking for a package with "JSON" in the ID. To allow any package to be install-able, we don't just mega boost exact matches with symbols in search. Additionally, the `packageid:` workaround (or any field scoped term) is not supported on the autocomplete endpoint so there is no workaround.

Offline validation of this change is the added autocomplete relevancy tests. Online validation will be observing no major change in weekly query volume. In short, we expect no better or worse than parity.

Progress on https://github.com/NuGet/NuGetGallery/issues/7190.

### DIFF
--- a/tests/NuGet.Services.AzureSearch.FunctionalTests/Relevancy/AutocompleteRelevancyFunctionalTests.cs
+++ b/tests/NuGet.Services.AzureSearch.FunctionalTests/Relevancy/AutocompleteRelevancyFunctionalTests.cs
@@ -1,0 +1,74 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace NuGet.Services.AzureSearch.FunctionalTests
+{
+    public class AutocompleteRelevancyFunctionalTests : NuGetSearchFunctionalTestBase
+    {
+        public AutocompleteRelevancyFunctionalTests(CommonFixture fixture, ITestOutputHelper testOutputHelper)
+            : base(fixture, testOutputHelper)
+        {
+        }
+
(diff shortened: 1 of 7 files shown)


tests/NuGet.Services.AzureSearch.FunctionalTests/Relevancy/AutocompleteRelevancyFunctionalTests.cs
```diff
@@ -0,0 +1,73 @@
...
+            }
+        }
+
+        public static IEnumerable<object[]> EnsureFirstResultsData()
+        {
+            yield return new object[] { "aws", new[] { "awssdk.core" } };
+            yield return new object[] { "log4", new[] { "log4net" } };
+            yield return new object[] { "dap", new[] { "dapper" } };
+            yield return new object[] { "json", new[] { "json" } };
+            yield return new object[] { "entityframeworkcore.relational", new[] { "microsoft.entityframeworkcore.relational" } };
```
loic-sharma: If I search for `jso` is `newtonsoft.json` the top result?

joelverhagen: Yes. I added that test now.

PR #669, opened 2019-10-01
Fix for icon ingestor to store icon cache explitcitly in its own storage.

agr:
The idea to use destination storage as the cache at the same time was wrong because we allowed editing metadata for a long time.

### DIFF
--- a/src/Catalog/Icons/IconCopyResultCache.cs
+++ b/src/Catalog/Icons/IconCopyResultCache.cs
@@ -74,7 +74,7 @@ namespace NuGet.Services.Metadata.Catalog.Icons
             return null;
         }
 
-        public async Task<Uri> SaveExternalIcon(Uri originalIconUrl, Uri storageUrl, IStorage cacheStorage, CancellationToken cancellationToken)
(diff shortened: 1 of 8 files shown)


joelverhagen:
We should probably stop the job right now, right? We auto-deploy to China 24 hours after global we we want to make sure the timing is okay there.

src/Catalog/Icons/IconCopyResultCache.cs
```diff
@@ -85,32 +88,63 @@ public async Task<Uri> SaveExternalIcon(Uri originalIconUrl, Uri storageUrl, ISt
                 throw new InvalidOperationException("Object was not initialized");
             }
 
-            if (_externalIconCopyResults.TryGetValue(originalIconUrl, out var copyResult))
+            var uriSemaphore = GetUriSemaphore(originalIconUrl);
+            // Attempting to copy to the same location from multiple sources at the same time will throw,
+            // so we'll guard the copy attempt with semaphore.
+            // We'll guard the whole operation so we wouln't even try to copy items to cache more than once.
+            await uriSemaphore.WaitAsync(cancellationToken);
```
joelverhagen: Log before waiting.

PR #672, opened 2019-10-04
ExistsAsync in IStorage

agr:
Currently, `Exists` (non-async) only exists in `Storage`. So if one wants to call it in their code, they have to use `Storage` instead of `IStorage` and then suffer with tests.
This change adds async version to the interface and implementation, leaving existing as is.

### DIFF
--- a/src/Catalog/Registration/RecordingStorage.cs
+++ b/src/Catalog/Registration/RecordingStorage.cs
@@ -81,5 +81,10 @@ namespace NuGet.Services.Metadata.Catalog.Registration
         {
             return _innerStorage.ListAsync(cancellationToken);
         }
+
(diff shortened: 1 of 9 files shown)


src/Catalog/Registration/RecordingStorage.cs
```diff
@@ -81,5 +81,10 @@ public Task<IEnumerable<StorageListItem>> ListAsync(CancellationToken cancellati
         {
             return _innerStorage.ListAsync(cancellationToken);
         }
+
+        public Task<bool> ExistsAsync(string fileName, CancellationToken cancellationToken)
+        {
+            throw new NotImplementedException();
```
cristinamanum: Why not calling into _innerStorage.ExistsAsync?

agr:
Worked around the need to have that method. Will close for now.

PR #674, opened 2019-10-04
Cleaning up icon on ingestion failure

agr:
Back in time, when it was allowed to edit package metadata, people used to update the icon URL among other things.
Those updates might introduce the Icon URL that is unavailable now while the initial URL is still available.
When we encounter an update that uses unavailable URL we have to delete the icon we might have in storage to make the storage correctly reflect what users actually observe.

It also introduces an update to `IconProcessor.DeleteIconAsync` so it does not call `Storage.Exists` anymore, so we only send one request to process deletion instead of two.


src/Catalog/Icons/IconProcessor.cs
```diff
@@ -59,22 +60,28 @@ public class IconProcessor : IIconProcessor
...
             CancellationToken cancellationToken,
             string packageId,
             string normalizedPackageVersion)
         {
             _logger.LogInformation("Deleting icon blob {IconPath}", destinationStoragePath);
-            if (destinationStorage.Exists(destinationStoragePath))
+            var iconUri = new Uri(destinationStorage.BaseAddress, destinationStoragePath);
+            try
+            {
+                await destinationStorage.DeleteAsync(iconUri, cancellationToken, new DeleteRequestOptionsWithAccessCondition(AccessCondition.GenerateIfExistsCondition()));
```
joelverhagen: When would the icon not exist? What do you expect to happen here if the icon does not exist? An exception to be thrown? The no-op to occur? IIRC if you specify `If-Match: *` and no blob is there a precondition failure exception will be thrown. Might want to confirm.

agr: This will happen, for example, when we see the first package details leaf for some package identity with icon url specified and we fail to retrieve that URL. Service does not track which package exist or not, so we need to clean up unconditionally in this case.

If icon does not exist, the call no-ops, I had it tested.

PR #675, opened 2019-10-08
Increase Auxiliary2AzureSearch max download count decrease safety valve

joelverhagen:
There 10,603 decreases in PROD right now related to the stats pipeline outage. Right now the job halts if there are 5,000 or more decreases. I made this max configurable and increases the default to 15,000.

Address https://github.com/NuGet/NuGetGallery/issues/7610

### DIFF
--- a/src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/DownloadSetComparer.cs
+++ b/src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/DownloadSetComparer.cs
@@ -6,26 +6,24 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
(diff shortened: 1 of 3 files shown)


src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/DownloadSetComparer.cs
```diff
@@ -69,6 +67,16 @@ public class DownloadSetComparer : IDownloadSetComparer
 
             _logger.LogInformation("There are {Count} package IDs with download count changes.", result.Count);
 
+            if (decreaseCount > 0)
```
scottbommarito: IMO we should log this number regardless of whether or not it's more than zero. If we wanted to graph the number of decreases over time, not having the data points for 0 could lead to an incomplete dataset.

joelverhagen: Sounds good.

PR #681, opened 2019-10-22
[Azure Search] Boost results that match all tokenized unscoped terms

loic-sharma:
This feature is also known as "dot split". This generates Azure Search text that boost results that match all unscoped terms, after tokenization. This greatly improves queries that aren't an exact match. For example:

Query | Expected Results | Old Rank | New rank
------- | ------------------ | ---------- | -----------
text.ui | Microsoft.VisualStudio.Text.UI | 10 | 1
windows.extensions | Microsoft.Windows.Extensions | >20 | 6
System.Drawing | System.Drawing.Common, System.Drawing.Common | 1, 8 | 1, 2
System.Net.Http.Formatting | Microsoft.AspNet.WebApi.Client, System.Net.Http.Formatting.Extensions | 5, 2 | 2, 1
System.Web | System.Web.Http.Common, System.Web.Providers, System.Web.Providers.Core | >20, >20, >20 | 16, 18, 20
socket.io | SocketIoClientDotNet, Socket.IO.Client | 13, 17 | 1, 2
system.web.mvc | System.Web.Mvc.Extensions | >20 | 4
microsoft.toolkit | Microsoft.Toolkit, Microsoft.Toolkit.* | 1, N/A | 1, 3-13, 17-19
Uno.Microsoft.Toolkit | Uno.Microsoft.Toolkit, Uno.Microsoft.Toolkit.* | 1, N/A | 1, 2-15
NuGet.Build | NuGet.Build, NuGet.Build.Tasks.Pack, NuGet.Build.Tasks, NuGet.Build.Packaging | 1, 18, 19, 20  | 1, 2, 3, 4
Pomelo.EntityFramework | Pomelo.EntityFrameworkCore.MySql | 1 | 4
EntityFramework.Sqlite | Microsoft.EntityFrameworkCore.Sqlite | 2 | 1
EntityFramework.Postgres | Npgsql.EntityFrameworkCore.PostgreSQL | 16 | 1
MicrosoftExpression | MicrosoftExpressionInteractions | >20 | 5
serilog.aspnet | Serilog.AspNetCore | > 20 | 1
dotnet-test | dotnet-test-nunit, dotnet-test-mstest | 11, 13 | 2, 3
baget.core | BaGet.Core, BaGet.Core.Service | 1, 11 | 1, 2

Verifications:

* ✔This improves the curated score from 0.8249 to 0.8251
* ✔This improves the feedback score from 0.7199 to 0.7315
* ✔This passes the relevancy functional tests

Alternative implementations that were considered: https://github.com/NuGet/NuGetGallery/issues/7391#issuecomment-543989727
Addresses https://github.com/nuget/nugetgallery/issues/7391

### DIFF
--- a/src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
+++ b/src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
@@ -17,9 +17,9 @@ namespace NuGet.Services.AzureSearch.SearchService
     {
         public const string MatchAllDocumentsQuery = "*";
         private static readonly char[] PackageIdSeparators = new[] { '.', '-', '_' };
-        private static readonly char[] TokenizationSeparators = new[] { '.', '-', '_', ',', ';' };
+        private static readonly char[] TokenizationSeparators = new[] { '.', '-', '_', ',' };
         private static readonly Regex TokenizePackageIdRegex = new Regex(
-            @"((?<=[a-z])(?=[A-Z])|((?<=[0-9])(?=[A-Za-z]))|((?<=[A-Za-z])(?=[0-9]))|[.\-_,;])",
+            @"((?<=[a-z])(?=[A-Z])|((?<=[0-9])(?=[A-Za-z]))|((?<=[A-Za-z])(?=[0-9]))|[.\-_,])",
             RegexOptions.None,
             matchTimeout: TimeSpan.FromSeconds(10));
 
--- a/src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
+++ b/src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
@@ -17,8 +17,9 @@ namespace NuGet.Services.AzureSearch.SearchService
     {
         public const string MatchAllDocumentsQuery = "*";
         private static readonly char[] PackageIdSeparators = new[] { '.', '-', '_' };
+        private static readonly char[] TokenizationSeparators = new[] { '.', '-', '_', ',', ';' };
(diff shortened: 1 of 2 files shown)


src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
```diff
@@ -16,6 +17,11 @@ public partial class SearchTextBuilder : ISearchTextBuilder
     {
         public const string MatchAllDocumentsQuery = "*";
         private static readonly char[] PackageIdSeparators = new[] { '.', '-', '_' };
+        private static readonly char[] TokenizationSeparators = new[] { '.', '-', '_', ',', ';' };
```
loic-sharma: I started by tokenizing on just package id separators, but I was able to improve the feedback score from  0.7285 to 0.7315 by adding `,`. I also added `;` arbitrarily, removing it doesn't affect the feedback score.

joelverhagen:
This should be merged into an `ab-*` branch, right?

src/NuGet.Services.AzureSearch/SearchService/SearchTextBuilder.cs
joelverhagen: There are other symbols that `PackageIdCustomAnalyzer` splits on. Why don't we do those ones too? How were `,` and `;` picked in addition to the ones that can be in package IDs?

loic-sharma: I went through data to see if including these characters could improve results. I found that almost all `;` were at the end of the query, which isn't helped by this change. As such, I've kept only `,`.

PR #687, opened 2019-11-05
Stop collector from fetching all pages when back cursor is before the end

joelverhagen:
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/686/files.

See the comment on `GetCommitsInRange` for details but the short version is that today the commit collector loads all pages after the front cursor even if the back cursor logically excludes them. Added unit tests for `CommitCollector` to cover this change and existing scenarios.


src/Catalog/CommitCollector.cs
```diff
@@ -119,12 +120,62 @@ public abstract class CommitCollector : CollectorBase
...
+
+        public static IEnumerable<CatalogCommit> GetCommitsInRange(
+            IEnumerable<CatalogCommit> commits,
+            DateTimeOffset minCommitTimestamp,
+            DateTimeOffset maxCommitTimestamp)
+        {
+            // Only consider pages that have a (latest) commit timestamp greater than the minimum bound. If a page has
+            // a commit timestamp greater than the minimum bound, then there is at least one item with a commit
+            // timestamp greater than the minimum bound. Sort the pages by commit timestamp so that they are
+            // chronological order.
```
loic-sharma:
```suggestion
            // in chronological order.
```

src/NuGet.Protocol.Catalog/Models/CatalogPageContext.cs
```diff
@@ -0,0 +1,34 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Newtonsoft.Json;
+
+namespace NuGet.Protocol.Catalog
+{
+    public class CatalogPageContext
```
loic-sharma: Just curious, why are you adding this? Does anything need to read/write this?

joelverhagen: I added it so I can use the catalog models in tests (for easy serialization). This product code needs the context:
https://github.com/NuGet/NuGet.Services.Metadata/blob/e921a7eef360ee6bcc11b7ca9f94d96e93193a1c/src/Catalog/CatalogCommitItem.cs#L69-L74

PR #690, opened 2019-11-06
Make Lightning case insensitive for package ID

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/7680

Verified by running `prepare` against PROD and looking at the `nspec` package ID. The latest 11 versions are "NSpec". The older ones are "nspec". They should all be in one batch, not two.


loic-sharma:
Out of curiosity, how did you catch this?

joelverhagen:
> Out of curiosity, how did you catch this?

Noticed an inconsistency from version to version within a single registration index generated by lightning.

PR #691, opened 2019-11-06
Bring Registration and Catalog models up to match nuget.org

joelverhagen:
Bring the Catalog and Registration models up to date with what nuget.org has. These are intentionally a superset of the documented protocol since they are used by nuget.org for nuget.org things. In the future if we want to make these packages part of an SDK we could separate the models into "documented" and "nuget.org-specific" classes where the latter inherits the former.

These are necessary for a job to write registration in such a way that is consistent with the current implementation (which uses RDF and JSON-LD).

As an added benefit, it means more information is available to people reading the catalog, e.g. an update to Azure Search that would look at deprecation information.

On its own, this PR has no functional impact on our jobs.

### DIFF
--- a/src/NuGet.Protocol.Catalog/Models/AlternatePackage.cs
+++ b/src/NuGet.Protocol.Catalog/Models/AlternatePackage.cs
@@ -2,11 +2,13 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Newtonsoft.Json;
-using NuGet.Protocol.Catalog.Serialization;
 
 namespace NuGet.Protocol.Catalog
 {
-    public class PackageDependency
+    /// <summary>
+    /// Source: https://docs.microsoft.com/en-us/nuget/api/registration-base-url-resource#alternate-package
+    /// </summary>
+    public class AlternatePackage
     {
         [JsonProperty("@id")]
         public string Url { get; set; }
@@ -18,7 +20,6 @@ namespace NuGet.Protocol.Catalog
         public string Id { get; set; }
 
         [JsonProperty("range")]
-        [JsonConverter(typeof(PackageDependencyRangeConverter))]
         public string Range { get; set; }
     }
 }
--- a/src/NuGet.Protocol.Catalog/Models/BasePackageDependencyGroup.cs
+++ b/src/NuGet.Protocol.Catalog/Models/BasePackageDependencyGroup.cs
@@ -1,12 +1,12 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
 using Newtonsoft.Json;
-using NuGet.Protocol.Catalog.Serialization;
 
 namespace NuGet.Protocol.Catalog
 {
-    public class PackageDependency
+    public abstract class BasePackageDependencyGroup<TDependency> where TDependency : PackageDependency
     {
         [JsonProperty("@id")]
         public string Url { get; set; }
@@ -14,11 +14,10 @@ namespace NuGet.Protocol.Catalog
         [JsonProperty("@type")]
         public string Type { get; set; }
 
-        [JsonProperty("id")]
-        public string Id { get; set; }
+        [JsonProperty("dependencies")]
+        public List<TDependency> Dependencies { get; set; }
 
-        [JsonProperty("range")]
-        [JsonConverter(typeof(PackageDependencyRangeConverter))]
-        public string Range { get; set; }
+        [JsonProperty("targetFramework")]
+        public string TargetFramework { get; set; }
     }
 }
--- a/src/NuGet.Protocol.Catalog/Models/PackageDeprecation.cs
+++ b/src/NuGet.Protocol.Catalog/Models/PackageDeprecation.cs
@@ -1,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace NuGet.Protocol.Catalog
+{
+    /// <summary>
+    /// Source: https://docs.microsoft.com/en-us/nuget/api/registration-base-url-resource#package-deprecation
+    /// </summary>
+    public class PackageDeprecation
+    {
+        [JsonProperty("@id")]
+        public string Url { get; set; }
+
+        [JsonProperty("@type")]
+        public string Type { get; set; }
+
+        [JsonProperty("alternatePackage")]
+        public AlternatePackage AlternatePackage { get; set; }
+
+        [JsonProperty("message")]
+        public string Message { get; set; }
+
+        [JsonProperty("reasons")]
+        public List<string> Reasons { get; set; }
+    }
+}
--- a/src/NuGet.Services.V3/Registration/Models/ICommitted.cs
+++ b/src/NuGet.Services.V3/Registration/Models/ICommitted.cs
@@ -1,9 +1,13 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-namespace NuGet.Protocol.Catalog
+using System;
+
+namespace NuGet.Protocol.Registration
 {
-    public class PackageDependencyGroup : BasePackageDependencyGroup<PackageDependency>
+    public interface ICommitted
     {
+        string CommitId { get; set; }
+        DateTimeOffset CommitTimestamp { get; set; }
     }
 }
--- a/src/NuGet.Services.V3/Registration/Models/RegistrationContainerContext.cs
+++ b/src/NuGet.Services.V3/Registration/Models/RegistrationContainerContext.cs
@@ -1,0 +1,59 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Newtonsoft.Json;
+using NuGet.Protocol.Catalog;
+
+namespace NuGet.Protocol.Registration
+{
+    public class RegistrationContainerContext
+    {
+        [JsonProperty("@vocab")]
+        public string Vocab { get; set; }
+
+        [JsonProperty("catalog")]
+        public string Catalog { get; set; }
+
+        [JsonProperty("xsd")]
+        public string Xsd { get; set; }
+
+        [JsonProperty("items")]
+        public ContextTypeDescription Items { get; set; }
+
+        [JsonProperty("commitTimeStamp")]
+        public ContextTypeDescription CommitTimestamp { get; set; }
+
+        [JsonProperty("commitId")]
+        public ContextTypeDescription CommitId { get; set; }
+
+        [JsonProperty("count")]
+        public ContextTypeDescription Count { get; set; }
+
+        [JsonProperty("parent")]
+        public ContextTypeDescription Parent { get; set; }
+
+        [JsonProperty("tags")]
+        public ContextTypeDescription Tags { get; set; }
+
+        [JsonProperty("reasons")]
+        public ContextTypeDescription Reasons { get; set; }
+
+        [JsonProperty("packageTargetFrameworks")]
+        public ContextTypeDescription PackageTargetFrameworks { get; set; }
+
+        [JsonProperty("dependencyGroups")]
+        public ContextTypeDescription DependencyGroups { get; set; }
+
+        [JsonProperty("dependencies")]
+        public ContextTypeDescription Dependencies { get; set; }
+
+        [JsonProperty("packageContent")]
+        public ContextTypeDescription PackageContent { get; set; }
+
+        [JsonProperty("published")]
+        public ContextTypeDescription Published { get; set; }
+
+        [JsonProperty("registration")]
+        public ContextTypeDescription Registration { get; set; }
+    }
+}
--- a/src/NuGet.Services.V3/Registration/Models/RegistrationPackageDependency.cs
+++ b/src/NuGet.Services.V3/Registration/Models/RegistrationPackageDependency.cs
@@ -1,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Newtonsoft.Json;
+using NuGet.Protocol.Catalog;
+
+namespace NuGet.Protocol.Registration
+{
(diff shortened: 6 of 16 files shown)


src/NuGet.Protocol.Catalog/Models/AlternatePackage.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Newtonsoft.Json;
+
+namespace NuGet.Protocol.Catalog
+{
+    public class AlternatePackage
```
scottbommarito: `AlternatePackage` was intended to be identical to `PackageDependency`, so you can replace this class with that.

If you would prefer not to do that, we should at least add `[JsonConverter(typeof(PackageDependencyRangeConverter))]` to `Range`.

joelverhagen: `PackageDependencyRangeConverter` is for quirks. Do we have quirks in `AlternatePackage`? My understanding was that gallery validated the package version range.

scottbommarito: Oh, there shouldn't be any quirks. Gallery doesn't accept any version range strings--it either supports a min version (`[1.0.0, )`) or any version (`(, )`). Db2Catalog creates these ranges itself.

It's possible that we may misalign `AlternatePackage` with `PackageDependency`, but I think there should be some link between them. Maybe a base class (`PackageReference`) that `AlternatePackage` and `PackageDependency` extend?

src/NuGet.Protocol.Catalog/Models/BasePackageDependencyGroup.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace NuGet.Protocol.Catalog
+{
+    public class BasePackageDependencyGroup<TDependency> where TDependency : PackageDependency
+    {
+        [JsonProperty("@id")]
```
scottbommarito: Another thought--all entities in SPARQL/JSON-LD/whatever have `@id` and `@type` properties. Should we make a shared interface or base class that all of these types extend, so that these fields don't have to be declared repeatedly?

src/NuGet.Protocol.Catalog/Models/AlternatePackage.cs
loic-sharma: Please add a link to the documentation. Also, it seems the documentation needs to be updated to describe the shape of the `alternatePackage` object

src/NuGet.Protocol.Catalog/Models/AlternatePackage.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        [JsonProperty("@id")]
+        public string Url { get; set; }
+
+        [JsonProperty("@type")]
+        public string Type { get; set; }
```
cristinamanum: What Type an AlternatePackage can have?

src/NuGet.Protocol.Catalog/Models/AlternatePackage.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        [JsonProperty("id")]
+        public string Id { get; set; }
+
+        [JsonProperty("range")]
+        public string Range { get; set; }
```
cristinamanum: Is this a VersionRange?

src/NuGet.Protocol.Catalog/Models/BasePackageDependencyGroup.cs
joelverhagen: I believe inheritance for the sole purpose of DRY is a bad pattern. I've seen inheritance used WAY too often as a method of code sharing where composition really should have been used. In my experience, I've found inheritance should be used a lot more sparingly than it is in practice.

In this PR, I only added inheritance in cases where there is benefit moving forward and the super-sub relationship is strong in the _ideal_ mental model for our entities. For the package dependency group in catalog vs. registration, this is clear subset/superset relationship. I don't even feel strongly about this one. I wouldn't feel too bad about cutting the relationship and introducing mapping code. For alternate package and package dependency, I don't feel as strongly but I think it's a bad idea to add inheritance just as a DRY convenience. What scenario would we be enabling by clubbing these together?

For these JSON models I am not convinced that inheritance actually makes our lives better. I think it's just as likely that we have unintended properties leaking to places we don't want them.

For `@id` and `@type` I don't think it's right to add a base class just so we can avoid defining these unfortunate properties multiple times. For most of the `@id` and `@type` cases, they are there as an artifact of our JSON-LD implementation not because we actually think they are useful or provide crucial new information. Consider the `@id` of a package dependency group or the `@type` of deprecations.

On a related note, I only added the `ICommitted` interface because it enables a single method to set the commit ID and commit timestamp for N different classes. I think this is a legitimate use of an interface but would be an inappropriate use for a base class.

If you want to throw all of my anti-OOP mumbo-jumbo out the window my point can be simplified to this: I wanted the most granular control possible for us to match the existing JSON shape in a very predictable way. Inheritance makes this harder.

src/NuGet.Protocol.Catalog/Models/AlternatePackage.cs
joelverhagen: It is a version range, yet.

src/NuGet.Protocol.Catalog/Models/BasePackageDependencyGroup.cs
loic-sharma: I wouldn't. `BaGet.Protocol` renames the `@id` property to more logical names like `CatalogLeafUrl` to describe the destination object. We can't do this if the property names are restricted by an interface

src/NuGet.Protocol.Catalog/Models/AlternatePackage.cs
scottbommarito: It's defined as "The package dependency that should be used instead", where "package dependency" is an object that's defined above. Perhaps it should have a link to the package dependency definition but I believe it is technically sufficient.

src/NuGet.Protocol.Catalog/Models/PackageDeprecation.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace NuGet.Protocol.Catalog
+{
+    public class PackageDeprecation
```
loic-sharma: Please add a link to the docs

src/NuGet.Services.V3/Registration/Models/RegistrationPackageDependency.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Newtonsoft.Json;
+using NuGet.Protocol.Catalog;
+
+namespace NuGet.Protocol.Registration
+{
+    public class RegistrationPackageDependency : PackageDependency
+    {
+        [JsonProperty("registration", Order = 1)]
```
loic-sharma: Just curious, what's the `Order = 1` for? Does it make the `Registration` property the first serialized property?

src/NuGet.Protocol.Catalog/Models/AlternatePackage.cs
joelverhagen: I have responded to your latest comment here: https://github.com/NuGet/NuGet.Services.Metadata/pull/691#discussion_r343277774.

src/NuGet.Services.V3/Registration/Models/RegistrationPackageDependency.cs
joelverhagen: It makes that property last, which is what the existing order on nuget.org has.

src/NuGet.Protocol.Catalog/Models/AlternatePackage.cs
cristinamanum: The PackageDependency and the AlternatePackage are very similar. Maybe add some comments what the difference between them? 
Do they need to be somehow in sync ? If yes I think that @scottbommarito 's suggestion is good.

src/NuGet.Protocol.Catalog/Models/BasePackageDependencyGroup.cs
scottbommarito: I don't think it's necessarily just a "DRY convenience"--they are both links to specific versions/ranges of a package. I can imagine some code that finds a package in registration from a `PackageDependency`, which could be reused for both finding dependency metadata and for finding alternate package metadata. The DRY aspect is just a bonus. Worst case scenario, if the types diverge down the line, we can split the types apart, which shouldn't be a very difficult change to make. I can agree that adding a base class is overdesigning, but the V3 docs have literally defined an alternate package as "The package dependency that should be used instead".

I would argue, regarding `@id` and `@type`, that the fact that they are "unfortunate" and generally unused is precisely the reason why we should put them on a single base class or interface. When we finally decide to fully get rid of them, we can remove their declaration from a single place and avoid updating N files. Even if we do have to update each file, for some other reason, it will be very clear to us which files will need to be updated.

I agree that too much OOP can slow development, but I think for these two changes we have a good reason to do the inheritance. Ultimately, however, I'm fine with this PR moving forward without them.

src/NuGet.Protocol.Catalog/Models/BasePackageDependencyGroup.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace NuGet.Protocol.Catalog
+{
+    public class BasePackageDependencyGroup<TDependency> where TDependency : PackageDependency
```
cristinamanum: Would it make sense to be abstract ?

src/NuGet.Protocol.Catalog/Models/BasePackageDependencyGroup.cs
joelverhagen: I didn't remember that an alternate package is a package dependency. I would have pushed back on that at the protocol stage if I caught it. I think from a domain perspective they are different things and adding properties to one won't always mean we want that property in the other. One critical difference is that a range can be `*` in alternate package but package dependencies cannot have `*`. They are different models IMO.

src/NuGet.Protocol.Catalog/Models/AlternatePackage.cs
loic-sharma: Ah. That's not evident, I would definitely include a link.

src/NuGet.Protocol.Catalog/Models/AlternatePackage.cs
joelverhagen: It's one of the strange RDF URLs that is the current URL plus a fragment like `#alternatePackage`. It is unused by the client and is there as a coincidence of JSON-LD.

src/NuGet.Protocol.Catalog/Models/BasePackageDependencyGroup.cs
joelverhagen: yes, you're right. It should be abstract.

src/NuGet.Services.V3/Registration/Models/RegistrationContainerContext.cs
```diff
@@ -0,0 +1,59 @@
...
+        public string Catalog { get; set; }
+
+        [JsonProperty("xsd")]
+        public string Xsd { get; set; }
+
+        [JsonProperty("items")]
+        public ContextTypeDescription Items { get; set; }
+
+        [JsonProperty("commitTimeStamp")]
+        public ContextTypeDescription CommitTimestamp { get; set; }
```
cristinamanum: Are all these properties supposed to be of type ContextTypeDescription?

src/NuGet.Protocol.Catalog/Models/AlternatePackage.cs
loic-sharma: https://github.com/NuGet/docs.microsoft.com-nuget/pull/1766

src/NuGet.Services.V3/Registration/Models/RegistrationContainerContext.cs
joelverhagen: Yes, this is all the `@context` goo at the bottom of our JSON files:
image

src/NuGet.Services.V3/Registration/Models/ICommitted.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace NuGet.Protocol.Registration
+{
+    public interface ICommitted
+    {
+        string CommitId { get; set; }
+        DateTimeOffset CommitTimestamp { get; set; }
```
cristinamanum: Why is the comittimestamp important? Is there an user scenario for it?

joelverhagen: There is no known user scenario around commit timestamp. That being said, it is currently written in today's blob's so having it in the model is required if we want to avoid any unintentional/unofficial protocol breaks.

src/NuGet.Protocol.Catalog/Models/AlternatePackage.cs
joelverhagen: Regarding renaming, I tried to make the models match the protocol as much as possible so that it would be easy to look at the protocol docs and know which property it is associated with in IntelliSense (which does not show `JsonProperty` values).

src/NuGet.Protocol.Catalog/Models/AlternatePackage.cs
joelverhagen: @cristinamanum, we have separated the notions of package dependency and alternate package.

src/NuGet.Protocol.Catalog/Models/BasePackageDependencyGroup.cs
joelverhagen: Done.

src/NuGet.Protocol.Catalog/Models/PackageDeprecation.cs
joelverhagen: Done.

src/NuGet.Protocol.Catalog/Models/AlternatePackage.cs
joelverhagen: Done.

PR #694, opened 2019-11-19
Cache expiration for icon copy failures

agr:
Addresses https://github.com/NuGet/Engineering/issues/2780.

Deploying this version will invalidate all the icon copy failures we had cached so far, so reflowing a package after a `failCacheTime` (or for any package processed before the deployment) will retry copying icon from the external source.


src/Catalog/Icons/IconCopyResultCache.cs
```diff
@@ -75,6 +78,10 @@ public ExternalIconCopyResult Get(Uri iconUrl)
 
             if (_externalIconCopyResults.TryGetValue(iconUrl, out var result))
             {
+                if (!result.IsCopySucceeded && (!result.Expiration.HasValue || result.Expiration.Value < DateTimeOffset.UtcNow))
+                {
```
loic-sharma: Minor suggestion: you could reduce nesting by inverting the `if` on line 79. I don't feel strongly about this though. Example:

```csharp
if (!_externalIconCopyResults.TryGetValue(iconUrl, out var result))
{
  return null;
}

if (!result.IsCopySucceeded && ...)
{
  return null;
}

return result;
```

src/Catalog/Icons/ExternalIconCopyResult.cs
```diff
@@ -13,20 +13,32 @@ public static ExternalIconCopyResult Success(Uri sourceUrl, Uri storageUrl)
             {
                 SourceUrl = sourceUrl,
                 StorageUrl = storageUrl,
+                Expiration = null,            // successes don't expire
             };
         }
 
-        public static ExternalIconCopyResult Fail(Uri sourceUrl)
+        public static ExternalIconCopyResult Fail(Uri sourceUrl, TimeSpan validityPeriod)
         {
```
zhhyu: Null check for the arguments?

PR #695, opened 2019-11-21
Align NuGet client packages to 5.0.0-preview1.5707

joelverhagen:
This version is already coming into the Azure Search jobs via its NuGet.Jobs dependency.
https://github.com/NuGet/NuGet.Jobs/blob/49d771844b56b3c2243d0dc78dcbc61321607b3c/src/Validation.Common.Job/Validation.Common.Job.csproj#L112

This aligns the versions and fixes one API change.

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -1,6 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
+    <clear />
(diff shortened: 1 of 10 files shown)


joelverhagen:
> Any chance we could get off the preview version?

Yes, but I would like to drive this from ServerCommon, NuGetGallery, or even NuGet.Jobs. NuGet.Services.Metadata is kind of the terminal node.

PR #696, opened 2019-11-21
Improve error message when an argument is not provided

joelverhagen:
`KeyNotFoundException` can be thrown because this method is used a lot:
https://github.com/NuGet/ServerCommon/blob/01f97075c87121e48abd428771cb4f758861d572/src/NuGet.Services.Configuration/DictionaryExtensions.cs#L38

It's better to show the error message here and helps local/ops execution of `ng.exe`.

### DIFF
--- a/src/Ng/Program.cs
+++ b/src/Ng/Program.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Net;
@@ -95,6 +96,12 @@ namespace Ng
             catch (ArgumentException ae)
             {
                 _logger?.LogError("A required argument was not found or was malformed/invalid: {Exception}", ae);
+
+                Console.WriteLine(job != null ? job.GetUsage() : NgJob.GetUsageBase());
+            }
+            catch (KeyNotFoundException knfe)
+            {
+                _logger?.LogError("An expected key was not found. One possible cause of this is required argument has not been provided: {Exception}", knfe);
 
                 Console.WriteLine(job != null ? job.GetUsage() : NgJob.GetUsageBase());
             }


src/Ng/Program.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
```
agr: Why is this needed?

src/Ng/Program.cs
```diff
@@ -92,7 +93,7 @@ public static async Task MainAsync(string[] args)
                 job = NgJobFactory.GetJob(jobName, telemetryService, loggerFactory);
                 await job.RunAsync(arguments, cancellationTokenSource.Token);
             }
-            catch (ArgumentException ae)
+            catch (Exception ae) when (ae is ArgumentException || ae is KeyNotFoundException)
             {
                 _logger?.LogError("A required argument was not found or was malformed/invalid: {Exception}", ae);
```
zhhyu: Maybe improve the log message to denote that the key is not found, or another catch block for this case.

src/Ng/Program.cs
joelverhagen: `KeyNotFoundException` is in this namespace:
https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.keynotfoundexception?view=netframework-4.8

src/Ng/Program.cs
joelverhagen: Good idea, I'll do that.

src/Ng/Program.cs
```diff
@@ -92,7 +93,7 @@ public static async Task MainAsync(string[] args)
                 job = NgJobFactory.GetJob(jobName, telemetryService, loggerFactory);
                 await job.RunAsync(arguments, cancellationTokenSource.Token);
             }
-            catch (ArgumentException ae)
+            catch (Exception ae) when (ae is ArgumentException || ae is KeyNotFoundException)
```
agr: In general, catching an `ArgumentException` here, and representing it as a "required argument id missing/invalid" is a bit of a stretch, since the `RunAsync` call is tried that runs complete job. `ArgumentException` can be thrown anywhere, not only at the argument validation level.
Ideally, we'd probably want a separate command line argument validation/consumption stage separate from running the job, but I'd imagine that would be quite out of scope for the intended change.

joelverhagen: Yup, I didn't want to fix the questionable design of the command line parsing -- I just wanted to make the current mess more usable.

PR #699, opened 2019-12-03
Move in-memory cloud storage to V3 tests

joelverhagen:
These are stub implementations of the blob storage interfaces so that writing integration tests is easy. I have used these with great success in testing the Azure Search jobs.


tests/NuGet.Services.V3.Tests/Support/InMemoryCloudBlobClient.cs
```diff
@@ -5,13 +5,13 @@
...
 
-namespace NuGet.Services.AzureSearch.Catalog2AzureSearch.Integration
+namespace NuGet.Services
 {
     public class InMemoryCloudBlobClient : ICloudBlobClient
     {
         private readonly object _lock = new object();
 
-        public Dictionary<string, InMemoryCloudBlobContainer> Containers { get; } = new Dictionary<string, InMemoryCloudBlobContainer>();
+        public SortedDictionary<string, InMemoryCloudBlobContainer> Containers { get; } = new SortedDictionary<string, InMemoryCloudBlobContainer>();
```
loic-sharma: Why `SortedDictionary`? I'm just curious, LGTM 😊

joelverhagen: Makes assertions of expected keys easier. Otherwise you have to sort the "actual" as well as hard coded sorted "expected".

PR #700, opened 2019-12-04
[Hotfix] Add missing reference in functional test solution

joelverhagen:
This project is not built during the CI so I missed it.

The break occurred here:
https://github.com/NuGet/NuGet.Services.Metadata/commit/b061be9a8ed9ffe0fadf6fbe005b7d6aa22129a2#diff-cb948c5b0d471668e46787a0fdfed3c8L68-L70

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -100,6 +100,12 @@ Invoke-BuildStep 'Building solution' {
     } `
     -ev +BuildErrors
 
+Invoke-BuildStep 'Building functional test solution' { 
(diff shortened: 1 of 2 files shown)


build.ps1
```diff
@@ -100,6 +100,12 @@ Invoke-BuildStep 'Building solution' {
     } `
     -ev +BuildErrors
 
+Invoke-BuildStep 'Building functional test solution' {
```
joelverhagen: This will make sure this sort of problem does not happen again.

PR #702, opened 2019-12-04
[PackageType] Add PackageType to catalog

ryuyu:
Adds packageType information in to catalog.


tests/NgTests/Data/TestCatalogEntries.resx
```diff
@@ -850,6 +866,10 @@
       "@id": "packageEntry",
       "@container": "@set"
     },
+		"packageTypes": {
```
joelverhagen: Indentation is inconsistent here

src/Catalog/xslt/nuspec.xslt
```diff
@@ -290,39 +334,39 @@
...
-                </ng:assembly>
-                <ng:targetFramework>
-                  <xsl:value-of select="." />
-                </ng:targetFramework>
-              </rdf:Description>
-            </ng:frameworkAssemblyGroup>
-          </xsl:for-each>
-        </xsl:when>
-        <xsl:otherwise>
+    <xsl:choose>
```
joelverhagen: nit: Can we revert these indentation changes? makes the diff noisy

tests/CatalogTests/PackageCatalogItemTests.cs
```diff
@@ -27,6 +27,11 @@ public class PackageCatalogItemTests
         [InlineData("InvalidDependencyVersionRange.0.1.0")] //A dependency with a version range that is invalid
         [InlineData("MissingDependencyVersionRange.0.1.0")] // A dependency with no version range attribute
         [InlineData("WhitespaceDependencyVersionRange.0.1.0")] // A dependency with a version range that is whitespace
+        [InlineData("PackageTypeCollapseDuplicate")]
```
joelverhagen: Same type different version?
Duplicate type name with different case?
Same type different version by System.Version but not SemVer (1.0, 1.0.0, 1.0.0.0)?
Whitespace around name? Whitepace around version?

It sucks we have to have the full nupkg and JSON data for each of these test variants. That's a lot of noise for small changes between tests.

tests/NgTests/Data/TestCatalogEntries.resx
ryuyu: Corrected all tabs to spaces. Look consistent for me now. Hopefully will translate correctly here.

src/Catalog/xslt/nuspec.xslt
ryuyu: I used VS auto format here. It's a bit noisy now, but it's at least internally consistent. Hopefully, we can just use autoformat for this doc any time we need to make changes going forward.

PR #706, opened 2019-12-06
Add configuration model for new Catalog2Registration

joelverhagen:
Progress on https://github.com/nuget/nugetgallery/issues/7736.

PR #707, opened 2019-12-06
Add HiveMerger and supporting bookkeeping models

joelverhagen:
Progress https://github.com/NuGet/NuGetGallery/issues/7736.
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/706.

This introduces the core merging logic and associated bookkeeping classes. Consider the bookkeeping classes as implementation details of the `HiveMerger`. Their purpose is to make the logic of merging easier to read. They also are the things that "remember" which non-inlined pages have been download. The bookkeeping classes are:

- `IndexInfo` - holds a `RegistrationIndex` and allows adding and removing pages
- `PageInfo` - holds a `RegistrationPage` and allows adding and removing items and can transition a page from inlined to non-inlined and vice-versa
- `LeafInfo` - holds a `RegistrationLeafItem`

Unit tests will be added by a subsequent PR.

Note that this logic is relatively tricky (it took multiple iterations to get right) but I am confident in the correctness because of some exhaustive unit tests I added.

The `HiveMerger.MergeAsync` method can be considered the entry point.

### DIFF
--- a/src/NuGet.Jobs.Catalog2Registration/Hives/Bookkeeping/LeafInfo.cs
+++ b/src/NuGet.Jobs.Catalog2Registration/Hives/Bookkeeping/LeafInfo.cs
@@ -1,0 +1,47 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using NuGet.Protocol.Registration;
+using NuGet.Versioning;
+
+namespace NuGet.Jobs.Catalog2Registration
+{
+    /// <summary>
+    /// A class that handled the bookkeeping of a leaf item. The leaf item has minimal bookkeeping required except
+    /// maintaining a parsed version object for easy comparison.
+    /// </summary>
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class LeafInfo
+    {
+        private string DebuggerDisplay => $"Leaf {Version.ToNormalizedString()}";
+
+        private LeafInfo(NuGetVersion version, RegistrationLeafItem leafItem)
+        {
+            Version = version ?? throw new ArgumentNullException(nameof(version));
+            LeafItem = leafItem ?? throw new ArgumentNullException(nameof(leafItem));
+        }
+
+        public static LeafInfo New(NuGetVersion version)
+        {
+            return new LeafInfo(version, new RegistrationLeafItem
+            {
+                CatalogEntry = new RegistrationCatalogEntry
+                {
+                    Version = version.ToFullString(),
+                }
+            });
+        }
+
+        public static LeafInfo Existing(RegistrationLeafItem leafItem)
+        {
+            return new LeafInfo(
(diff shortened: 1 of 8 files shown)


src/NuGet.Jobs.Catalog2Registration/Hives/Bookkeeping/LeafInfo.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            return new LeafInfo(
+                NuGetVersion.Parse(leafItem.CatalogEntry.Version),
+                leafItem);
+        }
+
+        public NuGetVersion Version { get; }
+        public RegistrationLeafItem LeafItem { get; }
+
+    }
```
loic-sharma: super nit: there's an empty line above this one

src/NuGet.Jobs.Catalog2Registration/Hives/HiveMerger.cs
```diff
@@ -0,0 +1,311 @@
...
+                    // The current position in the item list is too low for the catalog version. Keep looking.
+                    itemIndex++;
+                }
+                else if (entry.PackageIdentity.Version == items[itemIndex].Version)
+                {
+                    if (entry.IsPackageDelete)
+                    {
+                        // Remove the registration leaf item. The current catalog commit item represents a
+                        // delete for this version.
+                        _logger.LogInformation(
```
loic-sharma: Does this logging context have the package id? If not, do you think it's worth adding the package id to these logs for debugging purposes?

src/NuGet.Jobs.Catalog2Registration/Hives/HiveMerger.cs
```diff
@@ -0,0 +1,311 @@
...
+                }
+
+                // Push the remaining catalog leaves into the last page.
+                while (catalogIndex < sortedCatalog.Count)
+                {
+                    var catalog = sortedCatalog[catalogIndex];
+                    itemIndex = await MergeEntryIntoPageAsync(context, catalog, pageIndex, itemIndex);
+                    catalogIndex++;
+                }
+                RemovePageAtIfEmpty(context, pageIndex);
```
loic-sharma: Nit: new line between 135 and 136? Feel free to ignore if this is intentional

src/NuGet.Jobs.Catalog2Registration/Hives/Bookkeeping/LeafInfo.cs
joelverhagen: Unintentional, good catch!

src/NuGet.Jobs.Catalog2Registration/Hives/HiveMerger.cs
joelverhagen: It's part of the logging context. I am hoping to lean on that entirely. Logging the package IDs in every log line would be painful.

src/NuGet.Jobs.Catalog2Registration/Hives/HiveMerger.cs
joelverhagen: Not intentional, good catch.

PR #708, opened 2019-12-10
Add unit tests for HiveMerger

joelverhagen:
There are two approaches in this PR for unit testing the complex HiveMerger component (PR'd with https://github.com/NuGet/NuGet.Services.Metadata/pull/707).

1. Traditional unit tests covered happy path case and edge cases.
2. A "full enumeration" test that generates all possible subsets of a set of versions for both the existing versions in a registration index and the versions coming from the catalog.

The reason this full enumeration suite is valuable is because the naive implementation of HiveMerger is trivial and can be done in the assertion step of a unit test. The implementation we have for product code (not tests) is more complex because it needs to be more performance and avoid unnecessary HTTP requests.

Progress on https://github.com/NuGet/NuGetGallery/issues/7736

### DIFF
--- a/tests/NuGet.Jobs.Catalog2Registration.Tests/Hives/HiveMergerFacts.FullEnumeration.cs
+++ b/tests/NuGet.Jobs.Catalog2Registration.Tests/Hives/HiveMergerFacts.FullEnumeration.cs
@@ -1,0 +1,317 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using NuGet.Services.Metadata.Catalog.Helpers;
+using NuGet.Services.V3.Support;
+using NuGet.Versioning;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace NuGet.Jobs.Catalog2Registration
+{
+    public partial class HiveMergerFacts
+    {
+        public class FullEnumeration
+        {
+            private readonly ITestOutputHelper _output;
+
+            public FullEnumeration(ITestOutputHelper output)
+            {
+                _output = output;
+            }
+
+            [Theory]
+            [InlineData(1, 2, 4, 1, 7)] // Use page size 1 to try some extreme prepending cases.
+            [InlineData(2, 2, 6, 1, 7)] // Use page size 2 so we can easily have up to 3 pages.
+            [InlineData(3, 2, 5, 1, 6)] // Use page size 3 to allow insertions without bounds changes.
+            public async Task Execute(int maxLeavesPerPage, int minExisting, int maxExisting, int minUpdated, int maxUpdated)
+            {
+                var config = new Catalog2RegistrationConfiguration();
+                var options = new SimpleOptions<Catalog2RegistrationConfiguration>(config);
+                var logger = new NullLogger<HiveMerger>();
+                var target = new HiveMerger(options, logger);
+
+                config.MaxLeavesPerPage = maxLeavesPerPage;
+
+                var allExisting = Enumerable.Range(minExisting, (maxExisting - minExisting) + 1).Select(m => new NuGetVersion(m, 0, 0)).ToList();
+                var allUpdated = Enumerable.Range(minUpdated, (maxUpdated - minUpdated) + 1).Select(m => new NuGetVersion(m, 0, 0)).ToList();
+                var versionToNormalized = allExisting.Concat(allUpdated).Distinct().ToDictionary(v => v, v => v.ToNormalizedString());
+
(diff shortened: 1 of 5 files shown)


joelverhagen:
:bell: could I get some eyes on this?

tests/NuGet.Jobs.Catalog2Registration.Tests/Hives/HiveMergerFacts.MergeAsync.cs
```diff
@@ -0,0 +1,411 @@
...
+using System.Threading.Tasks;
+using NuGet.Versioning;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace NuGet.Jobs.Catalog2Registration
+{
+    public partial class HiveMergerFacts
+    {
+        public class MergeAsync : Facts
```
agr: Please, don't name test classes exactly like methods, it messes up Ctrl+T search in VS: classes pop up in search higher than methods and in most cases you'd want method.

tests/NuGet.Jobs.Catalog2Registration.Tests/Hives/HiveMergerFacts.Support.cs
```diff
@@ -0,0 +1,183 @@
...
+
+            return versions;
+        }
+
+        private static List<CatalogCommitItem> MakeSortedCatalog(params VersionAction[] versions)
+        {
+            return MakeSortedCatalog(versions.ToList());
+        }
+
+        private static List<CatalogCommitItem> MakeSortedCatalog(List<VersionAction> sortedVersionActions)
```
agr: `ICollection<VersionAction>`? Will save the `ToList()` call couple of lines above.

tests/NuGet.Jobs.Catalog2Registration.Tests/Hives/HiveMergerFacts.FullEnumeration.cs
```diff
@@ -0,0 +1,319 @@
...
+                // Build the input data structures.
+                var sortedCatalog = MakeSortedCatalog(allUpdated);
+                var indexInfo = MakeIndexInfo(existing, config.MaxLeavesPerPage, versionToNormalized);
+
+                // Determine which pages will definitely not be affected.
+                var unaffectedPages = new List<PageInfo>();
+                if (existing.Any())
+                {
+                    var minVersion = allUpdated.Min(x => x.Version);
+                    for (int i = 0; i < indexInfo.Items.Count - 1; i++)
```
agr: `unaffectedPages = indexInfo.Items.Where(ii => ii.Upper < minVersion).ToList()`?

tests/NuGet.Jobs.Catalog2Registration.Tests/Hives/HiveMergerFacts.FullEnumeration.cs
```diff
@@ -0,0 +1,319 @@
...
+                }
+
+                // Verify the modified pages are in the index.
+                foreach (var pageInfo in result.ModifiedPages)
+                {
+                    Assert.True(indexInfo.Items.Contains(pageInfo), "A modified page must be in the index.");
+                }
+
+                // Verify the leaf item infos match the leaf items.
+                Assert.True(
```
agr: `Assert.Equal`

joelverhagen:
> I am a little bit concerned that all versions in all tests have minor and patch portions set to 0. Do we have tests that make sure that inside the `HiveMerger` correct version comparison is used?

Great point! I'll look into it.

tests/NuGet.Jobs.Catalog2Registration.Tests/Hives/HiveMergerFacts.MergeAsync.cs
joelverhagen: Sounds good. Would something like `TheMergeAsyncMethod` be better?

tests/NuGet.Jobs.Catalog2Registration.Tests/Hives/HiveMergerFacts.FullEnumeration.cs
joelverhagen: During the debugging I found that the custom messages were more useful than the counts that come from `Assert.Equal`.

tests/NuGet.Jobs.Catalog2Registration.Tests/Hives/HiveMergerFacts.MergeAsync.cs
joelverhagen: Did that!

tests/NuGet.Jobs.Catalog2Registration.Tests/Hives/HiveMergerFacts.Support.cs
joelverhagen: Done.

tests/NuGet.Jobs.Catalog2Registration.Tests/Hives/HiveMergerFacts.FullEnumeration.cs
joelverhagen: Done.

joelverhagen:
Added more version variants.

PR #709, opened 2019-12-10
Add RegistrationUrlBuilder

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/7736.
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/708.

This will be used when building and updating entities with interlinking URLs.

### DIFF
--- a/src/NuGet.Jobs.Catalog2Registration/Schema/RegistrationUrlBuilder.cs
+++ b/src/NuGet.Jobs.Catalog2Registration/Schema/RegistrationUrlBuilder.cs
@@ -1,0 +1,145 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Options;
+using NuGet.Services;
+using NuGet.Versioning;
+
+namespace NuGet.Jobs.Catalog2Registration
+{
+    public class RegistrationUrlBuilder
+    {
+        private readonly IOptionsSnapshot<Catalog2RegistrationConfiguration> _options;
+        private readonly string _legacyBaseUrl;
+        private readonly string _gzippedBaseUrl;
+        private readonly string _semver2BaseUrl;
+
+        public RegistrationUrlBuilder(IOptionsSnapshot<Catalog2RegistrationConfiguration> options)
+        {
+            _options = options ?? throw new ArgumentNullException(nameof(options));
+
+            _legacyBaseUrl = EnsureTrailingSlash(_options.Value.LegacyBaseUrl);
+            _gzippedBaseUrl = EnsureTrailingSlash(_options.Value.GzippedBaseUrl);
+            _semver2BaseUrl = EnsureTrailingSlash(_options.Value.SemVer2BaseUrl);
+        }
+
+        public string GetIndexPath(string id)
+        {
+            if (id == null)
+            {
+                throw new ArgumentNullException(nameof(id));
+            }
+
+            return $"{Uri.EscapeUriString(id.ToLowerInvariant())}/index.json";
+        }
+
+        public string GetIndexUrl(HiveType hive, string id)
+        {
+            return GetBaseUrl(hive) + GetIndexPath(id);
+        }
+
+        public string GetInlinedPageUrl(HiveType hive, string id, NuGetVersion lower, NuGetVersion upper)
+        {
+            return GetIndexUrl(hive, id) + "#" + GetPageFragment(lower, upper);
+        }
+
+        public string ConvertHive(HiveType fromHive, HiveType toHive, string url)
+        {
+            var path = ConvertToPath(fromHive, url);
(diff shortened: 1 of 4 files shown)


src/NuGet.Jobs.Catalog2Registration/Schema/RegistrationUrlBuilder.cs
```diff
@@ -0,0 +1,110 @@
...
+            switch (hive)
+            {
+                case HiveType.Legacy:
+                    return _legacyBaseUrl;
+                case HiveType.Gzipped:
+                    return _gzippedBaseUrl;
+                case HiveType.SemVer2:
+                    return _semver2BaseUrl;
+                default:
+                    throw new NotImplementedException($"The hive type '{hive}' does not have configured base URL.");
```
xavierdecoster: nit: `The hive type '{hive}' does not have *a* configured base URL.`

src/NuGet.Jobs.Catalog2Registration/Schema/RegistrationUrlBuilder.cs
```diff
@@ -0,0 +1,110 @@
...
+        private static string GetPageFragment(NuGetVersion lower, NuGetVersion upper)
+        {
+            return
+                $"page/" +
+                $"{Uri.EscapeUriString(lower.ToNormalizedString().ToLowerInvariant())}/" +
+                $"{Uri.EscapeUriString(upper.ToNormalizedString().ToLowerInvariant())}";
+        }
+
+        public string GetLeafPath(string id, NuGetVersion version)
+        {
```
xavierdecoster: Null checks?

src/NuGet.Jobs.Catalog2Registration/Schema/RegistrationUrlBuilder.cs
```diff
@@ -0,0 +1,110 @@
...
+            return url.Substring(baseUrl.Length);
+        }
+
+        public string GetPageUrl(HiveType hive, string id, NuGetVersion lower, NuGetVersion upper)
+        {
+            return GetBaseUrl(hive) + GetPagePath(id, lower, upper);
+        }
+
+        public string GetPagePath(string id, NuGetVersion lower, NuGetVersion upper)
+        {
```
xavierdecoster: Null checks

src/NuGet.Jobs.Catalog2Registration/Schema/RegistrationUrlBuilder.cs
```diff
@@ -0,0 +1,110 @@
...
+            {
+                return null;
+            }
+
+            var path = ConvertToPath(fromHive, url);
+            return GetBaseUrl(toHive) + path;
+        }
+
+        public string ConvertToPath(HiveType hive, string url)
+        {
```
xavierdecoster: Null checks

src/NuGet.Jobs.Catalog2Registration/Schema/RegistrationUrlBuilder.cs
```diff
@@ -0,0 +1,110 @@
...
+            _options = options ?? throw new ArgumentNullException(nameof(options));
+
+            _legacyBaseUrl = EnsureTrailingSlash(_options.Value.LegacyBaseUrl);
+            _gzippedBaseUrl = EnsureTrailingSlash(_options.Value.GzippedBaseUrl);
+            _semver2BaseUrl = EnsureTrailingSlash(_options.Value.SemVer2BaseUrl);
+        }
+
+        public string GetIndexPath(string id)
+        {
```
xavierdecoster: Null check

src/NuGet.Jobs.Catalog2Registration/Schema/RegistrationUrlBuilder.cs
joelverhagen: In general I don't put null checks are all the levels. This is a low level component so I thought we wouldn't need to null check here. If you feel strongly I can put them in.

src/NuGet.Jobs.Catalog2Registration/Schema/RegistrationUrlBuilder.cs
joelverhagen: Oh nice catch.

src/NuGet.Jobs.Catalog2Registration/Schema/RegistrationUrlBuilder.cs
xavierdecoster: Even though low level, it's a public API (from technical point of view). 
I personally would just add them to the places I highlighted, as the other public methods are calling into these and would be implicitly covered. Your call :)

joelverhagen: Okay, sounds good. I'll make that change.

joelverhagen: Done.

PR #711, opened 2019-12-12
[PackageType]Add fields for PackageType to document

ryuyu:
Adds following fields to search document:
 1. PackageTypes
    1. Lowercased list of packageTypes.
    2. Filterable
 2. DisplayPackageTypes
    1. Capitalization as originally in nuspec
    2. Used in return values

### DIFF
--- a/src/NuGet.Services.AzureSearch/Models/SearchDocument.cs
+++ b/src/NuGet.Services.AzureSearch/Models/SearchDocument.cs
@@ -44,11 +44,12 @@ namespace NuGet.Services.AzureSearch
             public string SearchFilters { get; set; }
 
             [IsFilterable]
-            public string[] PackageTypes { get; set; }
+            [Analyzer(ExactMatchCustomAnalyzer.Name)]
+            public string[] FilterablePackageTypes { get; set; }
 
             public string FullVersion { get; set; }
             public string[] Versions { get; set; }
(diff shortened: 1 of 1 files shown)


src/NuGet.Services.AzureSearch/Models/SearchDocument.cs
```diff
@@ -43,8 +43,12 @@ public class UpdateLatest : BaseMetadataDocument, IVersions, IOwners
             [IsFilterable]
             public string SearchFilters { get; set; }
 
+            [IsFilterable]
+            public string[] PackageTypes { get; set; }
```
joelverhagen: can we put [Analyzer(ExactMatchCustomAnalyzer.Name)] on here too?

joelverhagen: nit: Personally I would switch the names to `FilterablePackageTypes` and `PackageTypes` -- I would say in general the `PackageTypes` are the real values and the filterable ones are a special case

ryuyu: Yea we can.  I wasn't sure if that was meaningful on filterable, since I think this requires exact match? I can definitely add it.

ryuyu: swapped name specificity and added the exact match analyser attribute.

PR #713, opened 2019-12-13
Add support for querying by package type in Azure Search Service

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/7754

### DIFF
--- a/src/NuGet.Services.AzureSearch/SearchService/SearchParametersBuilder.cs
+++ b/src/NuGet.Services.AzureSearch/SearchService/SearchParametersBuilder.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using Microsoft.Azure.Search.Models;
+using NuGet.Packaging;
 using NuGetGallery;
 
 namespace NuGet.Services.AzureSearch.SearchService
@@ -73,7 +74,7 @@ namespace NuGet.Services.AzureSearch.SearchService
             }
             else
             {
-                ApplySearchIndexFilter(searchParameters, request, isDefaultSearch);
+                ApplySearchIndexFilter(searchParameters, request, isDefaultSearch, packageType: null);
             }
 
             return searchParameters;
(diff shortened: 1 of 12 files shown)


src/NuGet.Services.AzureSearch/SearchService/SearchParametersBuilder.cs
```diff
@@ -93,7 +94,7 @@ public SearchParameters Autocomplete(AutocompleteRequest request, bool isDefault
         {
             var searchParameters = NewSearchParameters();
 
-            ApplySearchIndexFilter(searchParameters, request, isDefaultSearch);
+            ApplySearchIndexFilter(searchParameters, request, isDefaultSearch, packageType: null);
```
loic-sharma: Do we know that the .NET CLI folks don’t plan to use autocomplete endpoint? Maybe we should ask them? No need to block this PR on that though :)

joelverhagen: Hey, you're on vacation! Get outta here.

Good point, I asked them. Should be easy to add one way or another but if they don't expect to use it we can exclude it from the MVP.

src/NuGet.Services.AzureSearch/SearchService/SearchResponseBuilder.cs
```diff
@@ -384,6 +385,22 @@ private V3SearchPackage ToV3SearchPackage(SearchDocument.Full result, string reg
             };
         }
 
+        private List<V3SearchPackageType> GetV3SearchPackageTypes(SearchDocument.UpdateLatest document)
+        {
+            if (document.PackageTypes == null || document.PackageTypes.Length == 0)
+            {
+                return null;
```
ryuyu: This case is a pretty good indication that we are running new search service against an old index. Do we expect this to be a normal case at any point? or should we log something so we are aware? I don't expect us to ever spend any extended time in this state, but just in case (also sanity check once we go full new index).

joelverhagen: I chose not to log in this case because it would be very verbose -- essentially once per search entry.

joelverhagen: > Do we expect this to be a normal case at any point?

Normal from our roll-out plan perspective - no. We should rebuild first, then deploy these bits. But have having catastrophic failure in this case could be painful during the rollout (e.g. deploying a build to DEV environment prior to the rebuild).

PR #715, opened 2019-12-16
[PackageType]Support PackageType in db2azuresearch

ryuyu:
Enables insertion of packageType into the document from AzureSearch.
Note that this also explicitly causes catalog2azureSearch to insert always dependency (ignore actual packageTypes) for now.
Catalog2AzureSearch should NOT be deployed until the following change to enable real packageTypes in catalog2AzureSearch is merged.

### DIFF
--- a/tests/NuGet.Services.AzureSearch.Tests/SearchDocumentBuilderFacts.cs
+++ b/tests/NuGet.Services.AzureSearch.Tests/SearchDocumentBuilderFacts.cs
@@ -379,12 +379,18 @@ namespace NuGet.Services.AzureSearch
         ""azure-sdk""
       ],
       ""searchFilters"": """ + expected + @""",
+      ""filterablePackageTypes"": [
+        ""dependency""
+      ],
       ""fullVersion"": ""7.1.2-alpha+git"",
       ""versions"": [
         ""1.0.0"",
         ""2.0.0+git"",
         ""3.0.0-alpha.1"",
         ""7.1.2-alpha+git""
+      ],
+      ""packageTypes"": [
+        ""Dependency""
       ],
       ""isLatestStable"": false,
       ""isLatest"": true,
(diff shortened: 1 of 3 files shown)


tests/NuGet.Services.AzureSearch.Tests/SearchDocumentBuilderFacts.cs
```diff
@@ -734,6 +746,102 @@ public async Task SetsExpectedProperties(SearchFilters searchFilters, string exp
...
+                    isLatest: true,
+                    fullVersion: Data.FullVersion,
+                    package: package,
+                    owners: Data.Owners,
+                    totalDownloadCount: Data.TotalDownloadCount,
+                    isExcludedByDefault: false);
+
+                SetDocumentLastUpdated(document);
+                var json = await SerializationUtilities.SerializeToJsonAsync(document);
+                Assert.Equal(@"{
```
joelverhagen: Most of this assertion is covered above. Can we just assert the two properties?

tests/NuGet.Services.AzureSearch.Tests/SearchDocumentBuilderFacts.cs
```diff
@@ -824,6 +932,59 @@ public abstract class BaseFacts
...
+            public static IEnumerable<object[]> PackageTypesData => new[]
+            {
+                new object[] {
+                    new List<PackageType> {
+                        new PackageType
+                        {
+                            Name = "DotNetCliTool"
+                        }
+                    },
+                    @"""dotnetclitool""",
```
joelverhagen: Note DotnetTool is more the canonical example

joelverhagen: Can we cover both the null and empty array cases?

ryuyu: There is null case and empty array case below. Or do you mean when the page will have null/empty? This should never be the case.

joelverhagen: Oh sorry, I missed it. Looks great.

tests/NuGet.Services.AzureSearch.Tests/SearchDocumentBuilderFacts.cs
ryuyu: Updated to contains assertion for filterablePackageTypes and packageTypes. much more readable.

PR #716, opened 2019-12-17
Support PackageType in Catalog2AzureSearch

ryuyu:
Adds PackageType support to Catalog2AzureSearch.

### DIFF
--- a/src/NuGet.Protocol.Catalog/Models/PackageType.cs
+++ b/src/NuGet.Protocol.Catalog/Models/PackageType.cs
@@ -11,12 +11,6 @@ namespace NuGet.Protocol.Catalog
     /// </summary>
     public class PackageType
     {
-        [JsonProperty("@id")]
-        public string Url { get; set; }
-
-        [JsonProperty("@type")]
-        public string Type { get; set; }
-
         [JsonProperty("name")]
         public string Name { get; set; }
 
--- a/src/NuGet.Protocol.Catalog/Models/PackageType.cs
+++ b/src/NuGet.Protocol.Catalog/Models/PackageType.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace NuGet.Protocol.Catalog
+{
+    /// <summary>
+    /// Source: https://docs.microsoft.com/en-us/nuget/api/catalog-resource#catalog-leaf
+    /// </summary>
+    public class PackageType
+    {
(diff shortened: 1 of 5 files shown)


src/NuGet.Protocol.Catalog/Models/PackageType.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace NuGet.Protocol.Catalog
+{
+    /// <summary>
+    /// Source: https://docs.microsoft.com/en-us/nuget/api/catalog-resource#catalog-leaf
+    /// </summary>
+    public class PackageType
+    {
+        [JsonProperty("@id")]
+        public string Url { get; set; }
```
joelverhagen: Nothing but the `name` property are necessary. I would prefer excluding `@id` and `@type` since they are very stupid values. `version` is fine albeit only forward looking.

tests/NuGet.Services.AzureSearch.Tests/SearchDocumentBuilderFacts.cs
```diff
@@ -442,6 +442,36 @@ public async Task SetsExpectedProperties(SearchFilters searchFilters, string exp
...
+                    isLatestStable: false,
+                    isLatest: true,
+                    normalizedVersion: Data.NormalizedVersion,
+                    fullVersion: Data.FullVersion,
+                    leaf: leaf,
+                    owners: Data.Owners);
+
+                SetDocumentLastUpdated(document);
+                var json = await SerializationUtilities.SerializeToJsonAsync(document);
+                Assert.Contains(@"
```
joelverhagen: It might make the code a little more maintainable in these JSON snippet cases to assert the non-serialized values on the document rather than the JSON. I have the single mega-JSON assertion because the shape of the whole document is expected to adhere to a certain contract w.r.t. to a deployed Azure Search index.

tests/NuGet.Services.AzureSearch.Tests/SearchDocumentBuilderFacts.cs
```diff
@@ -442,6 +442,35 @@ public async Task SetsExpectedProperties(SearchFilters searchFilters, string exp
...
+                    isLatestStable: false,
+                    isLatest: true,
+                    normalizedVersion: Data.NormalizedVersion,
+                    fullVersion: Data.FullVersion,
+                    leaf: leaf,
+                    owners: Data.Owners);
+
+                SetDocumentLastUpdated(document);
+                Assert.Equal(document.FilterablePackageTypes.Length, document.PackageTypes.Length);
+                Assert.Equal(expectedFilterable.Length, document.FilterablePackageTypes.Length);
```
joelverhagen: nit: you can do `Assert.Equal` with entire collections, avoiding the length assertion and loop

PR #721, opened 2019-12-20
[C2R 2] Add hive storage to handle blob storage and replica hive logic

joelverhagen:
Address https://github.com/nuget/nugetgallery/issues/7738
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/720

### DIFF
--- a/src/NuGet.Jobs.Catalog2Registration/Hives/HiveStorage.cs
+++ b/src/NuGet.Jobs.Catalog2Registration/Hives/HiveStorage.cs
@@ -1,0 +1,434 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.IO.Compression;
+using System.Linq;
+using System.Net;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Microsoft.WindowsAzure.Storage;
+using Microsoft.WindowsAzure.Storage.Blob;
+using Newtonsoft.Json;
+using NuGet.Protocol;
+using NuGet.Protocol.Catalog;
+using NuGet.Protocol.Registration;
+using NuGet.Versioning;
+using NuGetGallery;
+
+namespace NuGet.Jobs.Catalog2Registration
+{
+    public class HiveStorage : IHiveStorage
+    {
+        private readonly ICloudBlobClient _cloudBlobClient;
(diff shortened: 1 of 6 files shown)


src/NuGet.Jobs.Catalog2Registration/Hives/HiveStorage.cs
```diff
@@ -0,0 +1,434 @@
...
+                {
+                    continue;
+                }
+
+                if (currentHive != previousHive)
+                {
+                    // The hive defines the base URL. If we're now writing to another hive (which is the case if this
+                    // is the 2nd ... Nth hive) then we need to update the URLs to point to the current hive.
+                    updateUrls(entity, previousHive, currentHive);
+                }
```
loic-sharma: This code assumes that each invocation of `updateUrls` will replace all properties on the entity. While this is true today, I'm concerned that this may break if we ever set a property only on a certain hive that is a replica of another hive.

For example, let's say I have hive `A` and replica hives `B` and `C`. I change `UpdateXUrls` builder so I update the property `newUrl` **only** when `toHive` is `B`. This would have the unexpected side effect of setting `newUrl` for hive `C` (and the entity won't be returned to the original state once `WriteAsync` completes). This seems like a subtle bug that one may easily miss.

Do you think this concern is worth addressing? One possible solution may be to clone the entity before updating its URLs.

joelverhagen: I considering a "clone" approach too but a properly written clone method needs to copy all properties of the POCO over as well. I thought it would be easier to get the "update URLs" method right than a "copy all properties" method since the former is less likely to be affected by future changes (since not all model changes are URL additions).

I think some other alternatives are:

1. Abstract URL setting so that both the initial population of the properties and the converting flow use the same data source (i.e. one method called in two different contexts)
1. Change the URL setting flow completely so one module has a "set URLs" method that is invoked for `A`, `B`, and `C` here (i.e one method called in one, low level context)

I'll think about this some more. I thought this would be enough but if you're concerned about it, it's probably worth rethinking.

loic-sharma: > I thought it would be easier to get the "update URLs" method right than a "copy all properties" method since the former is less likely to be affected by future changes (since not all model changes are URL additions).

That's a good point. Another suggestion: could you test all "from" and "to" hive combinations on all `updateUrls` handlers and ensure that all properties of the POCO are back to the original state?

joelverhagen: Haha, that's exactly the path I was going. And I recurse on the JSON object model finding all properties that look like URLs and assert the expected set.

PR #722, opened 2019-12-20
[C2R 3] Add hive updater to orchestrate the writes and deletes for a single hive index

joelverhagen:
Address https://github.com/nuget/nugetgallery/issues/7737
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/721

### DIFF
--- a/src/NuGet.Jobs.Catalog2Registration/Hives/HiveUpdater.cs
+++ b/src/NuGet.Jobs.Catalog2Registration/Hives/HiveUpdater.cs
@@ -1,0 +1,498 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using NuGet.Protocol.Catalog;
+using NuGet.Services;
+using NuGet.Services.Metadata.Catalog;
+using NuGet.Services.Metadata.Catalog.Helpers;
+using NuGet.Versioning;
+
+namespace NuGet.Jobs.Catalog2Registration
+{
+    public class HiveUpdater : IHiveUpdater
+    {
+        private static readonly IReadOnlyList<Uri> DeleteUris = new[] { Schema.DataTypes.PackageDelete };
+        
+        private readonly IHiveStorage _storage;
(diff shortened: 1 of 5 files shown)


src/NuGet.Jobs.Catalog2Registration/Hives/HiveUpdater.cs
```diff
@@ -0,0 +1,498 @@
...
+            }
+            GuardInput(entries, entryToCatalogLeaf);
+            var sortedCatalog = entries.OrderBy(x => x.PackageIdentity.Version).ToList();
+            var versionToCatalogLeaf = entryToCatalogLeaf.ToDictionary(x => x.Key.PackageIdentity.Version, x => x.Value);
+
+            // Remove SemVer 2.0.0 versions if this hive should only have SemVer 1.0.0 versions.
+            if (ShouldExcludeSemVer2(hive))
+            {
+                Guard.Assert(
+                    replicaHives.All(x => ShouldExcludeSemVer2(x)),
```
loic-sharma: >replicaHives.All(x => ShouldExcludeSemVer2(x)) [](start = 20, length = 46)

Super nit: `replicaHives.All(ShouldExcludeSemVer2)`. You can make line 72 `replicaHives.Any(ShouldExcludeSemVer2)` too if you'd like. I don't feel strongly about this, feel free to keep it as is if you prefer it.

src/NuGet.Jobs.Catalog2Registration/Hives/HiveUpdater.cs
```diff
@@ -0,0 +1,498 @@
...
+                {
+                    continue;
+                }
+
+                var version = catalogCommitItem.PackageIdentity.Version;
+                var catalogLeaf = versionToCatalogLeaf[version];
+                if (catalogLeaf.IsSemVer2())
+                {
+                    // Turn the PackageDetails into a PackageDelete to ensure that a known SemVer 2.0.0 version is not
+                    // in a hive that should not have SemVer 2.0.0. This may cause a little bit more work (like a
```
loic-sharma: Snazzy 👍

src/NuGet.Jobs.Catalog2Registration/Hives/HiveUpdater.cs
joelverhagen: Sure, sounds fine.

joelverhagen: Changing to `Any` won't work because `Any` returns false if there are no elements in the sequence.

PR #725, opened 2019-12-20
[C2R 6] Add in-memory integration tests for registration updater down to storage

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/7739
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/724

### DIFF
--- a/tests/NuGet.Jobs.Catalog2Registration.Tests/IntegrationTests.cs
+++ b/tests/NuGet.Jobs.Catalog2Registration.Tests/IntegrationTests.cs
@@ -1,0 +1,451 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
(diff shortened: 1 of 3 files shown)


tests/NuGet.Jobs.Catalog2Registration.Tests/IntegrationTests.cs
```diff
@@ -0,0 +1,451 @@
...
+        public async Task DeleteWhenNoVersionExists()
+        {
+            // Arrange & Act
+            await DeleteVersionsAsync("7.1.2-ALPHA");
+
+            // Assert
+            AssertContainerExistence();
+            AssertBlobExistence(
+                existing: new string[0],
+                deleted: new[]
```
loic-sharma: Super nitpick: this name is a tad misleading in this context. Is there another name that could mean "deleted or does not exist"? Maybe `missing`? I don't feel strongly about this, so feel free to keep as is.

joelverhagen: Yup, sounds fine.

PR #726, opened 2019-12-20
[C2R 7] Add the Catalog2Registration command which is component activated by DI

joelverhagen:
Progress on https://github.com/nuget/nugetgallery/issues/7739
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/725

### DIFF
--- a/src/NuGet.Jobs.Catalog2Registration/Catalog2RegistrationCommand.cs
+++ b/src/NuGet.Jobs.Catalog2Registration/Catalog2RegistrationCommand.cs
@@ -1,0 +1,127 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using System.Net.Http;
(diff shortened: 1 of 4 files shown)


src/NuGet.Jobs.Catalog2Registration/Catalog2RegistrationCommand.cs
```diff
@@ -0,0 +1,127 @@
...
+                backCursor.Value);
+
+            // Run the collector.
+            await _collector.RunAsync(
+                frontCursor,
+                backCursor,
+                token);
+        }
+
+        private async Task CreateContainerIfExistsAsync(Func<Catalog2RegistrationConfiguration, string> getContainer)
```
loic-sharma: This name seems off 😂

```suggestion
        private async Task CreateContainerIfNotExistsAsync(Func<Catalog2RegistrationConfiguration, string> getContainer)
```

joelverhagen: oops, good catch

PR #728, opened 2019-12-20
[C2R 1] Update Lightning to support the new Catalog2Registration with -driver

joelverhagen:
Address https://github.com/nuget/nugetgallery/issues/7740
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/727

PR #729, opened 2020-01-03
[C2R 2] Add registration comparer to compare two registration hives

joelverhagen:
This basically is a generic JSON document comparer with some normalization logic to handle known differences. These known differences are either by design or quirks.

My goal was NOT to make super re-usable or high performance code. Instead, the goal is to find differences in the two implementations. The `JsonComparer` class code found all of the problems mentioned in `Normalizers.cs`, so I am confident in the results.

Progress on https://github.com/nuget/nugetgallery/issues/7741.
Depends on https://github.com/NuGet/NuGet.Services.Metadata/pull/728.

### DIFF
--- a/src/NuGet.Jobs.RegistrationComparer/HiveComparer.cs
+++ b/src/NuGet.Jobs.RegistrationComparer/HiveComparer.cs
@@ -1,0 +1,244 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using NuGet.Protocol.Catalog;
+using NuGet.Protocol.Registration;
+using NuGet.Services.Metadata.Catalog.Helpers;
+
+namespace NuGet.Jobs.RegistrationComparer
+{
+    public class HiveComparer
+    {
+        private readonly HttpClient _httpClient;
+        private readonly JsonComparer _comparer;
+        private readonly ILogger<HiveComparer> _logger;
+
+        public HiveComparer(
+            HttpClient httpClient,
+            JsonComparer comparer,
+            ILogger<HiveComparer> logger)
+        {
+            _httpClient = httpClient ?? throw new ArgumentNullException(nameof(httpClient));
+            _comparer = comparer ?? throw new ArgumentNullException(nameof(comparer));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public async Task CompareAsync(
+            IReadOnlyList<string> baseUrls,
+            string id,
+            IReadOnlyList<string> versions)
+        {
+            if (baseUrls.Count <= 1)
+            {
+                throw new ArgumentException("At least two base URLs must be provided.", nameof(baseUrls));
+            }
+
+            // Compare the indexes.
+            var rawIndexes = await Task.WhenAll(baseUrls.Select(x => GetIndexAsync(x, id)));
+            var areBothMissing = false;
+            for (var i = 1; i < baseUrls.Count; i++)
+            {
+                if (AreBothMissing(
+                   rawIndexes[i - 1].Url,
+                   rawIndexes[i].Url,
+                   rawIndexes[i - 1].Data,
+                   rawIndexes[i].Data))
+                {
+                    areBothMissing = true;
+                    continue;
+                }
+
+                var comparisonContext = new ComparisonContext(
+                    id,
+                    baseUrls[i - 1],
+                    baseUrls[i],
+                    rawIndexes[i - 1].Url,
+                    rawIndexes[i].Url,
+                    Normalizers.Index);
+
+                _comparer.Compare(
+                   rawIndexes[i - 1].Data,
+                   rawIndexes[i].Data,
+                   comparisonContext);
+            }
+
+            if (areBothMissing)
+            {
+                return;
+            }
+
+            // Deserialize the indexes so we can get the page URLs.
+            var indexes = new List<DownloadedData<RegistrationIndex>>();
+            foreach (var rawIndex in rawIndexes)
+            {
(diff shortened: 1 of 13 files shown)


src/NuGet.Jobs.RegistrationComparer/Normalizers.cs
```diff
@@ -0,0 +1,266 @@
...
+    public class Normalizers
+    {
+        private readonly static HashSet<string> PackageIdsWithDotDotDependencies = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+        {
+            "angularjs.TypeScript.DefinitelyTyped",
+            "devextreme.TypeScript.DefinitelyTyped",
+        };
+
+        private static readonly IReadOnlyList<KeyValuePair<Func<string, bool>, Func<JToken, bool, ComparisonContext, string>>> DefaultScalarNormalizers =
+            new List<KeyValuePair<Func<string, bool>, Func<JToken, bool, ComparisonContext, string>>>
```
loic-sharma: The types are a bit complex. Do you think it'd be easier to understand them if we added a few abstractions? Maybe:

* Func<string, bool>` could be a delegate named `PathMatcher`
* `Func<JToken, bool, ComparisonContext, string>` could be a delegate named `Normalize`
* `KeyValuePair<...>` could be a private type named `Normalizer`

loic-sharma: (I don't feel too strongly about this, feel free to ignore if you feel the abstractions would just make things more confusing)

src/NuGet.Jobs.RegistrationComparer/RegistrationComparerCommand.cs
```diff
@@ -0,0 +1,75 @@
...
+            _options = options ?? throw new ArgumentNullException(nameof(options));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public async Task ExecuteAsync()
+        {
+            await ExecuteAsync(CancellationToken.None);
+        }
+
+        private async Task ExecuteAsync(CancellationToken token)
```
loic-sharma: Nit: what do you think of making this method public and passing the `CancellationToken.None` in the `Job.Run`?

src/NuGet.Jobs.RegistrationComparer/RegistrationComparerCollectorLogic.cs
```diff
@@ -0,0 +1,114 @@
...
+                    {
+                        _logger.LogInformation("Verifying hive {Hive} for {PackageId}.", hive, id);
+                        try
+                        {
+                            await _comparer.CompareAsync(baseUrls, id, versions);
+                        }
+                        catch (Exception ex)
+                        {
+                            failures.Add(ex.Message);
+                            ResultWriter.WriteError(ex.Message);
```
loic-sharma: The real exception may be in an inner exception. Maybe use `.ToString()` instead?

src/NuGet.Jobs.RegistrationComparer/ResultWriter.cs
```diff
@@ -0,0 +1,29 @@
...
+        private static readonly object _errorsLock = new object();
+
+        public static void WriteWarning(string message)
+        {
+            lock (_warningsLock)
+            {
+                File.AppendAllLines("warnings.txt", new[] { message, Environment.NewLine });
+            }
+        }
+        public static void WriteError(string message)
```
loic-sharma: Nit: please add a new line above

src/NuGet.Jobs.RegistrationComparer/Normalizers.cs
joelverhagen: Great idea. These shapes morphed over time so I didn't think it was right at first but now they are stable.

src/NuGet.Jobs.RegistrationComparer/RegistrationComparerCommand.cs
joelverhagen: I'll give that a try. Sounds fine.

src/NuGet.Jobs.RegistrationComparer/RegistrationComparerCollectorLogic.cs
joelverhagen: Sure, sounds fine.

src/NuGet.Jobs.RegistrationComparer/Normalizers.cs
joelverhagen: Done.

src/NuGet.Jobs.RegistrationComparer/HiveComparer.cs
```diff
@@ -0,0 +1,244 @@
...
+                            rightUrl,
+                            urlToJson[leftUrl],
+                            urlToJson[rightUrl]))
+                        {
+                            continue;
+                        }
+                    }
+                    catch (InvalidOperationException ex)
+                    {
+                        ResultWriter.WriteWarning(ex.Message);
```
loic-sharma: Nit: consider recording `.ToString()` instead

joelverhagen: Chose not to in this case since the exception this case is a well-known throw-er. I can do the whole string if you want, though.

src/NuGet.Jobs.RegistrationComparer/Normalizers.cs
```diff
@@ -0,0 +1,268 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+using Newtonsoft.Json.Linq;
+using NuGet.Versioning;
+using ValueNormalizer = System.Collections.Generic.KeyValuePair<NuGet.Jobs.RegistrationComparer.ShouldNormalizeByPath, NuGet.Jobs.RegistrationComparer.NormalizeToken>;
+using ArrayNormalizer = System.Collections.Generic.KeyValuePair<NuGet.Jobs.RegistrationComparer.ShouldNormalizeByArray, System.Comparison<Newtonsoft.Json.Linq.JToken>>;
```
loic-sharma: Nice 👍

src/NuGet.Jobs.RegistrationComparer/HiveComparer.cs
loic-sharma: Its' fine as is :)

PR #732, opened 2020-01-16
Upgrade telemetry and add heartbeats support

xavierdecoster:
This PR
* removes some `ITelemetryInitializer` implementations (which have been moved into `NuGet.Services.Logging` in ServerCommon)
* upgrades ServerCommon dependencies to v2.66.0
* upgrades NuGetGallery.Core dependency to latest dev which depends on v2.66.0 (thus removing some binding redirects in `web.config`)
* upgrades Application Insights similar to NuGetGallery (no longer depend on static `TelemetryConfiguration.Active`, the heartbeats work-around, loading TelemetryModules through code...)

I had to upgrade BasicSearch project too as it has not been deleted yet, and shared dependencies required changes to the BasicSearch project. If the project gets deleted before this PR is merged, I'll simply remove the changes to the BasicSearch project from this PR.


src/Catalog/Telemetry/ITelemetryService.cs
```diff
@@ -11,6 +11,8 @@ namespace NuGet.Services.Metadata.Catalog
 {
     public interface ITelemetryService
     {
+        TelemetryClientWrapper TelemetryClient { get; }
```
skofman1: why is this needed? Isn't the goal of this interface to "hide" the implementation details of telemetry logging?

xavierdecoster: The `LightningJob` is the only one requiring a singleton `TelemetryClient` to be [registered in the ServicesCollection](https://github.com/NuGet/NuGet.Services.Metadata/pull/732/files#diff-7b3387943c5f27215edf918eac4da605R525), but in order to do so, I need to be able to access the singleton created during ApplicationInsights initialization.

It's either this change, or I need to pass in the `TelemetryClient` to the constructor of all jobs in addition to the `ITelemetryService`.

xavierdecoster: Actually, I'm wondering why this particular job needs this `TelemetryClient` to be available in the `ServicesCollection`... No idea which type is consuming this as a constructor argument.

@joelverhagen you any idea why this might be the case? It got introduced as part of [this commit](https://github.com/NuGet/NuGet.Services.Metadata/commit/68f7b9a12c737076689857dfb8f135ef99d96c85). My gut feeling says it has to do with the `IV3TelemetryService` that is possibly used by this job which requires another `TelemetryClient`, whereas we manually construct the other `ITelemetryService` and pass it to the job constructors of all jobs...

xavierdecoster: Did a refactoring of how `TelemetryService` is used and constructed in commit https://github.com/NuGet/NuGet.Services.Metadata/pull/732/commits/09a0ae7e27cac32c91e249017d4acf52deaca395.

src/Ng/Jobs/Catalog2PackageFixupJob.cs
```diff
@@ -15,20 +15,20 @@
...
+using NuGet.Services.Logging;
 using NuGet.Services.Metadata.Catalog;
 
 namespace Ng.Jobs
 {
     public class Catalog2PackageFixupJob : NgJob
     {
-        private const int MaximumPackageProcessingAttempts = 5;
-        private static readonly TimeSpan MaximumPackageProcessingTime = TimeSpan.FromMinutes(10);
-        private static readonly TimeSpan DonePollingInterval = TimeSpan.FromMinutes(1);
```
xavierdecoster: These are unused.

xavierdecoster:
Deployed the following `ng.exe`-based jobs to DEV from this branch to verify heartbeats and telemetry (check Heartbeats > DEV dashboard in Geneva):

* catalog2dnx
* catalog2icon
* catalog2monitoring
* catalog2lucene
* db2catalog
* db2monitoring
* monitoring2monitoring
* monitoringprocessor

These deployments confirm that the newly added `JobLoopExitCode` property of the `HeartbeatState` metric works as expected, and reports _unhealthy_ heartbeats for those jobs that throw exceptions.

Todo: 
* [ ] deploy, verify, and confirm search service
* [ ] get rid of old `BasicSearchService`
* [ ] clean up some temporary widgets on the dashboard used for verification

src/Ng/Program.cs
```diff
@@ -56,62 +58,107 @@ public static async Task MainAsync(string[] args)
...
+
             Trace.Close();
-            TelemetryConfiguration.Active.TelemetryChannel.Flush();
+            applicationInsightsConfiguration?.TelemetryConfiguration.TelemetryChannel.Flush();
+        }
+
+        private static ILoggerFactory ConfigureLoggerFactory(ApplicationInsightsConfiguration applicationInsightsConfiguration)
+        {
+            var loggerConfiguration = LoggingSetup.CreateDefaultLoggerConfiguration(withConsoleLogger: true);
+            loggerConfiguration.WriteTo.File("Log.txt", retainedFileCountLimit: 3, fileSizeLimitBytes: 1000000, rollOnFileSizeLimit: true);
```
joelverhagen: Do we really need file logging? The volume is so high that it's rather hard to use in many cases.

src/NuGet.Services.SearchService/App_Start/WebApiConfig.cs
```diff
@@ -157,14 +150,109 @@ private static AutofacWebApiDependencyResolver GetDependencyResolver(HttpConfigu
...
+        }
+
+        private static void RegisterApplicationInsightsTelemetryModules(TelemetryConfiguration configuration)
+        {
+            RegisterApplicationInsightsTelemetryModule(
+                new Microsoft.ApplicationInsights.WindowsServer.AppServicesHeartbeatTelemetryModule(),
+                configuration);
+
+            RegisterApplicationInsightsTelemetryModule(
+                new Microsoft.ApplicationInsights.WindowsServer.AzureInstanceMetadataTelemetryModule(),
```
joelverhagen: Why not a using?

src/Ng/Program.cs
xavierdecoster: This is not a new addition to the code. It's what we currently do. See https://github.com/NuGet/NuGet.Services.Metadata/blob/master/src/Ng/Program.cs#L78.

If we don't need it nor use it, we should indeed remove this. So do we ever us this?

src/NuGet.Services.SearchService/App_Start/WebApiConfig.cs
xavierdecoster: It was convenient when migrating from `applicationinsights.config` file, but other than that no particular reason.

Addressed with https://github.com/NuGet/NuGet.Services.Metadata/pull/732/commits/5d8b42da043cc692527046eb96609c4959df56ad

src/Ng/Program.cs
joelverhagen: Oops forgot it was preexisting. Out of scope.

PR #733, opened 2020-01-23
Remove file logging from RegistrationComparer now that AI logging is set up

joelverhagen:
Make the comparison container (just a cursor) public for easy monitoring
Related to https://github.com/nuget/nugetgallery/issues/7741


src/NuGet.Jobs.RegistrationComparer/HiveComparer.cs
```diff
@@ -159,7 +159,7 @@ public class HiveComparer
                     }
                     catch (InvalidOperationException ex)
                     {
-                        ResultWriter.WriteWarning(ex.Message);
+                        _logger.LogWarning(ex.Message);
```
loic-sharma: Should we log the full warning given that we're no longer constrained by a file?

joelverhagen: Sure, seems fine

PR #743, opened 2020-02-13
Fix Azure Search functional test that broke from "json" results changing

joelverhagen:
These tests are broken right now. I found this at deployment time.

I added a "BaseTestPackage.NoStable" package to all environments to ensure that this test assertions are always true. "json" worked for a while but it is prone to failure since people can push stable versions as their latest.


tests/NuGet.Services.AzureSearch.FunctionalTests/BasicTests/V3SearchProtocolTests.cs
```diff
@@ -135,11 +135,13 @@ public async Task ResultsHonorPreReleaseField()
 
             var hasPrereleaseVersions = resultsWithPrerelease
                 .Data
+                .SelectMany(x => x.Versions)
```
joelverhagen: We can verify all of the versions in the response -- not just the latest ones.

PR #749, opened 2020-02-24
Allow storage different from CDN to be tested by RegistrationComparer

joelverhagen:
Improved logging around HTTP requests.
Removed duplicate HTTP request (OMG).
Moved response body fetching into HttpClient.GetAsync so HttpClient.Timeout acts on it.

Right now, this job can only test China storage if it's running from China. This change enables hitting storage that doesn't sit behind the client's resolution of api.nuget.org.

Related to https://github.com/nuget/nugetgallery/issues/7741.

### DIFF
--- a/src/NuGet.Jobs.RegistrationComparer/HiveComparer.cs
+++ b/src/NuGet.Jobs.RegistrationComparer/HiveComparer.cs
@@ -4,10 +4,12 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
@@ -21,6 +23,8 @@ namespace NuGet.Jobs.RegistrationComparer
 {
     public class HiveComparer
     {
+        private static long _requestId = 0;
+
         private readonly HttpClient _httpClient;
         private readonly JsonComparer _comparer;
         private readonly IOptionsSnapshot<RegistrationComparerConfiguration> _options;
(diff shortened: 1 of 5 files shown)


src/NuGet.Jobs.RegistrationComparer/HiveComparer.cs
```diff
@@ -97,67 +103,71 @@ public class HiveComparer
...
+                    .Select(u => new { Hive = x, Url = u })
+                    .ToList())
                 .ToList();
 
-            var urls = new ConcurrentBag<string>(pageUrlGroups
+            var pairs = new ConcurrentBag<KeyValuePair<string, HiveConfiguration>>(pageUrlGroups
                 .SelectMany(x => x)
-                .Concat(leafUrlGroups.SelectMany(x => x)));
+                .Concat(leafUrlGroups.SelectMany(x => x))
+                .Select(x => new KeyValuePair<string, HiveConfiguration>(x.Url, x.Hive)));
```
agr: `KeyValuePair.Create(x.Url, x.Hive)` would be a bit more compact.
What is `Url` here, by the way?

zhhyu:
No need to update the test coverage?

joelverhagen:
> No need to update the test coverage?

This is a validation job for the one-time transition from old registration to new registration impl. I did not write unit tests for this level.

src/NuGet.Jobs.RegistrationComparer/HiveComparer.cs
joelverhagen: This is the official (not blob storage) URL of a registration leaf that needs to be downloaded and validated.

joelverhagen: The URL is there so we can parallelize the downloads of many leaves and pages.

I'll use the `KeyValuePair.Create` trick.

PR #750, opened 2020-02-25
Make ng.exe jobs insensitive towards command-line argument casing

xavierdecoster:
This was causing issues with the `-heartbeatIntervalSeconds` argument, which was case-sensitively compared against `-HeartbeatIntervalSeconds` as the expected argument.

Turns out all arguments passed into the `ng.exe`-based jobs were treated case-sensitive, which IMHO is not a great experience for a CLI.

This change will also allow us to update the `ng.exe` jobs' startup scripts to use PascalCased arguments instead, and align with our regular `NuGet.Jobs` pattern which results in more consistency.


xavierdecoster:
Note: I verified this with a private build deployment to DEV for `monitoringprocessor`.

PR #751, opened 2020-02-25
[Db2AzureSearch] Support skipping packages by prefix

loic-sharma:
The DEV environment has many tests packages, which makes the `Db2AzureSearch` tool quite slow on DEV. This change lets us skip these test packages by adding their prefixes to the config file:

```json
{
  "Db2AzureSearch": {
    "SkipPackagePrefixes": [
      "TestPackage",
      "E2E.",
      "Nuggets4NuGet.IngestionTime.",
    ],
  },
}
```

<details>
<summary>Full skip list for DEV</summary>
```json
    "SkipPackagePrefixes": [
      "Test",
      "E2E.",
      "NuGetFunctionalTest_",
      "VerifyPackageKey",
      "ScopedApiKeys",
      "DuplicatePushesAreRejected",
      "PackagePush_",
      "Nuggets4NuGet.IngestionTime.",
      "GetUpdates",
      "UploadPackageTo",
      "UploadAndUnlistPackages",
      "VerifyKeyReturns",
    ],
```
</details>

### DIFF
--- a/src/NuGet.Services.AzureSearch/Db2AzureSearch/NewPackageRegistrationProducer.cs
+++ b/src/NuGet.Services.AzureSearch/Db2AzureSearch/NewPackageRegistrationProducer.cs
@@ -43,6 +43,7 @@ namespace NuGet.Services.AzureSearch.Db2AzureSearch
             var ranges = await GetPackageRegistrationRangesAsync();
 
             // Fetch exclude packages list from auxiliary files.
+            // These packages are excluded from the default search's results.
             var excludedPackages = await _auxiliaryFileClient.LoadExcludedPackagesAsync();
 
             Guard.Assert(
@@ -206,6 +207,7 @@ namespace NuGet.Services.AzureSearch.Db2AzureSearch
                 var packageRegistrations = await query.ToListAsync();
 
                 return packageRegistrations
+                    .Where(pr => !ShouldSkipPackageRegistration(pr))
                     .Select(pr => new PackageRegistrationInfo(
                         pr.Key,
                         pr.Id,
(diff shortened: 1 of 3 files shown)


src/NuGet.Services.AzureSearch/Db2AzureSearch/NewPackageRegistrationProducer.cs
```diff
@@ -215,6 +217,24 @@ private async Task<IReadOnlyList<PackageRegistrationInfo>> GetPackageRegistratio
...
+            if (_options.Value.SkipPackagePrefixes == null)
+            {
+                return false;
+            }
+
+            foreach (var skippedPackagePrefix in _options.Value.SkipPackagePrefixes)
+            {
+                if (packageRegistration.Id.StartsWith(skippedPackagePrefix, StringComparison.OrdinalIgnoreCase))
+                {
+                    return true;
```
loic-sharma: There was a `LogDebug` here originally, but I removed it as it added ~15 minutes to Db2AzureSearch on DEV

joelverhagen: o_o

joelverhagen: You can ask the logger if debug is enabled and skip the invocation... but yeah seems unnecessary.

src/NuGet.Services.AzureSearch/Db2AzureSearch/NewPackageRegistrationProducer.cs
```diff
@@ -215,6 +217,24 @@ private async Task<IReadOnlyList<PackageRegistrationInfo>> GetPackageRegistratio
...
         }
 
+        private bool ShouldSkipPackageRegistration(PackageRegistration packageRegistration)
+        {
+            if (_options.Value.SkipPackagePrefixes == null)
+            {
+                return false;
+            }
+
+            foreach (var skippedPackagePrefix in _options.Value.SkipPackagePrefixes)
```
agr: I think this and following lines can be replaced with something like
```
return _options.Value.SkipPackagePrefixes.Any(prefix => packageRegistration.Id.StartsWith(prefix, StringComparison.OrdinalIgnoreCase));
```

agr: Do we care about performance here? Running every ID through a list of prefixes to ignore can be slow.

loic-sharma: The bottleneck for db2azuresearch is indexing data not producing it, so this change makes db2azuresearch signficantly faster on DEV. Also, this list will be `null` on INT and PROD and should have negligible impact.

loic-sharma: I don't have a strong preference either ways here. LINQ is shorter but results in a longer line here.

PR #752, opened 2020-03-04
Don't allow a single blocked worker task to make the process hang

joelverhagen:
Also, improve logging
Address https://github.com/NuGet/Engineering/issues/3042

### DIFF
--- a/src/Ng/Jobs/MonitoringProcessorJob.cs
+++ b/src/Ng/Jobs/MonitoringProcessorJob.cs
@@ -29,6 +29,7 @@ namespace Ng.Jobs
     {
         private const int DefaultQueueLoopDurationHours = 24;
         private const int DefaultQueueDelaySeconds = 30;
+        private static readonly TimeSpan MaxShutdownTime = TimeSpan.FromMinutes(1);
 
         private PackageValidator _packageValidator;
         private IStorageQueue<PackageValidatorContext> _queue;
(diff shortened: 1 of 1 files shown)


src/Ng/Jobs/MonitoringProcessorJob.cs
```diff
@@ -135,13 +136,44 @@ protected override async Task RunInternalAsync(CancellationToken cancellationTok
...
+
+                var firstTask = await Task.WhenAny(allWorkersTask, loopDurationPlusShutdownTask);
+                if (firstTask == loopDurationPlusShutdownTask)
+                {
+                    Logger.LogWarning("Not all workers shut down gracefully after {Duration}.", MaxShutdownTime);
+                }
+                else
+                {
+                    timeoutCancellationTokenSource.Cancel();
+                    Logger.LogInformation("All workers gracefully shut down.", MaxShutdownTime);
```
loic-sharma: `MaxShutdownTime` is unused here

src/Ng/Jobs/MonitoringProcessorJob.cs
```diff
@@ -29,6 +29,7 @@ public class MonitoringProcessorJob : LoopingNgJob
     {
         private const int DefaultQueueLoopDurationHours = 24;
         private const int DefaultQueueDelaySeconds = 30;
+        private static readonly TimeSpan MaxShutdownTime = TimeSpan.FromMinutes(1);
```
zhhyu: Why do we pick up 1 minute here?

joelverhagen: I mimicked the time we picked for our subscription processor job. This was pretty arbitrary. I wouldn't want to wait too much longer than this because this window of time is essentially downtime in the "hung" state that we've seen.

zhhyu:
I am also still a little confused about how this could solve the hang problem.
In my understanding, we create another delay task, and see which one (these scheduled workers or this delay task) will finish firstly, if the scheduled workers finish firstly, then we know that everything works fine, and nothing hangs. If the delay task finishes, we know that maybe one worker hangs. This will give us more logs/clues about the reason, but how could this fix the current problem?

joelverhagen:
> This will give us more logs/clues about the reason, but how could this fix the current problem?

First step is to know which worker is hanging. That's why I added the `workerId` log scope so that each worker's logs can be sorted. Then when we see a hang again (and care) we can look at where the logs end to understand roughly where the problem is.

Agreed that this does not fix the root cause. It simply makes it so that root cause doesn't cause an indefinite hang.

zhhyu:
> > This will give us more logs/clues about the reason, but how could this fix the current problem?
> 
> First step is to know which worker is hanging. That's why I added the `workerId` log scope so that each worker's logs can be sorted. Then when we see a hang again (and care) we can look at where the logs end to understand roughly where the problem is.
> 
> Agreed that this does not fix the root cause. It simply makes it so that root cause doesn't cause an indefinite hang.

Yeah, that makes sense.

PR #755, opened 2020-03-12
[ECR-SQL] Pass the secret injector to SQL connection factory for NgJobs. 

zhhyu:
For NgJobs (Db2Catalog, Db2Monitoring, Monitoring Processor), in order to log the on-line status of the SQL certificates and update actively, we need to pass the secret injector to the SQL connection factory, and use the caching control to update the certificates actively.
Combine with the PR - https://github.com/NuGet/ServerCommon/pull/337

Verify locally against "Db2Monitoring".


src/Ng/Jobs/NgJob.cs
```diff
@@ -58,6 +61,11 @@ public virtual string GetUsage()
             return GetUsageBase();
         }
 
+        public void SetSecretInjector(ISecretInjector secretInjector)
```
joelverhagen: Is it possible to pass this into the constructor?

zhhyu: I think about this at the first time. But then I realize that only 3 jobs need this and if we change the constructor, we have to change every job's constructor.
Not feel strong about this, If you feel that passing to the constructor is better, I will just do that.

joelverhagen: Not a big deal. This is fine if you think it's more appropriate for the situation.

src/Ng/CommandHelpers.cs
```diff
@@ -29,13 +29,20 @@ namespace Ng
 {
     public static class CommandHelpers
     {
-        public static IDictionary<string, string> GetArguments(string[] args, int start)
+        private static readonly int DefaultKeyVaultSecretCachingTimeout = 60 * 60 * 6; // 6 hours;
+        private static readonly HashSet<string> NotInjectedKeys = new HashSet<string>(StringComparer.OrdinalIgnoreCase) {
```
joelverhagen: nit: curly on the next line

loic-sharma: Also it looks like line 35, `};`, has one too many indentations.

PR #765, opened 2020-04-13
[Package Renames 1] Add popularity transfers data client

loic-sharma:
The `auxiliary2azuresearch` job needs to know which popularity transfers have changed to properly update the search index. To do this, the last seen popularity transfers data need to be persisted so that it can be compared against the latest data.

This adds new types `IPopularityTransferDataClient` and `PopularityTransferDataClient` that are unused for now. They will be added to the dependency injection container later.

Part of https://github.com/nuget/nugetgallery/issues/7898

PR #766, opened 2020-04-13
[Package Renames 2] Support popularity transfers comparisons

loic-sharma:
The `auxiliary2azuresearch` job needs to know which popularity transfers have changed to properly update the search index.

Previous change: https://github.com/NuGet/NuGet.Services.Metadata/pull/765


src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/DataSetComparer.cs
```diff
@@ -0,0 +1,145 @@
...
+            return result;
+        }
+
+        private SortedDictionary<string, string[]> CompareData(
+            SortedDictionary<string, SortedSet<string>> oldData,
+            SortedDictionary<string, SortedSet<string>> newData,
+            string keyName,
+            string valuesName,
+            StringComparer valuesComparer)
+        {
```
loic-sharma, line 71: This logic is almost the same as the previous [`OwnerSetComparer.Compare` method](https://github.com/NuGet/NuGet.Services.Metadata/blob/91b715a871b538389d3f1f14717fed9eec2dd2e4/src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/OwnerSetComparer.cs#L30).

PR #767, opened 2020-04-13
[Package Renames 3] Load popularity transfers from database

loic-sharma:
The `db2azuresearch` and `auxiliary2azuresearch` jobs needs to get the latest popularity transfers from the database.

⚠ This change is not unit testable as it interacts with a database. This will be covered by end-to-end tests and monitoring.

Previous changes: https://github.com/NuGet/NuGet.Services.Metadata/pull/765 and https://github.com/NuGet/NuGet.Services.Metadata/pull/766.

### DIFF
--- a/src/NuGet.Services.AzureSearch/DatabaseAuxiliaryDataFetcher.cs
+++ b/src/NuGet.Services.AzureSearch/DatabaseAuxiliaryDataFetcher.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Data;
 using System.Data.Entity;
 using System.Diagnostics;
 using System.Linq;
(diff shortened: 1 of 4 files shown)


src/NuGet.Services.AzureSearch/DatabaseAuxiliaryDataFetcher.cs
```diff
@@ -137,6 +152,52 @@ public async Task<HashSet<string>> GetVerifiedPackagesAsync()
...
+                            var fromId = reader.GetString(0);
+                            var toId = reader.GetString(1);
+
+                            builder.Add(fromId, toId);
+                        }
+                    }
+
+                    totalResults += currentPageResults;
+                }
+                while (currentPageResults == GetPopularityTransfersPageSize);
```
joelverhagen, line 192: In light of the recent bug, what do you think about switching to a >= 0 check instead? This mimic streams and makes fewer assumptions. Mainly a philosophical question.

loic-sharma: Writing this code definitely gave me the heebie jeebies 😅

For others that don't know the context: we recently had a bug on the Visual Studio Marketplace where we skipped extensions that we mitigated by tweaking some paging logic. The previous paging logic would fetch results until it found a batch that wasn't full. This was problematic as some batches weren't full due to extensions in invalid states (the extensions pointed to publishers that did not exist, causing these extensions to be skipped by a `JOIN`). We mitigated this by paging through results until we reach an empty batch (a >0 check).

To answer your question: the >0 check is not a perfect fix, it's a mitigation that sweeps the problem under the rug. The bug may reappear if there are more results in an invalid state than the batch size! In fact, using the >0 check may make it harder to detect that there are results in an invalid state. I would rather catch these kinds of issues quickly.

PR #768, opened 2020-04-14
[Package Renames 4] Refactor download overrides

loic-sharma:
This change introduces a new type, `IDownloadTransferrer`, which determines what downloads should be changed to reflect the latest download overrides. In the future, this type will also apply popularity transfers.

Previous changes: https://github.com/NuGet/NuGet.Services.Metadata/pull/765, https://github.com/NuGet/NuGet.Services.Metadata/pull/766, and https://github.com/NuGet/NuGet.Services.Metadata/pull/767.

## Background

Originally, auxiliary2azuresearch updated downloads with these steps:
1. Load the previously indexed downloads
1. Load the latest downloads
1. Compare the previously indexed downloads and latest downloads
1. Push the download changes to search

Later we added download overrides. The steps became:
1. Load the previously indexed downloads
1. Load the latest downloads
1. Apply the download overrides to the latest downloads
1. Compare the previously indexed downloads and latest downloads
1. Push the download changes to search

Now we want to add popularity transfers. These popularity transfers are affected by download changes and can be overriden by download overrides. The new steps will be:

1. Load the previously indexed downloads
1. Load the latest downloads
1. Compare the previously indexed downloads and latest downloads
1. Apply popularity transfers to download changes. This uses download changes to determine which popularity transfers need to be recalculated
1. Apply download overrides to download changes
1. Push the download changes to search


src/NuGet.Services.AzureSearch/DownloadTransferResult.cs
```diff
@@ -0,0 +1,40 @@
...
+            Guard.Assert(
+                downloadChanges.Comparer == StringComparer.OrdinalIgnoreCase,
+                $"Latest popularity transfers should have comparer {nameof(StringComparer.OrdinalIgnoreCase)}");
+
+            DownloadChanges = downloadChanges ?? throw new ArgumentNullException(nameof(downloadChanges));
+            LatestPopularityTransfers = popularityTransfers ?? throw new ArgumentNullException(nameof(popularityTransfers));
+        }
+
+        /// <summary>
+        /// The downloads that were changed by transferring downloads.
```
loic-sharma, line 31: TODO: This message is no longer true

src/NuGet.Services.AzureSearch/DownloadTransferResult.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace NuGet.Services.AzureSearch
+{
+    /// <summary>
+    /// The result of applying popularity transfers to download data.
```
loic-sharma, line 10: TODO: This message is no longer true

src/NuGet.Services.AzureSearch/Db2AzureSearch/NewPackageRegistrationProducer.cs
```diff
@@ -91,6 +92,11 @@ public class NewPackageRegistrationProducer : INewPackageRegistrationProducer
 
                 foreach (var pr in packageRegistrationInfo)
                 {
+                    if (!packageToDownloads.TryGetValue(pr.Id, out var packageDownloads))
+                    {
+                        packageDownloads = 0;
+                    }
```
loic-sharma, line 98: TODO: This needs a unit test

src/NuGet.Services.AzureSearch/DownloadTransferrer.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            // Downloads are transferred from a "from" package to one or more "to" packages.
+            // The "outgoingTransfers" maps "from" packages to their corresponding "to" packages.
+            _logger.LogInformation("Fetching new popularity transfer data from gallery database.");
+            var outgoingTransfers = await _databaseFetcher.GetPackageIdToPopularityTransfersAsync();
+
+            return await GetTransferChangesAsync(
+                downloads,
+                outgoingTransfers);
+        }
```
loic-sharma, line 39: `GetTransferChangesAsync` and `GetUpdatedTransferChangesAsync` are very similar right now, but they will diverge in a subsequent change.

src/NuGet.Services.AzureSearch/IDownloadTransferrer.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// Determine changes that should be applied to the latest downloads data due to popularity transfers.
+        /// </summary>
+        /// <param name="downloads">The latest downloads data.</param>
+        /// <param name="downloadChanges">The downloads that have changed since the last index.</param>
+        /// <param name="oldTransfers">The popularity transfers that were previously indexed.</param>
+        /// <returns>The result of applying popularity transfers.</returns>
+        Task<DownloadTransferResult> GetUpdatedTransferChangesAsync(
+            DownloadData downloads,
+            SortedDictionary<string, long> downloadChanges,
+            SortedDictionary<string, SortedSet<string>> oldTransfers);
```
loic-sharma, line 32: This is intentionally not using `IReadOnlyX` interfaces as this method needs to check the `.Comparer` method.

src/NuGet.Services.AzureSearch/DownloadTransferResult.cs
```diff
@@ -0,0 +1,40 @@
...
+                $"Latest popularity transfers should have comparer {nameof(StringComparer.OrdinalIgnoreCase)}");
+
+            DownloadChanges = downloadChanges ?? throw new ArgumentNullException(nameof(downloadChanges));
+            LatestPopularityTransfers = popularityTransfers ?? throw new ArgumentNullException(nameof(popularityTransfers));
+        }
+
+        /// <summary>
+        /// The downloads that should be changed due to popularity transfers.
+        /// </summary>
+        public SortedDictionary<string, long> DownloadChanges { get; }
```
joelverhagen, line 33: nit, comment what the key and value represent for both of these dictionary properties

src/NuGet.Services.AzureSearch/DownloadTransferrer.cs
```diff
@@ -0,0 +1,123 @@
...
+            IAuxiliaryFileClient auxiliaryFileClient,
+            IDatabaseAuxiliaryDataFetcher databaseFetcher,
+            ILogger<DownloadTransferrer> logger)
+        {
+            _auxiliaryFileClient = auxiliaryFileClient ?? throw new ArgumentException(nameof(auxiliaryFileClient));
+            _databaseFetcher = databaseFetcher ?? throw new ArgumentNullException(nameof(databaseFetcher));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public async Task<DownloadTransferResult> GetTransferChangesAsync(DownloadData downloads)
```
joelverhagen, line 29: What do you think about the name `InitializeTransferChangesAsync`? Seems like this should only be called in a case like Db2AzureSearch

src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/UpdateDownloadsCommand.cs
```diff
@@ -106,23 +112,26 @@ private async Task<bool> PushIndexChangesAsync()
...
-
             _logger.LogInformation("Detecting download count changes.");
-            var changes = _downloadSetComparer.Compare(oldResult.Data, overridenDownloads);
+            var changes = _downloadSetComparer.Compare(oldResult.Data, newData);
+            _logger.LogInformation("{Count} package IDs have download count changes.", changes.Count);
+
+            // The "old" data in this case is the popularity transfers data that was last indexed by this job (or
+            // initialized by Db2AzureSearch).
+            _logger.LogInformation("Fetching old popularity transfer data from blob storage.");
+            var oldTransfers = await _popularityTransferDataClient.ReadLatestIndexedAsync();
```
joelverhagen, line 128: it's a little confusing to have network IO in these methods vs. all of the network IO occurring at the top `Command` level for the other cases. consider lifting that up here if you think it might make the flow clearer.

src/NuGet.Services.AzureSearch/Db2AzureSearch/Db2AzureSearchCommand.cs
```diff
@@ -195,6 +201,12 @@ private async Task WriteVerifiedPackagesDataAsync(HashSet<string> verifiedPackag
             _logger.LogInformation("Done uploading the initial verified packages data file.");
         }
 
+        private async Task WritePopularityTransfersDataAsync(SortedDictionary<string, SortedSet<string>> popularityTransfers)
```
shishirx34, line 204: suggest moving this to a common place and call this from `auxiliary2azuresearch` TODO above... the implementation can be TODO, one less thing to worry about.

src/NuGet.Services.AzureSearch/Db2AzureSearch/NewPackageRegistrationProducer.cs
```diff
@@ -54,15 +57,13 @@ public class NewPackageRegistrationProducer : INewPackageRegistrationProducer
...
             var downloads = await _auxiliaryFileClient.LoadDownloadDataAsync();
 
-            // Fetch the download overrides from the auxiliary file. Note that the overriden downloads are kept
-            // separate from downloads data as the original data will be persisted to auxiliary data, whereas the
-            // overriden data will be persisted to Azure Search.
-            var downloadOverrides = await _auxiliaryFileClient.LoadDownloadOverridesAsync();
-            var overridenDownloads = downloads.ApplyDownloadOverrides(downloadOverrides, _logger);
+            // Apply changes from popularity transfers.
+            var transferResult = await _downloadTransferrer.GetTransferChangesAsync(downloads);
+            var packageToDownloads = GetPackageToDownloads(downloads, transferResult.DownloadChanges);
```
shishirx34, line 66: nit `packageToDownloads` doesn't make sense.. may be `updatedPackageDownloads`?

src/NuGet.Services.AzureSearch/DownloadTransferResult.cs
loic-sharma: This class was removed but I updated other comments to clarify key and values.

src/NuGet.Services.AzureSearch/DownloadTransferrer.cs
loic-sharma: I renamed this to `InitializeDownloadTransfers` and renamed the other method to `UpdateDownloadTransfers`

src/NuGet.Services.AzureSearch/Db2AzureSearch/Db2AzureSearchCommand.cs
loic-sharma: I'm not sure I understand this comment. This code path is separate from `auxiliary2azuresearch`. 

I removed this method since having an empty method is weird though

src/NuGet.Services.AzureSearch/Db2AzureSearch/NewPackageRegistrationProducer.cs
loic-sharma: Renamed to `transferredDownloads`

src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/UpdateDownloadsCommand.cs
loic-sharma: Done. This was much bigger change than I thought it'd be, but I think it turned out quite nicely. Thanks for the suggestion

src/NuGet.Services.AzureSearch/DownloadTransferrer.cs
```diff
@@ -0,0 +1,105 @@
...
+        private readonly ILogger<DownloadTransferrer> _logger;
+
+        public DownloadTransferrer(ILogger<DownloadTransferrer> logger)
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public SortedDictionary<string, long> InitializeDownloadTransfers(
+            DownloadData downloads,
+            SortedDictionary<string, SortedSet<string>> outgoingTransfers,
```
loic-sharma, line 23: TODO: Needs guard

src/NuGet.Services.AzureSearch/DownloadTransferrer.cs
```diff
@@ -0,0 +1,105 @@
...
+            ApplyDownloadOverrides(downloads, downloadOverrides, downloadTransfers);
+
+            return downloadTransfers;
+        }
+
+        public SortedDictionary<string, long> UpdateDownloadTransfers(
+            DownloadData downloads,
+            SortedDictionary<string, long> downloadChanges,
+            SortedDictionary<string, SortedSet<string>> oldTransfers,
+            SortedDictionary<string, SortedSet<string>> newTransfers,
```
loic-sharma, line 41: TODO: Needs guard

PR #769, opened 2020-04-16
[Package Renames 5] Transfer package popularity

loic-sharma:
This updates `auxiliary2azuresearch` to push download changes whenever a popularity transfer changes, or, whenever a download change affects a popularity transfer.

Previous changes: https://github.com/NuGet/NuGet.Services.Metadata/pull/765, https://github.com/NuGet/NuGet.Services.Metadata/pull/766, https://github.com/NuGet/NuGet.Services.Metadata/pull/767, and https://github.com/NuGet/NuGet.Services.Metadata/pull/768

Config changes: https://nuget.visualstudio.com/NuGetMicrosoft/_git/NuGetDeployment/pullrequest/1402?_a=overview


src/NuGet.Services.AzureSearch/DownloadTransferrer.cs
```diff
@@ -64,6 +113,143 @@ public DownloadTransferrer(ILogger<DownloadTransferrer> logger)
...
+            // Next, calculate packages with incoming transfers. These packages receive downloads
+            // from one or more "outgoing" packages.
+            if (incomingTransfers.TryGetValue(packageId, out var incomingTransferIds))
+            {
+                var result = originalDownloads;
+
+                foreach (var incomingTransferId in incomingTransferIds)
+                {
+                    var incomingDownloads = downloads.GetDownloadCount(incomingTransferId);
+                    var incomingSplit = outgoingTransfers[incomingTransferId].Count;
```
joelverhagen, line 241: Is it possible for `incomingTransferId` to not be present in `outgoingTransfers`?

tests/NuGet.Services.AzureSearch.Tests/DownloadTransferrerFacts.cs
```diff
@@ -28,6 +28,197 @@ public void ReturnsEmptyResult()
                 Assert.Empty(result);
             }
 
+            [Fact]
+            public void DoesNothingIfNoTransfers()
```
joelverhagen, line 32: Consider adding an integration test with a stateful, in-memory blob service to simulate from transitions like:

- No rename to A -> B rename
- A -> B rename to A -> C rename
- A -> B rename to B -> A rename
- A -> B rename to A -> B, C -> D rename

I think those state transitions are the most likely source for bugs.

joelverhagen: (this is in contrast to unit tests which may not cover the integration with the DAL)

tests/NuGet.Services.AzureSearch.Tests/DownloadTransferrerFacts.cs
```diff
@@ -28,6 +28,197 @@ public void ReturnsEmptyResult()
...
+                Assert.Equal(200, result["C"]);
+            }
+
+            [Fact]
+            public void RejectsCyclicalPopularityTransfers()
+            {
+                PopularityTransfer = 1;
+
+                DownloadData.SetDownloadCount("A", "1.0.0", 100);
+                DownloadData.SetDownloadCount("B", "1.0.0", 100);
```
joelverhagen, line 190: It's possible to "lose" a download, right? Like if the percentage against the total does no result in an exact number. We should have tests around that. It's okay to lose a download here or there but the behavior should be asserted IMO

zhhyu:
>         {

Should we do null-check here considering the public method?

---
Refers to: src/NuGet.Services.AzureSearch/DownloadTransferrer.cs:69 in 032b459. [](commit_id = 032b45976f6f01bc7c53e05a7d65c98688e8f628, deletion_comment = False)

src/NuGet.Services.AzureSearch/DownloadTransferrer.cs
```diff
@@ -23,9 +33,25 @@ public DownloadTransferrer(ILogger<DownloadTransferrer> logger)
             SortedDictionary<string, SortedSet<string>> outgoingTransfers,
             IReadOnlyDictionary<string, long> downloadOverrides)
         {
-            // TODO: Add download changes due to popularity transfers.
-            // See: https://github.com/NuGet/NuGetGallery/issues/7898
-            var downloadTransfers = new SortedDictionary<string, long>(StringComparer.OrdinalIgnoreCase);
+            Guard.Assert(
```
zhhyu, line 36: Should we have null-check here considering the public methods?

loic-sharma: We generally don't add null checks to our public methods.

src/NuGet.Services.AzureSearch/DownloadTransferrer.cs
loic-sharma: `incomingTransfers` is generated by inversing `outgoingTransfers`, so no, that should never happen.

tests/NuGet.Services.AzureSearch.Tests/DownloadTransferrerFacts.cs
loic-sharma: Yup, that's possible. I'll add test coverage 👍

loic-sharma: Added

zhhyu:
I am thinking that there is an edge case, but as I am not sure about the whole picture here, like how Azure Search actually works, so I just list it here:
If the original package is hard deleted, it seem that we will still keep it with the download counts 0, which will be fed to Azure Search, in my understanding. Is this expected?

joelverhagen:
> If the original package is hard deleted, it seem that we will still keep it with the download counts 0, which will be fed to Azure Search, in my understanding. ("affectedPackages" will contain this deleted package Id.) Is this expected?

Great thing to be thinking about. Given renames are on the package registration level, deletes would only cascade to the rename record if we release the package registration, I think. Given this option is unchecked by default and only available if all versions are deleted, I think it would very rarely occur. This begs the question, if all versions are deleted, should we clear the rename?

loic-sharma:
> If the original package is hard deleted, it seem that we will still keep it with the download counts 0....

Yup, you are correct that this will produce a transfer of `0`. This behavior is covered by the `DownloadTransferrerFacts.UnknownPackagesTransferZeroDownloads` test.

> ... the download counts 0, which will be fed to Azure Search

We generate a batch of "search index actions" to mutate the Azure Search index. 

The search pipeline has a "hidden" V3 resource called the version list, which is used by `catalog2azuresearch` and `auxiliary2azuresearch` to know which search documents should be updated. Using this resource, we detect and then skip updates to packages that do not exist. This behavior is covered by `SearchIndexActionBuilderFacts.UpdatesSearchDocumentsWithNoVersions`.

> This begs the question, if all versions are deleted, should we clear the rename?

The popularity transfer feature does little for packages with low download counts. I doubt we will need to clean "dangling" renames on "small" packages. Now say that we delete all versions of a popular package but keep its registration... This seems like a truly exceptional scenario, which is hard to anticipate. We may want to evaluate keeping/removing the popularity transfers on a case-by-case basis.

tests/NuGet.Services.AzureSearch.Tests/DownloadTransferrerFacts.cs
loic-sharma: Added

PR #773, opened 2020-04-24
[Package Renames] Add feature flag to disable popularity transfers

loic-sharma:
Adds the feature flag `Search.TransferPopularity` that, if disabled, removes all popularity transfers on the search index. Note that download overrides are still applied even if the popularity transfer feature flag is disabled.

Addresses https://github.com/NuGet/NuGetGallery/issues/7944
The feature flag staleness monitoring is tracked by https://github.com/NuGet/NuGetGallery/issues/7966

### DIFF
--- a/src/NuGet.Jobs.Auxiliary2AzureSearch/Job.cs
+++ b/src/NuGet.Jobs.Auxiliary2AzureSearch/Job.cs
@@ -3,6 +3,7 @@
 
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using NuGet.Jobs.Validation;
 using NuGet.Services.AzureSearch;
 using NuGet.Services.AzureSearch.Auxiliary2AzureSearch;
 using NuGet.Services.AzureSearch.AuxiliaryFiles;
@@ -12,6 +13,7 @@ namespace NuGet.Jobs
     public class Job : AzureSearchJob<Auxiliary2AzureSearchCommand>
     {
         private const string ConfigurationSectionName = "Auxiliary2AzureSearch";
+        private const string FeatureFlagConfigurationSectionName = "FeatureFlags";
 
         protected override void ConfigureJobServices(IServiceCollection services, IConfigurationRoot configurationRoot)
         {
@@ -21,6 +23,7 @@ namespace NuGet.Jobs
             services.Configure<AuxiliaryDataStorageConfiguration>(configurationRoot.GetSection(ConfigurationSectionName));
             services.Configure<AzureSearchJobConfiguration>(configurationRoot.GetSection(ConfigurationSectionName));
             services.Configure<AzureSearchConfiguration>(configurationRoot.GetSection(ConfigurationSectionName));
+            services.Configure<FeatureFlagConfiguration>(configurationRoot.GetSection(FeatureFlagConfigurationSectionName));
         }
     }
 }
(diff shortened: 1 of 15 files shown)


src/NuGet.Services.AzureSearch/Db2AzureSearch/NewPackageRegistrationProducer.cs
```diff
@@ -160,6 +163,19 @@ private bool ShouldWait(ConcurrentBag<NewPackageRegistration> allWork, bool log)
             return false;
         }
 
+        private async Task<SortedDictionary<string, SortedSet<string>>> GetPopularityTransfersAsync()
+        {
+            if (!_featureFlags.IsPopularityTransferEnabled())
+            {
+                _logger.LogWarning(
+                    "Popularity transfers feature flag is disabled. " +
+                    "All popularity transfers will be removed.");
```
loic-sharma, line 172: TODO: Update

src/NuGet.Services.AzureSearch/Auxiliary2AzureSearch/UpdateDownloadsCommand.cs
```diff
@@ -171,6 +174,19 @@ private async Task<bool> PushIndexChangesAsync()
             return true;
         }
 
+        private async Task<SortedDictionary<string, SortedSet<string>>> GetPopularityTransfersAsync()
+        {
+            if (!_featureFlags.IsPopularityTransferEnabled())
```
shishirx34, line 179: nit: personally I feel an enabled check is more readable and intuitive than not enabled..

loic-sharma: I prefer the negative pattern as it avoids nesting if there are multiple conditions that must be met before returning the result. For example:

```csharp
if (condition1)
{
    if (condition2)
    {
        return data;
    }
    else
    {
        return errorResult2;
    }
}
else
{
    return errorResult1;
}

return emptyResults;
```

Vs:

```csharp
if (!condition1)
{
    return errorResult1;
}

if (!condition2)
{
    return errorResult2;
}

return results;
```

Please let me know if you feel strongly about this.

src/NuGet.Jobs.Auxiliary2AzureSearch/Job.cs
```diff
@@ -12,6 +13,7 @@ namespace NuGet.Jobs
     public class Job : AzureSearchJob<Auxiliary2AzureSearchCommand>
     {
         private const string ConfigurationSectionName = "Auxiliary2AzureSearch";
+        private const string FeatureFlagConfigurationSectionName = "FeatureFlags";
```
joelverhagen, line 16: Thoughts on putting this in `AzureSearchJob<T>` instead?

src/NuGet.Services.V3/DependencyInjectionExtensions.cs
```diff
@@ -50,5 +60,60 @@ public static IServiceCollection AddV3(this IServiceCollection services, IDictio
...
+
+        public static void AddFeatureFlags(this ContainerBuilder containerBuilder)
+        {
+            containerBuilder
+                .Register(c =>
+                {
+                    var options = c.Resolve<IOptionsSnapshot<FeatureFlagConfiguration>>();
+                    var requestOptions = new BlobRequestOptions
+                    {
+                        ServerTimeout = TimeSpan.FromMinutes(2),
```
zhhyu, line 72: Should we make it a private const value to configure the timeout? Maybe it's fine here because we only use the value at this place.

loic-sharma: Added

src/NuGet.Jobs.Auxiliary2AzureSearch/Job.cs
loic-sharma: Done

PR #774, opened 2020-05-01
[Package Renames] Expose popularity transfers in search

loic-sharma:
This exposes the popularity transfer data on the `/search/query` endpoint. It also exposes the popularity transfer metadata on the `/search/diag` endpoint. This will later be used to monitor popularity transfer's staleness.

I also introduced a new type, `PopularityTransferData` to remove the usage of `SortedDictionary<string, SortedSet<string>>` and remove the need to check case insensitivity for popularity transfers.

Addresses https://github.com/nuget/nugetgallery/issues/7903


src/NuGet.Services.AzureSearch/AzureSearchTelemetryService.cs
```diff
@@ -403,15 +407,15 @@ public void TrackV2GetDocumentWithHijackIndex(TimeSpan elapsed)
...
+        public void TrackReadLatestVerifiedPackages(int? packageIdCount, bool modified, TimeSpan elapsed)
         {
             _telemetryClient.TrackMetric(
                 Prefix + "ReadLatestVerifiedPackagesSeconds",
                 elapsed.TotalSeconds,
                 new Dictionary<string, string>
                 {
                     { "PackageIdCount", packageIdCount?.ToString() },
-                    { "NotModified", notModified.ToString() },
+                    { "Modified", modified.ToString() },
```
loic-sharma, line 418: This property's name was incorrect. See:

https://github.com/NuGet/NuGet.Services.Metadata/blob/64af0b59c5a79e0143f0808b39946df9f16cb2e7/src/NuGet.Services.AzureSearch/AuxiliaryFiles/VerifiedPackagesDataClient.cs#L55-L80

src/NuGet.Services.AzureSearch/SearchService/AuxiliaryData.cs
```diff
@@ -3,27 +3,35 @@
...
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using NuGet.Services.AzureSearch.AuxiliaryFiles;
 
 namespace NuGet.Services.AzureSearch.SearchService
 {
     public class AuxiliaryData : IAuxiliaryData
     {
+        private static readonly string[] EmptyStringArray = new string[0];
```
joelverhagen, line 13: nit: I discovered `Array.Empty<T>()` recently. Seems nice.

loic-sharma: Nice find, I like removing boilerplate :D

PR #777, opened 2020-05-05
V3 package source for tests

agr:


tests/NuGet.Config
```diff
@@ -8,10 +8,14 @@
     <add key="automatic" value="True" />
   </packageRestore>
   <packageSources>
-    <add key="v2.nuget.org" value="https://nuget.org/api/v2" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
```
joelverhagen, line 11: We should have clear here too.

tests/NuGet.Config
```diff
@@ -8,10 +8,14 @@
...
     <add key="nugetbuild" value="https://dotnet.myget.org/F/nuget-build/api/v2/" />
   </packageSources>
   <activePackageSource>
     <add key="All" value="(Aggregate source)" />
   </activePackageSource>
+  <disabledPackageSources>
+    <clear />
+    <add key="https://www.nuget.org/api/v2/" value="true" />
```
joelverhagen, line 19: Why disable this?

tests/NuGet.Config
agr: Added

tests/NuGet.Config
agr: Copied from other NuGet.configs we have here and there.

tests/NuGet.Config
```diff
@@ -8,10 +8,14 @@
     <add key="automatic" value="True" />
   </packageRestore>
   <packageSources>
-    <add key="v2.nuget.org" value="https://nuget.org/api/v2" />
+    <clear />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
     <add key="nugetbuild" value="https://dotnet.myget.org/F/nuget-build/api/v2/" />
```
joelverhagen, line 13: nit: v3 here too 🥺 ?

joelverhagen: Not a big deal, just if you feel like it.

PR #783, opened 2020-05-18
[Popularity transfers] Disable download overrides if popularity transfers are enabled

loic-sharma:
This change disables download overrides if popularity transfers are enabled.

### Why?

Today download overrides are applied on top of popularity transfers. These overrides should only be applied if they were greater than the packages' downloads, after transfers. However, this logic would incorrectly override a package's downloads if we did not its recalculate transferred downloads. Here's an example of the bug:

1. We override `New`'s downloads from 0 downloads to 5 million
1. We transfer popularity from `Old`, with 500 million downloads, to `New`.
1. We run `db2azuresearch`.
    1. The package `Old` transfers 495 million downloads to `New`.
    1. We check if we should apply the 5 million download override. Since 495 million is greater than 5 million, we skip it.
1. We run `auxiliary2azuresearch`
    1. There are no popularity transfer changes so we don't calculate any downloads after transfers.
    1. We check if we should apply the 5 million download overrides. We haven't recalculated transfers, so at this point it appears like `New` has 0 downloads. We incorrectly apply the download override :'(

Adding logic to apply download overrides correctly is tricky (to do this properly we'd need to track download override changes). Popularity transfers are meant to replace download overrides, and there's no scenario where we need both transfers and overrides. Disabling overrides when transfers are enabled is the easiest solution.


loic-sharma:
Removing package overrides is tracked here: https://github.com/nuget/engineering/issues/3089

PR #785, opened 2020-06-17
Retry once on certain WebException codes

joelverhagen:
Address https://github.com/NuGet/Engineering/issues/3212
Only added retry for WebException codes that were clearly transient in logs

### DIFF
--- a/src/NuGet.Services.AzureSearch/DependencyInjectionExtensions.cs
+++ b/src/NuGet.Services.AzureSearch/DependencyInjectionExtensions.cs
@@ -10,6 +10,7 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Microsoft.Rest;
+using Microsoft.Rest.TransientFaultHandling;
 using Microsoft.WindowsAzure.Storage;
 using NuGet.Protocol;
 using NuGet.Protocol.Catalog;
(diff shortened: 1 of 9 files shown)


src/NuGet.Services.AzureSearch/DependencyInjectionExtensions.cs
```diff
@@ -276,12 +290,33 @@ private static void RegisterAuxiliaryDataStorageServices(ContainerBuilder contai
...
+
+        /// <summary>
+        /// Defaults originally taken from:
+        /// https://github.com/Azure/azure-sdk-for-net/blob/96421089bc26198098f320ea50e0208e98376956/sdk/mgmtcommon/ClientRuntime/ClientRuntime/RetryDelegatingHandler.cs#L19-L22
+        /// </summary>
+        private static RetryPolicy GetSearchRetryPolicy()
+        {
+            return new RetryPolicy(
+                new HttpStatusCodeErrorDetectionStrategy(),
+                retryCount: 3,
```
skofman1, line 308: You retry once in WebExceptionRetryDelegatingHandler  and here the retryCount is 3. Does this mean we are going to make 6 requests to Azure Search? What am I missing.

src/NuGet.Services.AzureSearch/WebExceptionRetryDelegatingHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+    public class WebExceptionRetryDelegatingHandler : DelegatingHandler
+    {
+        private readonly ILogger<WebExceptionRetryDelegatingHandler> _logger;
+
+        public WebExceptionRetryDelegatingHandler(ILogger<WebExceptionRetryDelegatingHandler> logger)
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public static HashSet<WebExceptionStatus> TransientWebExceptionStatuses => new HashSet<WebExceptionStatus>(new[]
```
loic-sharma, line 23: Does this need to be public? It looks like your tests have a copy of this. Also can this be a static field instead of a property (to avoid allocating on each access)?

joelverhagen: Yup, good catch. This should be a `static` and either a `readonly` property or field. Field is easier for `private` so I will go with that.

src/NuGet.Services.AzureSearch/DependencyInjectionExtensions.cs
joelverhagen: This retry policy only applies to the retry logic provided by the Azure Search SDK which is based on HTTP status codes. This is the existing policy, but constructed explicitly for clarity.

This retry policy has nothing to do with `WebExceptionRetryDelegatingHandler`.

The layers are:

```
Customer
-> network
-> Search Service
-> Azure Search SDK
-> HttpClient
-> WebExceptionRetryDelegatingHandler (does not retry for HTTP status codes)
-> RetryDelegatingHandler             (retries for a select few HTTP status codes, see tests)
-> RetryAfterDelegatingHandler        (retries for 429 with a Retry-After header)
-> network
-> Azure Search
```

`WebExceptionRetryDelegatingHandler` will only retry once on specific `WebException` so the only way to get more than 3 total requests is if an individual search request encounters both connection errors and HTTP errors. For example, if Azure Search returns this sequence for a single customer request: 500, 500, 500, connection timeout, 500, 500, 500. This would result in the final 500 bubbling out. But from the data I have not seen mixtures of failure cases so I don't think it's worth complicating the implementation. The Azure Search SDK is hard-coded in `RetryDelegatingHandler` to not retry on `WebException` cases so we would need to re-implement `RetryDelegatingHandler` if we wanted to cover this case. I didn't think it was worth it. The cases we have seen have been only a `WebException`, not failure HTTP status codes.

skofman1: Got it. So we might get into 6 requests, but it's extremely unlikely. Thanks for the explanation!

PR #786, opened 2020-06-18
Catalog2icon to ignore missing packages

agr:
Addresses https://github.com/NuGet/Engineering/issues/3100.

The issue manifests itself when someone uploads a package with embedded icon and then it gets deleted before catalog2icon has a chance to process the initial catalog entry, it will fail trying to access the package contents that is now deleted.

In the worst case the catalog entries about package creation and deletion are going to end up in separate pages and since catalog2icon processes catalog page by page, we can't reliably determine that we are in this situation just by observing the contents of a single page. However, assuming that the packages container used as a package source functions correctly (otherwise, we'll notice it through catalog2dnx monitoring), we can simply treat the absence of a package in that container as an indication that the package was later deleted and simply skip processing current catalog entry.
Looking at the logs for the past 3 months, that specific exception (404 while trying to get package contents) happened only twice, when the issue reproduced.

Added exception logging for exceptions occur while accessing the package contents to keep proper logging scope for easier tracking the cause.


src/Catalog/Icons/CatalogLeafDataProcessor.cs
```diff
@@ -170,17 +173,40 @@ private async Task ProcessEmbeddedIconAsync(IStorage destinationStorage, Catalog
...
-                    targetStoragePath,
-                    cancellationToken,
-                    item.PackageIdentity.Id,
-                    item.PackageIdentity.Version.ToNormalizedString());
+                Stream packageStream;
+                try
+                {
+                    packageStream = await packageBlobReference.GetStreamAsync(cancellationToken);
+                }
+                catch (StorageException ex) when (ex.RequestInformation.HttpStatusCode == 404)
```
loic-sharma, line 182: Nit: consider using `HttpStatusCode.NotFound` instead of `404`. Just a thought, I don't feel strongly about this.

PR #787, opened 2020-06-18
Add sortable attribute to TotalDownloadCount field

t-mog-msft:
Making the TotalDownloadCount field sortable will help us to implement the sorting by total download count as part of the Advanced search epic


joelverhagen:
Have you tested this with a) a real Azure Search resource and b) a query that sorts by this field?

t-mog-msft:
> Have you tested this with a) a real Azure Search resource and b) a query that sorts by this field?

I tested it with an index I created on Azure and it worked. So I guess yes 😄

joelverhagen:
Sounds good. :shipit:

PR #788, opened 2020-06-19
Add BatchPusherResult to surface package IDs that failed

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/8051. This does not fix the issue but lays the groundwork for callers to retry in this case. The plan is for the `Auxiliary2AzureSearch` `UpdateDownloadsCommand` to be clever (only retry failed IDs) and everything else to just retry everything (easier to implement). The download update is the longest running use of `BatchPusher` so that's why I picked that.

After this PR, jobs will still crash from HTTP 412. This PR is just setting up for future changes. I don't want the PRs to be too big.

### DIFF
--- a/tests/NuGet.Services.AzureSearch.Tests/Auxiliary2AzureSearch/UpdateDownloadsCommandFacts.cs
+++ b/tests/NuGet.Services.AzureSearch.Tests/Auxiliary2AzureSearch/UpdateDownloadsCommandFacts.cs
@@ -38,8 +38,8 @@ namespace NuGet.Services.AzureSearch.Auxiliary2AzureSearch
                         It.IsAny<string>(),
                         It.IsAny<Func<SearchFilters, KeyedDocument>>()),
                     Times.Never);
-                BatchPusher.Verify(x => x.FinishAsync(), Times.Never);
-                BatchPusher.Verify(x => x.PushFullBatchesAsync(), Times.Never);
+                BatchPusher.Verify(x => x.TryFinishAsync(), Times.Never);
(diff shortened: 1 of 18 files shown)


tests/NuGet.Services.AzureSearch.Tests/Auxiliary2AzureSearch/UpdateDownloadsCommandFacts.cs
```diff
@@ -522,8 +522,12 @@ public Facts(ITestOutputHelper output)
                         PushedIds.Add(id);
                     });
                 BatchPusher
-                    .Setup(x => x.FinishAsync())
-                    .Returns(Task.Delay(TimeSpan.FromMilliseconds(1)))
+                    .Setup(x => x.TryFinishAsync())
+                    .Returns(async () =>
+                    {
+                        await Task.Delay(TimeSpan.FromMilliseconds(1));
```
loic-sharma, line 528: What's the sleep for? Could you instead return a `Task.FromResult(...)`?

joelverhagen: Pre-existing. Intention is to have a bit of time to assert against in a duration metric.

loic-sharma: Oh there's a comment on line 513 explaining this. My bad :)

PR #790, opened 2020-06-22
Add vulnerability details to catalog

drewgillies:
Extending query to include vulnerability details in package/version rows. Multiple vulnerabilities will mean multiple rows for a given package/version, so I've changed the reader processing to use a control break for when next package key encountered. Note that this avoids the need for secondary queries for vulnerabilties, but does require reimagining the read loop. 

Results in an entry like this in a catalog package json:

```
  "vulnerabilities": [
    {
      "@id": "https://apidev.nugettest.org/v3/catalog0/data/2020.06.26.05.44.09/dotnetzip.1.9.1.8.json#GitHub-700",
      "@type": "Vulnerability",
      "advisoryUrl": "https://nvd.nist.gov/vuln/detail/CVE-2018-1002205",
      "gitHubDatabaseKey": "GitHub-700",
      "severity": "2"
    }
  ],
```

Note that gitHubDatabaseKey is used (coupled with package URI) as a unique identifier for the vulnerability. It's prefixed with "GitHub-" in case we wish to add other database sources in future (we may be stuck with the column name in the table though, but that's not a big deal).

Even though there are 40 files in this PR, note that a bunch are adjustments to test data json files (adding vulnerabilities to the context). The salient tests are in CatalogWriterHelperTests where I've added vulnerability json to the expected (in CatalogTestData) and actual outputs.

The main logic is in GalleryDatabaseQueryService and Db2CatalogProjection, where the SQL query is altered, the reader loop is adjusted for multiple row per package/version processing, and vulnerability columns are read. Most of the rest is plumbing, and the building of the IGraph in PackageCatalogItem.


src/Catalog/PackageCatalogItem.cs
```diff
@@ -103,23 +107,23 @@ public override IGraph CreateContentGraph(CatalogContext context)
             //  identity and version
             SetIdVersionFromGraph(graph);
 
-            // deprecation
+            //  deprecation
             if (Deprecation != null)
             {
-                // assert deprecation root node to subject
+                //  assert deprecation root node to subject
```
joelverhagen, line 113: nit: double space after `//` is nonstandard

joelverhagen:
For the `gitHubDatabaseKey` property, I don't think we need the word `GitHub` prefixed since it is already in the property name. For the URI fragment, we need it since a bare number like `#700` is ambiguous. I think the URI should include the word `vulnerability`. Consider dependency groups:
image

Maybe a fragment like `#vulnerability/GitHub/700`?

src/Catalog/Helpers/GalleryDatabaseQueryService.cs
```diff
@@ -178,7 +179,7 @@ internal static string BuildDb2CatalogSqlQuery(Db2CatalogCursor cursor)
             return $@"SELECT TOP {cursor.Top} WITH TIES
```
joelverhagen, line 179: won't this `TOP` value be a little weird now? Like if TOP is 20 today and one package has 10 vulnerabilities, won't this only return 11 unique packages now (but still 20 rows)?

src/Catalog/Helpers/Db2CatalogProjection.cs
```diff
@@ -81,7 +103,7 @@ public PackageDeprecationItem ReadDeprecationInfoFromDataReader(DbDataReader dat
 
                 if (deprecationStatus.HasFlag(deprecationStatusFlag))
                 {
-                    deprecationReasons.Add(deprecationStatusFlag.ToString());
+                    deprecationReasons.Add(deprecationStatusFlag.ToString());
```
loic-sharma, line 106: Nit: trailing whitespace

```suggestion
                    deprecationReasons.Add(deprecationStatusFlag.ToString());
```

src/Catalog/Helpers/GalleryDatabaseQueryService.cs
```diff
@@ -223,13 +223,49 @@ private async Task<IReadOnlyCollection<FeedPackageDetails>> GetPackages(Db2Catal
 
                 using (_telemetryService.TrackGetPackageDetailsQueryDuration(cursor))
                 {
-                    using (var packagesReader = await packagesCommand.ExecuteReaderAsync())
+                    return await ReadPackages(packagesCommand);
+                }
+            }
+        }
+
+        private async Task<IReadOnlyCollection<FeedPackageDetails>> ReadPackages(SqlCommand packagesCommand)
```
loic-sharma, line 231: Nit: our style is to always add an `Async` suffix to async methods.

```suggestion
        private async Task<IReadOnlyCollection<FeedPackageDetails>> ReadPackagesAsync(SqlCommand packagesCommand)
```

joelverhagen:
How will we get all of the existing package vulnerabilities into the catalog? Presumably the data is already in the DB and the last edited timestamp has already gone by. Is there an ops task to reflow all affected packages? If the number is smallish, we have a script to do this simply by scripting the reflow link in Gallery UI.

src/Catalog/PackageCatalogItem.cs
```diff
@@ -130,31 +134,54 @@ public override IGraph CreateContentGraph(CatalogContext context)
...
+            {
+                INode vulnerabilityPredicate = graph.CreateUriNode(Schema.Predicates.Vulnerability);
+                INode vulnerabilityType = graph.CreateUriNode(Schema.DataTypes.Vulnerability);
+                INode gitHubDatabaseKeyPredicate = graph.CreateUriNode(Schema.Predicates.GitHubDatabaseKey);
+                INode advisoryUrlPredicate = graph.CreateUriNode(Schema.Predicates.AdvisoryUrl);
+                INode severityPredicate = graph.CreateUriNode(Schema.Predicates.Severity);
+
+                foreach (PackageVulnerabilityItem vulnerability in Vulnerabilities)
+                {
+                    Uri vulnerabilityUri = new Uri(resource.Subject.ToString() + "#" + vulnerability.GitHubDatabaseKey);
```
joelverhagen, line 167: What is the type of `Subject`? `Uri`? If so `ToString()` should not be used. Instead, `AbsoluteUri` should be used. Alternatively, you can use `UriBuilder` to ensure we don't concat a fragment after another fragment.

src/Catalog/Helpers/GalleryDatabaseQueryService.cs
joelverhagen: BTW I might be totally wrong, I have not used `WITH TIES` much

joelverhagen:
I think the catalog docs should be updated in light of this.
https://docs.microsoft.com/en-us/nuget/api/catalog-resource

It's a separate PR to https://github.com/nuget/docs.microsoft.com-nuget/. @ryuyu recently did this for package types and can help you out.

src/Catalog/PackageCatalogItem.cs
drewgillies: I was making the comments in the file consistent (see others that pre-date this change), but I can change them all to 1 space, sure.

src/Catalog/Helpers/GalleryDatabaseQueryService.cs
drewgillies: It's certainly something I should investigate.

src/Catalog/PackageCatalogItem.cs
drewgillies: See https://github.com/NuGet/NuGet.Services.Metadata/blob/e4b4b1392492a04a9e3b0cf72bb04a303338ede1/src/Catalog/PackageCatalogItem.cs#L89. I can change both to `AbsoluteUri`, but if the resultant string from both approaches is different, could we possibly pollute data in the PackageEntry collections by changing this now?

drewgillies:
> For the gitHubDatabaseKey property, I don't think we need the word GitHub prefixed
@joelverhagen what I'm trying to provide for here is the likelihood of another source in the future. What if we start to egress from data-base `FooHub` as well as `GitHub`? Would we want to add a new field called `fooHubDatabaseKey` or would we want to add some other differentiator into the `gitHubDatabaseKey` value? It seems that the earlier GitHub2Db work having a column with `gitHub` in the name may (1) not have been taking possible future sources into account, (2) been taking them into account by planning for other sources to use new columns (which I wouldn't think is great for scaling), or (3) been taking a we'll-cross-that-bridge-when-we-come-to-it approach. I'm favoring (3) here, thinking that we'll put `FooHub-` in the value to distinguish. Personally, I'd like the column to have been named ThirdPartyDatabaseExternalKey and have an additional column ThirdPartyDatabaseId/Key (which a possible FK to a VulnerabilitiesDatabases table) or the like. Thoughts?

> How will we get all of the existing package vulnerabilities into the catalog?
I'm expecting we will need to run a one-off job to update LastEdited to >cursor date for all `Packages` rows with associated vulnerability information.

src/Catalog/PackageCatalogItem.cs
joelverhagen: Given the current pattern of URLs the behavior of `ToString()` should match `AbsoluteUri` in the vast majority of cases. Suspicious cases are when the ID has non-ASCII or non-URL safe characters. There are known problems with `Uri.ToString()` such as decoding URL characters for display purposes but incorrectly for data purposes. See this blog for more details: https://dhvik.blogspot.com/2019/12/uritostring-automatically-decodes-url.html

There is no known consumer of package entries so the risk is very low. If you are concerned have the bandwidth, sample a couple of cases to see if anything jumps out.

If you want to avoid the analysis, you can make yours correct and file a bug on the other.

joelverhagen:
> Thoughts?

IIRC the plan was to add columns (2). This is the easiest to implement with EF and can work for quite a while before not scaling well. I think the likelihood for 5+ sources of vulnerabilities is slim. (3) is always an option in that we can say (2) now and do something else later if the details are different.

src/Catalog/Helpers/GalleryDatabaseQueryService.cs
drewgillies: @joelverhagen Here's how TIES seems to operate--SELECT TOP 550 (this is the constant we use--MaxPageSize) will give the top 550 rows based on ordering. WITH TIES will add rows after the 550th row whose ordinal columns equal those of row 550. So, your concern about the 10-vulnerability scenario is founded--you would only get 541 package versions in that case (we don't actually have a situation yet where a  package version has >1 vulnerabilities but we should design for it).

Prior to this change we were ordering only by LastEdited, which means TIES would be rows after row n (e.g. n=550) whose LastEdited value equals that of row n.

This change introduces the package key into the ordering, which means that if we have:
row 549:     LastEdited: '2020-06-30 00:30:34.986', Key: 101, Vulnerability: 200
row 550:     LastEdited: '2020-06-30 00:30:34.986', Key: 101, Vulnerability: 453
row 551:     LastEdited: '2020-06-30 00:30:34.986', Key: 101, Vulnerability: 664
...without WITH TIES we run the risk of splitting a single package version catalog entry into two, if this series is broken at max page size n like above. WITH TIES will include row 551 in our set. When we build our catalog entry, of course, these 3 rows become a single json blob.

Really what we have here is a decision to make:
Option A: MaxPageSize will continue to operate as a maximum, and for every package version in the dataset which has i>1 vulnerabilities, our page size will diminish by i-1 entries. We could decide that this is good enough, allowing MaxPageSize to be our safety net but to no longer create uniform page sizes.

Option B: If we do want MaxPageSize to act as an exact requirement on page size (except of course when there are fewer available), our query here may need two passes:
- the first would use TOP, grouping all vulnerability key values, giving us a dataset containing only 1 row per package version (unique on package.key). We'd store this dataset in a temporary table
- the second being a >cursor only (no TOP) and inner joining with the temp table to give us our final set.
Another possibility is we do this as a single query, grouping again on package.key and aggregating vulnerabilty data into delimited strings which we then process. This seems the better idea, actually.

So, I'm all ears on opinions re this decision. I personally would expect that option A is the better course of action. What do you think?

drewgillies: /cc @agr @ryuyu ^^^

src/Catalog/PackageCatalogItem.cs
drewgillies: @joelverhagen Did a little research, and the type is `VDS.RDF.UriNode`, not `Uri`, and its `ToString()` method does indeed return `AbsoluteUri`, as we can see in its base class here:
https://github.com/dotnetrdf/dotnetrdf/blob/1e3a6228eea930bb2d7a2b0fb2ce47eadb71d4a5/Libraries/dotNetRDF/Core/URINode.cs#L262
So we're good to go on both of these counts.

src/Catalog/Helpers/GalleryDatabaseQueryService.cs
drewgillies: @joelverhagen Have taken another look and our cursor size is 20 (from the config). I'm going to revisit this in a bit. I may go for the aggregated string version of option B above.

src/Catalog/PackageCatalogItem.cs
joelverhagen: Awesome, thanks for investigating this!

src/Catalog/Helpers/GalleryDatabaseQueryService.cs
drewgillies: After some perf analysis of 
1. GROUP BY all fields in original query to aggregate vulnerability fields into delimited strings for later processing (giving us 1 row per package), 
2. GROUP BY key field only, taking MAX of each original field and aggregating vulnerabilty fields as in 1.,
3. moving existing query into an inner query to preserve TOP count, and then joining to vulnerability tables and utilising existing control break logic to process additional rows,
...3 is the clear winner, so I have implemented this approach.

drewgillies:
All feedback is addressed (separate commit for now). Thank you! 
@joelverhagen @loic-sharma

PR #791, opened 2020-06-23
Add retries to jobs that use BatchPusher

joelverhagen:
This allows 412s to be handled gracefully in most cases.
Address https://github.com/NuGet/NuGetGallery/issues/8051
Follow-up to https://github.com/NuGet/NuGet.Services.Metadata/pull/788

### DIFF
--- a/src/NuGet.Services.AzureSearch/BatchPusherResult.cs
+++ b/src/NuGet.Services.AzureSearch/BatchPusherResult.cs
@@ -9,13 +9,13 @@ namespace NuGet.Services.AzureSearch
 {
     public class BatchPusherResult
     {
-        public BatchPusherResult() : this(new HashSet<string>())
+        public BatchPusherResult() : this(Array.Empty<string>())
         {
         }
 
-        public BatchPusherResult(HashSet<string> failedPackageIds)
+        public BatchPusherResult(IEnumerable<string> failedPackageIds)
         {
-            FailedPackageIds = failedPackageIds;
+            FailedPackageIds = failedPackageIds.ToHashSet(StringComparer.OrdinalIgnoreCase);
         }
 
         /// <summary>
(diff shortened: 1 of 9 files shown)


src/NuGet.Services.AzureSearch/BatchPusherResult.cs
```diff
@@ -9,28 +9,37 @@ namespace NuGet.Services.AzureSearch
...
-                    + string.Join(", ", FailedPackageIds));
+                    + string.Join(", ", FailedPackageIds.OrderBy(x => x, StringComparer.OrdinalIgnoreCase)));
             }
         }
+
+        public BatchPusherResult Merge(BatchPusherResult other)
+        {
+            return new BatchPusherResult(FailedPackageIds
+                .Concat(other.FailedPackageIds)
+                .ToHashSet(FailedPackageIds.Comparer));
```
loic-sharma, line 42: Nit: is there a benefit to not be explicit here? As in, `.ToHashSet(StringComparer.OrdinalIgnoreCase));`? I don't feel strongly about this, feel free to skip if you prefer as-is.

src/NuGet.Services.AzureSearch/Catalog2AzureSearch/AzureSearchCollectorLogic.cs
```diff
@@ -85,21 +105,45 @@ private async Task ProcessItemsAsync(IEnumerable<CatalogCommitItem> items, bool
...
+            public ProcessItemsResult(bool success, List<CatalogCommitItem> items)
+            {
+                Success = success;
+                Items = items ?? throw new ArgumentNullException(nameof(items));
+            }
+
+            public bool Success { get; }
+
+            /// <summary>
+            /// The item list to used for the next ieration.
```
loic-sharma, line 142:
```suggestion
            /// The item list to used for the next iteration.
```

src/NuGet.Services.AzureSearch/BatchPusherResult.cs
joelverhagen: I was thinking a generic merge method should defer as much logic as possible and not "own" it's own logic if it's already defined elsewhere but sure I can change it

joelverhagen: I've thought many times about making a strong type for package ID but that's probably too prollerhead

PR #793, opened 2020-06-26
[Search] Add functional tests for the tags custom analyzer

loic-sharma:
This adds functional tests for the new custom analyzer for packages' tags.

/cc @khalidabuhakmeh


tests/BasicSearchTests.FunctionalTests.Core/BaseFunctionalTests.cs
```diff
@@ -21,6 +21,7 @@ public BaseFunctionalTests()
         public BaseFunctionalTests(string baseUrl)
         {
             // Arrange
+            ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12;
```
loic-sharma, line 24: This is now necessary as otherwise the Azure Search API will close the connection.

joelverhagen: Please do `|=`

loic-sharma: Oh whoops. Nice catch

PR #794, opened 2020-06-29
[Advanced Search] Add sorting & filtering to the search service

t-mog-msft:
This PR adds the sorting by TotalDownloadCount feature to the internal API endpoint (V2) of the search service.

Both totalDownloads-asc and totalDownloads-desc values for the "orderBy" parameter are supported. We also make sure to sort the data on the download count available from the Auxiliary file (the same data shown to users on the Gallery side) to avoid sorting bugs.

Some functional tests were changed to easily make them run locally (i.e: without having to upload "json" packages and simply use those available).

P.S: This PR is going to the "mog-advanced-search" branch which will be later merged to "dev" in another PR


src/NuGet.Services.AzureSearch/SearchService/SearchParametersBuilder.cs
```diff
@@ -34,6 +34,8 @@ public class SearchParametersBuilder : ISearchParametersBuilder
         private static readonly List<string> SortableTitleDesc = new List<string> { IndexFields.SortableTitle + Desc, IndexFields.Created + Desc }; // Last title by lex order first, then oldest
         private static readonly List<string> CreatedAsc = new List<string> { IndexFields.Created + Asc }; // Newest first
         private static readonly List<string> CreatedDesc = new List<string> { IndexFields.Created + Desc }; // Oldest first
+        private static readonly List<string> TotalDownloadsAsc = new List<string> { IndexFields.Search.TotalDownloadCount + Asc }; // Least downloads first
```
loic-sharma, line 37:
```suggestion
        private static readonly List<string> TotalDownloadsAsc = new List<string> { IndexFields.Search.TotalDownloadCount + Asc }; // Fewest downloads first
```

src/NuGet.Services.AzureSearch/SearchService/SearchResponseBuilder.cs
```diff
@@ -323,16 +323,27 @@ private static string TitleThenId(IBaseMetadataDocument document)
...
             else
             {
+                var resultData = results.Select(x =>
+                {
+                    var package = toPackage(x.Document);
+                    package.Debug = request.ShowDebug ? x : null;
+                    return package;
+                });
+
+                // We sort here to make sure to sort on the same data displayed to the user (i.e: to avoid having inconsistencies in the displayed order of the packages)
```
loic-sharma, line 333: I would mention that the real sorting is done by Azure Search. However, the downloads data we sort on (the Azure Search index's downloads) is different from the downloads data we display (the downloads auxiliary file), so we do another round of sorting here to remove inconsistencies. I would also mention that the other sort orders don't have this problem and don't need sorting here.

src/NuGet.Services.AzureSearch/SearchService/SearchParametersBuilder.cs
joelverhagen, line 37: Consider sorting by `Created` as a second level of sorting so that you have more stability. A similar approach was taking on the other sort-by. We did this because Azure Search has undefined order when the a sort-by is tied (internally I think it is by document ID which is nondeterministic and can vary from query to query due to sharding)

src/NuGet.Services.SearchService/Controllers/SearchController.cs
```diff
@@ -22,12 +22,15 @@ public class SearchController : ApiController
 
         private static readonly IReadOnlyDictionary<string, V2SortBy> SortBy = new Dictionary<string, V2SortBy>(StringComparer.OrdinalIgnoreCase)
         {
+            { "relevance", V2SortBy.Popularity},
```
joelverhagen, line 25:
```suggestion
            { "relevance", V2SortBy.Popularity },
```

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchParametersBuilderFacts.cs
```diff
@@ -179,7 +180,11 @@ public void AllSortByFieldsAreSortable(V2SortBy v2SortBy)
             {
                 var metadataProperties = typeof(BaseMetadataDocument)
                     .GetProperties()
+                    .Union(typeof(SearchDocument.Full).GetProperties()) // Properties can also be in a SearchDocument (e.g: TotalDownloadCount)
```
loic-sharma, line 183: Nice 👍

src/NuGet.Services.SearchService/Controllers/SearchController.cs
joelverhagen: IIRC this is not required due to the `switch` statement but perhaps this improves readability. The logic is now duplicated though.

tests/NuGet.Services.SearchService.Tests/Controllers/SearchControllerFacts.cs
```diff
@@ -243,6 +243,10 @@ public async Task UsesProvidedParameters()
             [InlineData("Created-asc", V2SortBy.CreatedAsc)]
             [InlineData("CREATED-desc", V2SortBy.CreatedDesc)]
             [InlineData("Created-desc", V2SortBy.CreatedDesc)]
+            [InlineData("totalDownloads", V2SortBy.Popularity)]
+            [InlineData("totalDownloads-asc", V2SortBy.TotalDownloadsAsc)]
+            [InlineData("totalDownloads-desc", V2SortBy.TotalDownloadsDesc)]
+            [InlineData("TotalDownloads-desc", V2SortBy.TotalDownloadsDesc)]
```
loic-sharma, line 249: Looks like line 248 and 249 are duplicated

loic-sharma: Oh never mind, this is a case sensitivity test. I'm blind :)

src/NuGet.Services.AzureSearch/SearchService/SearchParametersBuilder.cs
loic-sharma: If you add this, you'll need to update the in-service sorting as well

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchResponseBuilderFacts.cs
```diff
@@ -357,6 +360,145 @@ public void UsesVersionSpecificDownloadCount()
                 Assert.Equal(4, response.Data[0].DownloadCount);
             }
 
+            [Fact]
+            public void SortDownloadsAscUsingAuxilaryFile()
```
loic-sharma, line 364: These tests a little bit long because of the data setup. I would recommend creating a new field, like `_fiveSearchResults`, that is created in the constructor similar to how `_searchResult` and `_manySearchResults` are created. 

I would also recommend building each result by using `Data.SearchDocument` and then modifying its fields. Note that `Data.SearchDocument` creates a new document each time it is accessed. For example:

```cs
var document1 = Data.SearchDocument;
document1.PackageId = "Hello world";
document1.TotalDownloadCount = 123
```

src/NuGet.Services.SearchService/Controllers/SearchController.cs
t-mog-msft: Yeah this was to improve readability since it took me by surprise in the beginning when I was trying to look for orderBy relevance and it had the "same" behaviour as orderBy "ajnhjfkbashsfbhsbf"

src/NuGet.Jobs.Db2AzureSearch/settings.json
```diff
@@ -0,0 +1,52 @@
+{
```
loic-sharma, line 1: Please remove this file

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchResponseBuilderFacts.cs
```diff
@@ -357,6 +360,88 @@ public void UsesVersionSpecificDownloadCount()
...
+            {
+                var mockAuxilaryDataDownloads = new Dictionary<string, long>();
+                var rnd = new Random();
+                for (int i = 0; i < _fiveSearchResults.Count; ++i)
+                {
+                    var newDoc = _fiveSearchResults.Results[i];
+                    newDoc.Document.PackageId += i;
+                    newDoc.Document.TotalDownloadCount += i;
+                    newDoc.Document.Title += i;
```
loic-sharma, line 373: I would consider lines 371-373 and 411-414 to the constructor as well since it's duplicated.

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchResponseBuilderFacts.cs
```diff
@@ -357,6 +360,88 @@ public void UsesVersionSpecificDownloadCount()
...
+
+                    // Set random download count in the Auxilary file for this package
+                    mockAuxilaryDataDownloads.Add(newDoc.Document.PackageId, rnd.Next(0, int.MaxValue));
```
loic-sharma, line 376: I think using random here will work as expected, but it scares me a little. Using random values means that these tests are not deterministic and may be flaky. Do you think that the randomness here covers more cases than having a deterministic test? If so, could we instead add these cases manually such that they are deterministic?

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchResponseBuilderFacts.cs
```diff
@@ -357,6 +360,88 @@ public void UsesVersionSpecificDownloadCount()
...
+                    .Results
+                    .OrderBy(x => mockAuxilaryDataDownloads[x.Document.PackageId])
+                    .ThenBy(x => x.Document.Created)
+                    .Select(x => x.Document.PackageId);
+
+                _auxiliaryData
+                    .Setup(x => x.GetTotalDownloadCount(It.IsAny<string>()))
+                    .Returns((string packageId) => {
+                        return mockAuxilaryDataDownloads[packageId];
+                    });
```
loic-sharma, line 389: Nit: C# style is to put `{` on a new line. In this case, you can shorten this up a little bit

```suggestion
                    .Returns(packageId => mockAuxilaryDataDownloads[packageId]);
```

loic-sharma: Same comment for lines 428-431

src/NuGet.Services.SearchService/Controllers/SearchController.cs
```diff
@@ -81,7 +84,8 @@ public async Task<HttpResponseMessage> GetStatusAsync(HttpRequestMessage request
             string q = null,
             string sortBy = null,
             bool? luceneQuery = true,
-            bool? debug = false)
+            bool? debug = false,
+            string packageType = null)
```
loic-sharma, line 88: Super nit: let's be consistent with the V3 route and keep `debug` last, and place `packageType` second to last.

src/NuGet.Services.SearchService/Controllers/SearchController.cs
```diff
@@ -97,6 +101,7 @@ public async Task<HttpResponseMessage> GetStatusAsync(HttpRequestMessage request
                 SortBy = GetSortBy(sortBy),
                 LuceneQuery = luceneQuery ?? true,
                 ShowDebug = debug ?? false,
+                PackageType = packageType,
```
loic-sharma, line 104: Same comment here, please swap ordering of `ShowDebug` and `PackageType`

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchParametersBuilderFacts.cs
```diff
@@ -7,6 +7,7 @@
 using Microsoft.Azure.Search;
 using Microsoft.Azure.Search.Models;
 using Moq;
+using NuGet.Packaging;
```
loic-sharma, line 10: It looks like this new `using` can be removed

src/NuGet.Services.AzureSearch/SearchService/SearchParametersBuilder.cs
```diff
@@ -202,6 +204,12 @@ private static IList<string> GetOrderBy(V2SortBy sortBy)
                 case V2SortBy.CreatedDesc:
                     orderBy = CreatedDesc;
                     break;
+                case V2SortBy.TotalDownloadsAsc:
+                    orderBy = TotalDownloadsAsc;
+                    break;
+                case V2SortBy.TotalDownloadsDesc:
+                    orderBy = TotalDownloadsDesc;
+                    break;
```
loic-sharma, line 212: The `IndexOperationBuilder.V2SearchWithHijackIndex` should check that the request does not have `packageType` or `orderBy` parameters. We should cover this with unit tests and functional tests.

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchResponseBuilderFacts.cs
t-mog-msft: If the _fiveSearchResults is used in the future (like _searchResult is used), it may require changing the initialization and move it to the tests. Since this behaviour is specific to these 2 tests, I would think that putting it here is better.

src/NuGet.Services.AzureSearch/SearchService/AzureSearchService.cs
```diff
@@ -34,6 +35,15 @@ public async Task<V2SearchResponse> V2SearchAsync(V2SearchRequest request)
         {
             if (request.IgnoreFilter)
             {
+                if (request.PackageType != null && request.PackageType.Any())
+                {
+                    throw new InvalidOperationException($"Can't apply the packageType filter on the Hijack index");
```
joelverhagen, line 40: nit: don't need `$`
This should bubble out at an HTTP 400, right? Does it?

tests/NuGet.Services.SearchService.Tests/Controllers/SearchControllerFacts.cs
```diff
@@ -184,8 +186,9 @@ public async Task SupportsNullParameters()
                 Assert.False(lastRequest.IncludePrerelease);
                 Assert.False(lastRequest.IncludeSemVer2);
                 Assert.Null(lastRequest.Query);
-                Assert.True(lastRequest.LuceneQuery);
+                Assert.True(lastRequest.LuceneQuery); ;
```
joelverhagen, line 189: nit: remove extra `;`

tests/NuGet.Services.AzureSearch.Tests/SearchService/SearchResponseBuilderFacts.cs
loic-sharma: > If the _fiveSearchResults is used in the future (like _searchResult is used), it may require changing the initialization and move it to the tests. 

Generally optimizing for what's known today is preferable than what may be in the future.  In any case, I don't feel strongly about this so I'll mark this as resolved.

src/NuGet.Services.AzureSearch/SearchService/AzureSearchService.cs
loic-sharma: You should be able to add a functional test to verify this does return an HTTP 400 response.

PR #798, opened 2020-07-02
Include vulnerability data in catalog2registration

drewgillies:
This change propagates vulnerability data from catalog blobs to registration indexes for a package.

Example:
```
            "vulnerabilities": [
              {
                "@id": "https://apidev.nugettest.org/v3/catalog0/data/2020.07.02.06.46.04/dotnetzip.1.0.0.json#vulnerability/GitHub/700",
                "@type": "Vulnerability",
                "advisoryUrl": "https://nvd.nist.gov/vuln/detail/CVE-2018-1002205",
                "gitHubDatabaseKey": "700",
                "severity": "2"
              },
              {
                "@id": "https://apidev.nugettest.org/v3/catalog0/data/2020.07.02.06.46.04/dotnetzip.1.0.0.json#vulnerability/GitHub/702",
                "@type": "Vulnerability",
                "advisoryUrl": "https://nvd.nist.gov/vuln/detail/CVE-2017-0929",
                "gitHubDatabaseKey": "702",
                "severity": "2"
              }
            ]
```


src/NuGet.Protocol.Catalog/Models/PackageVulnerability.cs
```diff
@@ -0,0 +1,25 @@
...
+        public string Id { get; set; }
+
+        [JsonProperty("@type")]
+        public string Type { get; set; }
+
+        [JsonProperty("advisoryUrl")]
+        public string AdvisoryUrl { get; set; }
+
+        [JsonProperty("gitHubDatabaseKey")]
+        public string GitHubDatabaseKey { get; set; }
```
loic-sharma, line 20: Is this property supposed to be included? The OneNote spec only mentions `severity`, `advisoryUrl`, and `@type` (see [this](https://microsoft.sharepoint.com/teams/NuGet/_layouts/OneNote.aspx?id=%2Fteams%2FNuGet%2FTeam%2FNugetServer%2FNugetServerTeamNote&wd=target%28Specs%20%28in%20progress%5C%29.one%7C910181C5-8695-4A1C-9792-A56DBB86F1C3%2FV3%20Protocol%20-%20Exposing%20Vulnerability%20information%7CE859AEBF-5731-4AAA-8066-DF8617F464C4%2F%29
onenote:https://microsoft.sharepoint.com/teams/NuGet/Team/NugetServer/NugetServerTeamNote/Specs%20(in%20progress).one#V3%20Protocol%20-%20Exposing%20Vulnerability%20information&section-id={910181C5-8695-4A1C-9792-A56DBB86F1C3}&page-id={E859AEBF-5731-4AAA-8066-DF8617F464C4}&object-id={DA482374-BD34-4266-B41F-5518BC39241F}&5D)).

src/NuGet.Services.V3/Registration/Models/RegistrationPackageVulnerability.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using Newtonsoft.Json;
+using NuGet.Protocol.Catalog;
+
+namespace NuGet.Protocol.Registration
+{
+    public class RegistrationPackageVulnerability : PackageVulnerability
+    {
+        [JsonProperty("registration", Order = 1)]
+        public string Registration { get; set; }
```
loic-sharma, line 12: Is this property used?

src/NuGet.Services.V3/Registration/Models/RegistrationPackageVulnerability.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Newtonsoft.Json;
+using NuGet.Protocol.Catalog;
+
+namespace NuGet.Protocol.Registration
+{
+    public class RegistrationPackageVulnerability : PackageVulnerability
```
loic-sharma, line 9: I don't think `RegistrationPackageVulnerability` should extend `PackageVulnerability` given that we in the future we may add properties to `PackageVulnerability` that shouldn't be added to `RegistrationPackageVulnerability`.

For example, registration has its own copy of the catalog entry with `RegistrationCatalogEntry`. I think it'd be best to follow that model.

src/NuGet.Services.V3/Registration/Models/RegistrationCatalogEntry.cs
```diff
@@ -75,5 +75,9 @@ public class RegistrationCatalogEntry
 
         [JsonProperty("version")]
         public string Version { get; set; }
+
+        [JsonProperty("vulnerabilities")]
+        public List<PackageVulnerability> Vulnerabilities { get; set; }
```
loic-sharma, line 80: It's better to be explicit with the type. If/when we ever move to System.Text.Json, that doesn't support polymorphic serialization.

```suggestion
        public List<RegistrationPackageVulnerability> Vulnerabilities { get; set; }
```

src/NuGet.Jobs.Catalog2Registration/Schema/EntityBuilder.cs
```diff
@@ -169,6 +169,15 @@ public void UpdateLeafItem(RegistrationLeafItem leafItem, HiveType hive, string
...
             catalogEntry.Title = packageDetails.Title ?? string.Empty;
             catalogEntry.Version = parsedVersion.ToFullString();
+
+            if (packageDetails.Vulnerabilities != null && packageDetails.Vulnerabilities.Count > 0)
+            {
+                catalogEntry.Vulnerabilities = packageDetails.Vulnerabilities;
+            }
+            else
+            {
+                catalogEntry.Vulnerabilities = new List<NuGet.Protocol.Catalog.PackageVulnerability> { new NuGet.Protocol.Catalog.PackageVulnerability() };
```
loic-sharma, line 179: Should we set this to `null` if there's no vulnerabilities? That seems to be the pattern we use if there's no deprecation.

src/NuGet.Jobs.Catalog2Registration/Schema/EntityBuilder.cs
```diff
@@ -169,6 +169,15 @@ public void UpdateLeafItem(RegistrationLeafItem leafItem, HiveType hive, string
 
             catalogEntry.Title = packageDetails.Title ?? string.Empty;
             catalogEntry.Version = parsedVersion.ToFullString();
+
+            if (packageDetails.Vulnerabilities != null && packageDetails.Vulnerabilities.Count > 0)
+            {
+                catalogEntry.Vulnerabilities = packageDetails.Vulnerabilities;
```
loic-sharma, line 175: The spec says:

> we will be inserting vulnerability info into the gz-semver2 hive only

It looks like this inserts the vulnerabilities information to all hives currently

src/NuGet.Protocol.Catalog/Models/PackageVulnerability.cs
drewgillies: It seems odd not to add it, as it would be useful metadata to have if we want to provide a link in the PM UI, for example. As well as your thoughts, @loic-sharma, I'd like @xavierdecoster and @joelverhagen on this one. Thoughts?

src/NuGet.Services.V3/Registration/Models/RegistrationPackageVulnerability.cs
drewgillies: Good catch--it was left over from copying existing types and deleting irrelevancies.

src/NuGet.Services.V3/Registration/Models/RegistrationPackageVulnerability.cs
drewgillies: Yes, agree. Thanks.

src/NuGet.Protocol.Catalog/Models/PackageVulnerability.cs
joelverhagen: I'll try to take a look at the rest of this PR tomorrow. Sorry about the delay. Regarding this particular issue, I recommend putting the bare minimum that we need for the client experience. If there is no scenario for this field in registration, we can exclude it. 

Catalog is immutable so it's trickier to change later (reflow all the things). For registration, we simply (...) do a rebuild and we can add the field later if we need it.

I agree with @loic-sharma. In fact, I would question even the `@type`. This V3 protocol spec was written before the Catalog2Registration rebuild so it would have been hard to not do the `@type`. Perhaps that was the consideration. Now it's very easy to do the minimal properties so I think we should consider dropping all but `severity` and `advisoryUrl`. The `@type` and `@id` made sense for generic graph processing a la JSON-LD and RDF. The reality is that the readers and writers of this JSON are very likely strongly typed code with models, etc. Client uses Newtonsoft.Json in a similar way to how this is written.

This brings up the question... has the client experience been designed? Do we know what we want to show on the client experience? How do we even know `advisoryUrl` and `severity` are needed? Or perhaps there is more needed?

Let's be sure the cart doesn't go before the horse (or is it the other way around...)

src/NuGet.Jobs.Catalog2Registration/Schema/EntityBuilder.cs
drewgillies: It is currently emitting a single empty vulnerability. I was simply following the lead of https://github.com/NuGet/NuGet.Services.Metadata/blob/b2a5100fb6111ddb569eb74af0cb2bc88918a22d/src/NuGet.Jobs.Catalog2Registration/Schema/EntityBuilder.cs#L167 for consistency. However, I expect that for actual processing of vulnerabilities client-side it would be neater to have nothing here. So, I could go either way on this, and generally opt for consistency in scenarios where I'm otherwise ambivalent. But if you have a strong opinion on this, I'm happy to be swayed. :)

loic-sharma: Hm the tags point is interesting. I'm not sure why we chose to have an empty string as the default, that may be vestigial from when we used JSON-LD and RDF to generate registration blobs.

> However, I expect that for actual processing of vulnerabilities client-side it would be neater to have nothing here.

I agree. Currently, clients will need to know how to interpret an empty vulnerability object. This does seem like a drawback.

> It is currently emitting a single empty vulnerability.

Keep in mind that unless we rebuild registration, we'll have a mix of registration blobs with no vulnerabilities and blobs with a single empty vulnerability. It seems like being consistent across all our blobs on the no vulnerabilities case would be better than being consistent with tags.

> But if you have a strong opinion on this, I'm happy to be swayed. :)

I'm currently leaning towards not emitting anything if a package has no vulnerabilities, but I could be swayed too 😄

src/NuGet.Jobs.Catalog2Registration/Schema/EntityBuilder.cs
drewgillies: Nice catch again! Thanks, @loic-sharma.

src/NuGet.Protocol.Catalog/Models/PackageVulnerability.cs
drewgillies: Sold. :) I'll include this change.

src/NuGet.Jobs.Catalog2Registration/Schema/EntityBuilder.cs
drewgillies: All good--you've sold me. I've made this change. Thanks!

src/NuGet.Jobs.Catalog2Registration/Schema/JsonLdConstants.cs
```diff
@@ -65,6 +65,11 @@ public static class JsonLdConstants
                 Container = "@set",
                 Id = "tag",
             },
+            Vulnerabilities = new ContextTypeDescription
```
joelverhagen, line 68: I don't think we need this at all.

tests/NuGet.Jobs.Catalog2Registration.Tests/Schema/EntityBuilderFacts.cs
```diff
@@ -221,6 +221,89 @@ public void PopulatesProperties()
...
+
+            [Fact]
+            public void PopulatesVulnerabilityProperties()
+            {
+                Hive = HiveType.SemVer2;
+                var leaf = V3Data.Leaf;
+                leaf.Vulnerabilities = new List<PackageVulnerability>() {
+                    new PackageVulnerability
+                    {
+                        Id = "https://apidev.nugetfoo.org/v3/catalog0/data/2020.07.06.06.49.47/bar.1.0.0.json#vulnerability/GitHub/999",
```
joelverhagen, line 233: nit: avoid domains that may be owned by someone else. apidev.nugettest.org is fine, or just https://example

src/NuGet.Services.V3/Registration/Models/RegistrationCatalogEntry.cs
```diff
@@ -75,5 +75,9 @@ public class RegistrationCatalogEntry
 
         [JsonProperty("version")]
         public string Version { get; set; }
+
+        [JsonProperty("vulnerabilities")]
+        public List<RegistrationPackageVulnerability> Vulnerabilities { get; set; }
+
```
loic-sharma, line 81: Nit: extra space here

src/NuGet.Protocol.Catalog/Models/PackageVulnerability.cs
loic-sharma: > has the client experience been designed? Do we know what we want to show on the client experience? How do we even know advisoryUrl and severity are needed? Or perhaps there is more needed?

This is a great point. We avoid breaking changes in the V3 protocol, so we shouldn't update the protocol until we understand the client experience. Should we block this PR until that's ironed out?

drewgillies: The client experience indicates a need for the advisory URL and the severity of the vulnerability:
* In the [dotnet CLI spec](https://github.com/NuGet/Home/blob/dev/designs/PackageVulnerability/DotnetListPackageVulnerable.md) we see a simple flagging of vulnerabilities with a (V), along with the advisory URL
* In the [PM UI spec](https://www.figma.com/file/8gCQx9iAQfPXTtxoks6Qd7/%F0%9F%9B%A1%EF%B8%8F-NuGet-Vulnerable-Packages) we can see that a vulnerable package will be flagged as such, with two hyperlinks: "more details" (advisory URL) and "fix" (seek out the first patched version, which may need to be added to metadata). I'll check in with @anangaur again on this detail.

src/NuGet.Jobs.Catalog2Registration/Schema/JsonLdConstants.cs
drewgillies: @joelverhagen np--it does add the vulnerabilities entry to the json context though:
```
    ""vulnerabilities"": {
      ""@id"": ""vulnerability"",
      ""@container"": ""@set""
    }
```
and without it we don't see that there. I put it in for consistency with other collections like `tags`, but I'll remove it as you've mentioned a de-emphasis on JSON-LD. Is this inconsistency a problem though? I can make a separate change if you want it back in.

PR #809, opened 2020-07-22
Remove gitHubDatabaseKey property from Catalog leaves

drewgillies:
This value while informational is not needed in the catalog leaves, and it is not propagated to registration metadata. Removing for consistency. Addresses https://github.com/NuGet/NuGetGallery/issues/8116.


src/Catalog/Helpers/GalleryDatabaseQueryService.cs
```diff
@@ -28,9 +28,9 @@ public class GalleryDatabaseQueryService : IGalleryDatabaseQueryService
         // {2} - outer ORDER BY if needed
         private static readonly string Db2CatalogSqlSubQuery = @"SELECT 
              P_EXT.*,
-                PV.[GitHubDatabaseKey] AS 'VulnerabilityGitHubDatabaseKey',
-                PV.[AdvisoryUrl] AS 'VulnerabilityAdvisoryUrl',
-                PV.[Severity] AS 'VulnerabilitySeverity'
+                PV.[GitHubDatabaseKey] AS '" + Db2CatalogProjectionColumnNames.VulnerabilityGitHubDatabaseKey + @"',
```
drewgillies, line 31: This change is cleanup--should have been done with original query changes.

